PR #107, opened 2017-07-24
Fix README typo

vyp:
typo right?

PS - I'm actually *enjoying* reading the readme, you've thought of a lot of stuff. :+1: Of course there may be more typos I haven't gone through everything to check.


raxod502:
I have no idea how this happened. I could swear I just copied and pasted from my terminal.

PR #111, opened 2017-07-24
Fix use-package integration when use-package is not loaded at runtime

PythonNut:
If `use-package` is never loaded, the `(with-eval-after-load 'use-package ...)` body never runs. To avoid this, I define the functions required by the compiled output of `use-package` unconditionally. The rest of the integration remains in the `(with-eval-after-load ...)` as before.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -3293,7 +3293,49 @@ This is an `:override' advice for
 
 ;;;; use-package integration
 
+;; Make it so that `:ensure' uses `straight-use-package' instead of
+;; `package-install'.
+(eval-and-compile
+  (defun straight-use-package-ensure-function
+      (name ensure state context &optional only-if-installed)
+    "Value for `use-package-ensure-function' that uses straight.el.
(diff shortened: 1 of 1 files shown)


raxod502:
This makes sense. We want to merge into `develop`, though, not `master`.

raxod502:
Wow, I didn't actually know you could do that without opening a new pull request.

raxod502:
In general we like commit messages that are 50 characters or less, see https://chris.beams.io/posts/git-commit/.

PythonNut:
@raxod502 what would you have suggested then?

raxod502:
How about:

```
Don't require use-package to be loaded at runtime

If use-package is never loaded, the (with-eval-after-load 'use-package
...) body never runs. To avoid this, I define the functions required
by the compiled output of use-package unconditionally. The rest of the
integration remains in the (with-eval-after-load ...) as before.
```

In general, I think it's unnecessary to capture all the important information in the subject line. That's what the body is for.

Issue #103, opened 2017-07-21
[Feature request] Options for async repo processing, deferring git actions


therockmandolinist:
I get the package doesn't even have a proper release yet, so for the future: It would be nice to have an option for the repo processing (pulling, normalizing, etc.) to be done asynchronously. As it is currently, with a lot of packages you sorta just have to sit and wait it out, which is mildly tedious. 

Along the same vein, it would be cool if one could "defer" the pop up asking what to do, that is, go through and pull and stuff, but don't ask, simply notify that an action can be taken. Then the user could call a function or something that prompts for what to do with the repos where an action is needed.

raxod502:
Good idea. It would be rather complicated to do this for the initial clone (as it would require the user to put all their package declarations in one place), but the bulk repository management operations could certainly have this done.

To your second suggestion, I agree that this would be interesting. It would however require a rethinking of how repository management works. The way it currently goes is in two layers. The first layer is a mapping over all the repositories that keeps track of skipped and canceled repositories; this layer calls some function to perform an action on any given repository. The second layer is a loop that repeatedly checks a series of conditions. If any of the conditions are not satisfied, some action is taken (possibly via a popup), and the loop restarts from the beginning. The reason it's done this way is because later validation steps (like doing a checkout) might violate earlier checks (like making sure the working directory is clean). So it's not that each repository has a single action run on it; it's actually much more sophisticated than that.

If you want to take a look at the code, the first layer is `straight--map-repos-interactively` and the second layer is the `straight--vc-git-*` functions.

Let me be clear though that I think your ideas are good and I'd like to see them implemented. It's just that the conceptual complexity means that without someone providing a pull request, I probably won't be able to get to them for a while :)

sooheon:
What are these "actions to be taken"? 

If we want installation to always be declarative and for init files to always be the source of truth, these actions should get variables inside `use-package` and `straight-use-package` and simply be declared up front. Then everything should be able to be moved to non-blocking, yes?

raxod502:
> What are these "actions to be taken"?

* Updating packages.
* Clearing local changes and resetting back to the canonical state.
* Pushing local changes so they can be made canonical.

[...]

PR #173, opened 2017-10-26
Separate fetch and merge steps in the pulling operation.

SkySkimmer:
The primary motivation is to be able to fetch all packages without
interrupting for user interaction (which takes network dependent time)
then in a separate step merge (which requires user interaction even if
only to hit the fast-forward button).

In a more conventional package manager these steps would be called
`update` and `upgrade` ;)

Depending on feedback the `pull` vc operations may be removed with the
high level pull-package and pull-all implemented in terms of fetch and
merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,7 +283,9 @@ interactive workflows to perform bulk operations on your packages.
 
 * To pull from each package's configured remote, run `M-x
   straight-pull-package` or `M-x straight-pull-all`. To also pull from
-  the upstream, if one is configured, provide a prefix argument.
+  the upstream, if one is configured, provide a prefix argument. The
+  fetch and merge steps can be separated using the commands
+  `straight-fetch-package` and `straight-merge-package` (or `-all`).
 
 * To push all local changes to each package's configured remote, run
   `M-x straight-push-package` or `M-x straight-push-all`.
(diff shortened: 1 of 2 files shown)


vyp:
Target should be `develop` btw.

SkySkimmer:
>Target should be develop btw.

It is now.

raxod502:
In the end, I'd like to make this feature unnecessary by implementing the more powerful general-purpose interface described in #103. However, pulling updates is indeed a significant pain point at the moment, so I think this is a good intermediate solution. Hopefully I will get to reviewing it soon.

Thank you for the contribution!

raxod502:
I have made some typographical edits to your patch, and split the `fetch` operation into `-from-remote` and `-from-upstream` variants for symmetry with `merge` and `pull`. The result is pushed to `develop` in https://github.com/raxod502/straight.el/commit/3f0276579239ba5f5d62322b5108aad826d1fbbf. I think this is a great improvement over the previous situation. Thank you!

Let me know if you have any feedback.

SkySkimmer:
I still believe more in `fetch --all` but OK.

raxod502:
Is there a use case that would benefit from using `fetch --all` instead of fetching individual remotes?

PR #181, opened 2017-11-07
Only attempt vc operations on packages with avaliable repos, fixes #110

### LINKED ISSUE #110: Packages with recipes, but which are not installed, error on straight-pull-all
If a package has a registered recipe, but is not installed, as may happen with `(use-package packagename :defer-install t ...)`, `straight-pull-all` will error when trying to pull the package, saying the repository directory does not exist (it doesn't). In this case, I believe it would be better to silently skip the package entirely.

I'll grab an exact copy of the message whenever I get a chance.

PythonNut:
1. In `straight--select-package` add a new parameter `repo-present` which filters recipes by whether they have an available repository.
2. User vc operations (e.g. `straight-pull-package`) have been modified to filter for package with available repositories when using `straight--select-package`.
3. Only map over recipes with available repos in `straight--map-repos`.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -3519,18 +3519,23 @@ RECIPE is a straight.el-style plist."
 ;;;; Interactive helpers
 ;;;;; Package selection
 
-(defun straight--select-package (message &optional for-build)
+(defun straight--select-package (message &optional for-build installed)
   "Use `completing-read' to select a package.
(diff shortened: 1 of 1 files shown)


raxod502:
This breaks `straight-freeze-versions` by causing the revisions of packages which are registered but not currently installed to be silently dropped.

PR #215, opened 2018-01-03
Add Edebug specs

xuchunyang:
I'm trying to figure out how `straight-use-package` / `straight--build-package` works with Edebug, but it can't step inside macros due to lacking the Edebug specs.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -353,7 +353,7 @@ alist, to ensure correct results."
 PROPS is a list of symbols. Each one is converted to a keyword
 and then its value is looked up in the PLIST and bound to the
 symbol for the duration of BODY."
-  (declare (indent 2))
+  (declare (indent 2) (debug (form sexp body)))
   (let ((plist-sym (make-symbol "plist")))
     `(let* ((,plist-sym ,plist)
             ,@(mapcar (lambda (prop)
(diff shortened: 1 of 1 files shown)


raxod502:
I'm happy to merge this. But given that I don't use Edebug, could you explain what these changes do?

xuchunyang:
[(elisp) Edebug](https://www.gnu.org/software/emacs/manual/html_node/elisp/Edebug.html)

> Edebug is a source-level debugger for Emacs Lisp programs, with which
> you can:
> 
>    * Step through evaluation, stopping before and after each expression.

[(elisp) Edebug and Macros](https://www.gnu.org/software/emacs/manual/html_node/elisp/Edebug-and-Macros.html)

> To make Edebug properly instrument expressions that call macros, some
> extra care is needed

[(elisp) Declare Form](https://www.gnu.org/software/emacs/manual/html_node/elisp/Declare-Form.html)

>     `(debug EDEBUG-FORM-SPEC)'
>           This is valid for macros only.  When stepping through the
>           macro with Edebug, use EDEBUG-FORM-SPEC.  *Note Instrumenting
>           Macro Calls::.

raxod502:
OK! Thanks for the information, good to know.

Issue #214, opened 2018-01-03
Add minimal integration test to be run in CI


raxod502:
We are unlikely to see real unit (#69) or integration (#70) tests for quite a while, but in the meantime it would probably be a good idea to have a minimal integration test that just bootstraps `straight.el` and installs a package, for sanity's sake.

See #212.

PR #216, opened 2018-01-05
Actually use straight.el on Travis CI

xuchunyang:
For #214.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,11 @@ all:
 .PHONY: travis
 travis:
 	@elint/elint checkdoc compile longlines
+	@mkdir -p ~/.emacs.d/straight/repos/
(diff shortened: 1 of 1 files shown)


raxod502:
This is good, but there is a significant problem, I think: it does not test the version of `straight.el` that is being tested on Travis, but rather the latest one from `master` (as that is what `install.el` installs, by default). How about linking or copying the repository into `~/.emacs.d/straight/repos/straight.el` and loading `bootstrap.el`, rather than `install.el`?

xuchunyang:
I have followed your suggestion. Now it uses the correct version of straight.el. I also squash and force-push my commits.

raxod502:
Great, thanks!

PR #219, opened 2018-01-08
Warn user if building info documentation fails

### LINKED ISSUE #210: Don't just fail silently when texinfo compilation fails
After switching straight.el from package.el, I notice the info manual of use-package is missing, it takes me a while to figure out it is because the command fails

> $ makeinfo use-package.texi
> ...
> makeinfo: Removing output file `/Users/xcy/.emacs.d/straight/repos/use-package/use-package.info' due to errors; use --force to preserve.

(the built-in `/usr/bin/makeinfo` on macOS is too old for this use-package.texi)

I hope straight.el can at least let me know when the command fails, even `message` something like the following helps.

>Command "makeinfo use-package.texi" fails

xuchunyang:
This is my attempt to fix #210.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -789,6 +789,15 @@ is the stdout and stderr of the command."
                                  nil '(t t) nil args))))
         (cons success (buffer-string))))))
 
+(defun straight--warn-call (command &rest args)
+  "Call COMMAND with ARGS, warning user if it fails."
+  (let ((result (apply #'straight--call command args)))
+    (unless (car result)
+      (straight--warn
+       "Command failed: %s %s (default-directory: %S):\n%s"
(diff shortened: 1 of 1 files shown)


raxod502:
Thanks.

raxod502:
I have partially reverted this pull request. Dumping the entire error message to `*Warnings*` has not been a good user experience, I think. It would be better to print a simple (and non-scary warning), such as "Couldn't build the documentation for \<package>; this is a packaging error". Then have a way to view the whole log (print to a special buffer and introduce an interactive command for showing that buffer?).

In the end, I'd like to do something like Magit's process buffer, which allows you to see output optionally by expanding via TAB.

PR #222, opened 2018-01-16
Fix return value of straight-use-package

xuchunyang:
Without the fix, straight-use-package returns nil for built-in packages, however, according to the docstring and realted comments, it should return non-nil instead.


raxod502:
Thanks! Good catch.

PR #224, opened 2018-02-01
Replace if-let and when-let with if-let* and when-let*

xuchunyang:
I built Emacs from the latest git source recently and `make` fails

```
-*- mode: compilation; default-directory: "~/.emacs.d/straight/repos/straight.el/" -*-
Compilation started at Thu Feb  1 10:47:11

make -k 
==> elint: compile: straight.el

In toplevel form:
straight.el:329:53:Error: ‘if-let’ is an obsolete macro (as of 26.1); use ‘if-let*’ instead.
==> elint: compile: bootstrap.el
straight.el: ‘if-let’ is an obsolete macro (as of 26.1); use ‘if-let*’ instead.
straight.el: ‘if-let’ is an obsolete macro (as of 26.1); use ‘if-let*’ instead.
straight.el: ‘if-let’ is an obsolete macro (as of 26.1); use ‘if-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘if-let’ is an obsolete macro (as of 26.1); use ‘if-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
straight.el: ‘if-let’ is an obsolete macro (as of 26.1); use ‘if-let*’ instead.
straight.el: ‘if-let’ is an obsolete macro (as of 26.1); use ‘if-let*’ instead.
straight.el: ‘when-let’ is an obsolete macro (as of 26.1); use ‘when-let*’ instead.
make: *** [all] Error 1

Compilation exited abnormally with code 2 at Thu Feb  1 10:47:13
```

As you can see `if-let` and `when-let` becomes obsolete since Emacs 26.1.

---
However, I notice we are copy-and-paste about 100 lines of code just to have `when-let` and `if-let` available, maybe it is better to roll our own for less lines of code and easy to maintain? for example,

```elisp
(defmacro straight-if-let (var-val then &rest else)
  (declare (indent 2) (debug ((symbolp form) form body)))
  `(let ((,(car var-val) ,(cadr var-val)))
     (if ,(car var-val)
         ,then
       ,@else)))

(defmacro straight-when-let (var-val &rest body)
  (declare (indent 1) (debug ((symbolp form) body)))
  `(straight-if-let ,var-val ,(macroexp-progn body)))
```

this is not as powerful as Emacs's but it is enough for all our current uses, i.e., straight.el should work as usual if we replace `if-let` with `straight-if-let` now.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -38,7 +38,7 @@
 
 ;;;; Libraries
 
-;; For `if-let', `when-let', `hash-table-keys', `string-join',
+;; For `if-let*', `when-let*', `hash-table-keys', `string-join',
 ;; `string-trim', etc.
 (require 'subr-x)
 
@@ -136,9 +136,6 @@ threading."
 (eval-when-compile
   (when (version< emacs-version "25.1")
 
-    ;; Defined by Emacs 25.3 in C source code
-    (defvar inhibit-message)
-
     ;; Definition from Emacs 25.3, subr.el
     (gv-define-expander alist-get
       (lambda (do key alist &optional default remove)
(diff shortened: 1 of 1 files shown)


straight.el
```diff
@@ -167,35 +164,44 @@ threading."
                                  (,p ,(funcall setter
                                                `(delq ,p ,getter)))))))))))))))
 
+    ;; Defined by Emacs 25.3 in C source code
+    (defvar inhibit-message)))
```
raxod502: Why did this move, just out of curiosity?

raxod502:
This is all good (aside from minor comment above). As for `straight-if-let`, I don't have strong feelings about it. My inclination is to leave it the way it is for now, since I don't anticipate a maintenance burden (the definition doesn't really need to be updated ever) and switching to `straight-if-let` would make the lines longer :P

straight.el
xuchunyang: Because otherwise `make longlines` fails at line 165 (81 chars)

```
                                               `(delq ,p ,getter)))))))))))))))))
```

xuchunyang:
> switching to straight-if-let would make the lines longer

Right, I tried it at https://github.com/xuchunyang/straight.el/commit/31ad3e5d0512cbf4dd52abac890330d22836cd8e and I have to re-indent many places to pass `make longlines`.

xuchunyang:
Another way to silence the byte compiler warnings for Emacs 26 is

```
(require 'subr-x)
(put 'if-let   'byte-obsolete-info nil)
(put 'when-let 'byte-obsolete-info nil)
```

straight.el
raxod502: I think a more elegant solution in general would be to break one of the lines in the function definition. But it's not all that important, so I just let it be.

raxod502:
Thanks!

raxod502:
Reverted now that Emacs 26.1-rc1 is out, fixing the obnoxious decision to make `if-let` and `when-let` obsolete. Thank goodness.

PR #225, opened 2018-02-01
negate the return value of straight--checkhash

jvshahid:
according to the docs this functions should return non-NIL if the key is found
in the hash. the actual code is returning non-NIL if the key is absent from the
hash table.

this also fixes rebuilding modified packages since straight--checkhash was used
straight--package-might-be-modified-p

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -426,10 +426,11 @@ API didn't actually know how to write hash table APIs.")
   "Return non-nil if KEY is present in hash TABLE.
 If PARANOID is non-nil, ensure correctness even for hash tables
 that may contain `straight--not-present' as a value."
-  (and (eq (gethash key table straight--not-present) straight--not-present)
-       (or paranoid
-           (eq (gethash key table straight--not-present-paranoid)
-               straight--not-present-paranoid))))
+  (not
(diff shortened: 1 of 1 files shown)


jvshahid:
just noticed that the tests failed. i'll fix them and push once i get a chance

raxod502:
Thanks!

raxod502:
Oops, I didn't notice this was based against `master`. Development actually takes place on `develop`. Since this has happened quite a few times, I am obviously not communicating that fact very clearly. I've [made a note at the top of the README](https://github.com/raxod502/straight.el/commit/cd6cf530805df28c0158456a823fea9c561af333) about it.

jvshahid:
You can change the default branch of the repo to be `develop' instead of master. That could help with catching those kinds of mistakes since new PRs will be created with `develop' as the base branch.

raxod502:
That's true, but it would rather defeat the purpose of having `master` as a more stable default that people see first (and I'd rather not have people notice a cool feature in the README, then install `straight.el` and find that the feature doesn't exist since installation uses `master` by default).

PR #227, opened 2018-02-04
Improve handling of built-in packages

justbur:
Use straight-built-in-pseudo-packages in straight--convert-recipe.
Gather and check list of packages found to be built-in with
straight--found-built-in-packages.

This improves my startup time significantly.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -2307,7 +2307,7 @@ is built in to Emacs (e.g. the \"emacs\" package). This is used
 for dependency resolution."
   ;; Special case for the `emacs' pseudo-package, so that by default
   ;; we don't try to look up a recipe in recipe repositories.
-  (when (eq melpa-style-recipe 'emacs)
+  (when (memq melpa-style-recipe straight-built-in-pseudo-packages)
     (cl-return-from straight--convert-recipe))
   ;; Firstly, if the recipe is only provided as a package name, and
   ;; we've already converted it before, then we should just return the


raxod502:
What is the bottleneck that this change alleviates?

If the issue is that `straight-recipes-retrieve` takes nontrivial time to execute, we should look into caching the result of `straight-recipes-list` from each recipe repository when inside a transaction.

Just inhibiting lookup for the common dependency `cl-lib` is not really appropriate, given that some package repositories (e.g. GNU ELPA) actually do provide the package `cl-lib`.

justbur:
I can see removing `cl-lib` from the list. I just did that.

I was doing some profiling, and `straight-recipes-retrieve` was getting called for `cl-lib`, `let-alist`, and `seq` many times at startup. Excluding these using one of these lists gets me half a second now (it seemed to be getting me closer to a second earlier). 

At the very least, I'd like to be able to use `straight-built-in-pseudo-packages`. It was defined, but I couldn't find a place it was used in the code. I don't care if the variable that automatically records the built-in packages is added, since I can use the one that's already there.

justbur:
My goal was to match my el-get startup time, and this is one way for me to get there. It's roughly the same amount of savings as setting `straight-check-for-modifications` to `live`for me.

raxod502:
Whoops, you're indeed right that `straight-built-in-pseudo-packages` was not used anywhere. I think I meant to check it with `memq` instead of just checking for `emacs`, but then forgot. I've fixed that now.

My concern with your current idea, though, is that it introduces a potentially surprising corner case: let's say that someone tries to install `cl-lib` and `straight.el` doesn't do anything, because it's detected to be built-in. But what they really meant was to install `cl-lib` from a custom recipe repository, so they register that recipe repository and try again. This will still fail, because `cl-lib` will remain in `straight--found-built-in-packages` forever (unless the user specifically evaluates a `setq` for it).

I think it may be possible to fix this pretty easily. Instead of adding the package name to `straight--found-built-in-packages`, how about we make it so that `straight--convert-recipe` will store a nil value in `straight--recipe-cache` for that package, and then return it on further requests? That way, when the user reloads their init-file the cached value will be re-set.

justbur:
Yep, seems reasonable. It's more work to use the cache. Just using `nil` won't work with the existing code, because there would be no distinction between a found `nil` and `gethash` not finding a key. 

For now, I just fixed the `straight-built-in-pseudo-packages` issue. That's sufficient for my purposes at the moment. I reverted the other changes.

raxod502:
You can use `straight--checkhash` to determine if the hash has a key. It will be a bit of control-flow refactoring, though.

justbur:
The original issue is fixed on develop. Didn't notice that's where the main development was happening

PR #229, opened 2018-02-05
Support :straight (pkg-a pkg-b) syntax in use-package

justbur:
Previously this resulted in an error because it expanded
to (straight-use-package '(pkg-a pkg-b)).

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -4586,10 +4586,11 @@ documentation."
        ((null arg) (setq parsed-args nil))
        ((eq arg t) (push name-symbol parsed-args))
        ((symbolp arg) (push arg parsed-args))
-       ((not (listp arg))
-        (use-package-error ":straight wants a symbol or list"))
-       ((keywordp (car arg)) (push (cons name-symbol arg) parsed-args))
-       (t (push arg parsed-args))))
+       ((and (listp arg) (keywordp (car arg)))
+        (push (cons name-symbol arg) parsed-args))
+       ((listp arg) (setq parsed-args (append arg parsed-args)))
+       (t
+        (use-package-error ":straight wants a symbol or list"))))
     parsed-args))
 
 (defun straight-use-package--straight-handler


raxod502:
What you want is already supported:

```elisp
(use-package pkg
  :straight pkg-a
  :straight pkg-b)
```

Also keep in mind that the following syntax must also be supported:

```elisp
(use-package tex
  :recipe (auctex :type git :host github :repo "emacsmirror/auctex"))
```

justbur:
The syntax `:straight (pkg-a pkg-b)` is common in `use-package` (see `:commands`, `:load-path`, and more), is not treated as an error, and gives an invalid expansion. 

You are supporting the example you gave, as well as `:straight pkg-a pkg-b`, but both of those seem unconventional compared to `:straight (pkg-a pkg-b)`. 

I don't understand the significance of your second point, unless you meant to write `:straight (auctex :type git :host github :repo "emacsmirror/auctex")`.

justbur:
Just looked at the README again. I see the problem. Let me try something else

justbur:
@raxod502 Can you see if it's acceptable now? I detect recipes by the presence of a keyword and otherwise recursively process the arg.

straight.el
```diff
@@ -4586,10 +4586,11 @@ documentation."
        ((null arg) (setq parsed-args nil))
        ((eq arg t) (push name-symbol parsed-args))
        ((symbolp arg) (push arg parsed-args))
-       ((not (listp arg))
-        (use-package-error ":straight wants a symbol or list"))
```
raxod502: Why did you move this condition down to the end? It seems like that introduces a duplicate check for `(listp arg)`, instead of having just one.

raxod502:
Regarding the substance of the change, I'm not opposed to allowing multiple packages and/or recipes wrapped in a list rather than as separate top-level forms.

But we also need to continue supporting a single recipe at the top-level, which your change doesn't currently do. In addition to checking the first element of a list, you need to check the _second_ element to see if it's a keyword—that would indicate a recipe, like in:

```elisp
(use-package tex
  :straight (auctex :type git :host github :repo "emacsmirror/auctex"))
```

rather than a list of package names `(pkg-a pkg-b)`.

Also, for consistency, I'd like to support lists of recipes and/or package names:

```elisp
(use-package tex
  :straight (no-littering (auctex :type git :host github :repo "emacsmirror/auctex")))
```

which would translate to

```elisp
(straight-use-package 'no-littering)
(straight-use-package '(auctex :type git :host github :repo "emacsmirror/auctex"))
```

straight.el
raxod502: This might be outdated though, it seems that the branch has been deleted.

raxod502:
Oh, I see this was moved to a new pull request. Just FYI, GitHub does let you change the base branch of a pull request and also force-push new commits to the target branch, so you don't have to create a new pull request for every revision.

PR #231, opened 2018-02-05
Support :straight (pkg-a pkg-b) syntax in use-package

justbur:
Previously this resulted in an error because it expanded
to (straight-use-package '(pkg-a pkg-b)).

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -4599,8 +4599,17 @@ documentation."
        ((symbolp arg) (push arg parsed-args))
        ((not (listp arg))
         (use-package-error ":straight wants a symbol or list"))
-       ((keywordp (car arg)) (push (cons name-symbol arg) parsed-args))
-       (t (push arg parsed-args))))
+       ((keywordp (car arg))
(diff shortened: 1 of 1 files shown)


straight.el
```diff
@@ -4599,8 +4599,17 @@ documentation."
...
+        ;; recipe without package name
+        (push (cons name-symbol arg) parsed-args))
+       ((cl-some #'keywordp arg)
+        ;; assume it's a recipe
+        (push arg parsed-args))
+       (t
+        (setq parsed-args
+              (append parsed-args
+                      (straight-use-package--straight-normalizer
+                       name-symbol nil arg))))))
```
raxod502: For consistency in ordering, can you have this part put the normalized list on the beginning of `parsed-args` rather than the end?

raxod502:
Looks good to me. Just one comment.

straight.el
justbur: No problem. It's fixed now

raxod502:
Thanks.

PR #241, opened 2018-02-21
Refactor pull operations to use fetch and merge

xendk:
`straight-pull-all` now feches first, before merging all.

I've removed the low-level git pull commands as they're functionally covered by fetch and merge, and keeping the VC interface as simple as possible seems like a good idea.

Addresses some of #103

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -1183,39 +1183,6 @@ repository directory and delegates to the relevant
       (let ((straight--default-directory (straight--repos-dir local-repo)))
         (straight-vc 'merge-from-upstream type recipe)))))
 
-(defun straight-vc-pull-from-remote (recipe)
-  "Pull from the primary remote for straight.el-style RECIPE.
-
-If the RECIPE does not specify a local repository, then no action
-is taken.
-
-This method sets `straight--default-directory' to the local
-repository directory and delegates to the relevant
-`straight-vc-TYPE-pull-from-remote' method, where TYPE is the
-`:type' specified in RECIPE."
-  (straight--with-plist recipe
-      (local-repo type)
-    (when local-repo
-      (let ((straight--default-directory (straight--repos-dir local-repo)))
-        (straight-vc 'pull-from-remote type recipe)))))
-
-(defun straight-vc-pull-from-upstream (recipe)
-  "Pull from the upstream remote for straight.el-style RECIPE.
-If there is no upstream configured, this method does nothing.
-
-If the RECIPE does not specify a local repository, then no action
-is taken.
-
-This method sets `straight--default-directory' to the local
-repository directory and delegates to the relevant
-`straight-vc-TYPE-pull-from-upstream' method, where TYPE is the
-`:type' specified in RECIPE."
-  (straight--with-plist recipe
-      (local-repo type)
-    (when local-repo
-      (let ((straight--default-directory (straight--repos-dir local-repo)))
-        (straight-vc 'pull-from-upstream type recipe)))))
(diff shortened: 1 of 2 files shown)


xendk:
`straight-check-package` is broken, but I get the same behavior on the ndevelop branch.

straight.el
```diff
@@ -1745,20 +1712,10 @@ name."
...
-        (and (straight-vc-git--validate-local recipe)
-             (or already-fetched
-                 (progn
-                   (straight--get-call
-                    "git" "fetch" remote)
-                   (setq already-fetched t)))
-             (straight-vc-git--validate-head
-              local-repo branch (format "%s/%s" remote remote-branch))
-             (cl-return-from straight-vc-git--pull-from-remote-raw t))))))
+  (straight-vc-fetch-from-remote recipe)
```
raxod502: I would prefer to avoid round-tripping through the top-level VC API here. Can we use `straight-vc-git-fetch-from-remote` instead?

straight.el
```diff
@@ -1745,20 +1712,10 @@ name."
...
+  (cl-return-from straight-vc-git--pull-from-remote-raw
```
raxod502: I think this `cl-return-from` is superfluous.

straight.el
```diff
@@ -4223,9 +4152,11 @@ not just from primary remote but also from configured upstream."
   (interactive (list (straight--select-package "Pull package" nil 'installed)
                      current-prefix-arg))
   (let ((recipe (gethash package straight--recipe-cache)))
-    (and (straight-vc-pull-from-remote recipe)
+    (and (and (straight-vc-fetch-from-remote recipe)
+              (straight-vc-merge-from-remote recipe))
```
raxod502: These two `and`s can be collapsed together.

raxod502:
I think this is a good idea. Further work is needed for the user experience, but this pull request is a step in the right direction. I just had a few stylistic comments, and then I will merge it.

raxod502:
Also, could you elaborate on what you mean by `straight-check-package` being broken? It works for me.

xendk:
I'll fix those things when I have time.

I've tried taking a random package, doing a `git checkout HEAD^`, run `straight-rebuild-package` on it and restarting Emacs. After running fetch/merge (and telling it to reattach to origin/HEAD), I would expect it to rebuild the package, but it does nothing (as does `straight-check-all`). `straight-check-for-modifications` is `live`

straight.el
xendk: Sure.

straight.el
xendk: Why? Doesn't it make sense for `straight-vc-git--pull-from-remote-raw` to return the same as `straight-vc-git--merge-from-remote-raw`?

Or is is that `cl-return-from` is just for "early breakout"? I must admit that I had to read up on `cl-return-from` when I encountered it, so I haven't quite internalized it's working. But if that's the case, I think there's more that can be eliminated (`straight-vc-git--merge-from-remote-raw` for instance).

straight.el
xendk: Oh, of course.

raxod502:
If you have `straight-check-for-modifications` set to `live`, then not all modifications will be detected. In particular, `straight.el`'s repository management operations are not detected as modifications, which is definitely a bug. (I would not classify live modification detection as production-ready yet, unlike eager modification detection.) Feel free to open another issue for that.

xendk:
I've implemented your suggestions.

Issue #158, opened 2017-10-05
Lots of prompts when doing straight-pull-all is tiring


vyp:
After reproducing an emacs config from the version lockfile on another machine, I did `straight-pull-all` on it to update all my packages. Then this prompt popped up for each package:

https://github.com/raxod502/straight.el/blob/87525b5b89ed8b6a6823422b0d9b0cb097eeac04/straight.el#L1038-L1040

So I had to press `a` for all packages to update them, which you can believe can be really repetitive and tiring especially if you have a lot of packages (like 100s). I think it was detached because `git checkout <commit>` was used for all the packages to get them on the correct revisions as specified by the lockfile? Would using `git reset --hard <commit>` instead work (to avoid the detached head state)?

raxod502:
Yes, using `git reset --hard` would be preferable, and I was already planning to do that in #66. Another issue is that `M-x straight-pull-all` gives you a prompt for each package that must be upgraded to a newer version, which could be remedied by a better interface as discussed [here](https://github.com/raxod502/straight.el/issues/103#issuecomment-331798428).

vyp:
Oh right so you're saying there would be a popup for each package even if the repos were not in detached head states?

Yeah so similar to #103 I just want to say update all my packages and then come back to them all updated (or not if there's failures) without having to stay at emacs answering all the prompts. Even if the operation is synchronous, it would still be better than how it is now in my opinion.

My searches didn't find #66 though, I feel like this issue is a duplicate of that? PS - Nice title for #66.

xendk:
In the vein of "baby steps" how about making `straight-vc-git--validate-head` quietly fast forward or `git reset --hard` whenever the currently checked out commit is an ancestor of the remote branch head? That would eliminate all questions in the most common case.

IvanMalison:
@xendk Yeah that seems very reasonable to me.

raxod502:
Sounds good. I can't deny it's an improvement over the current situation.

xendk:
I did this, but then got cold feet.

Feels wrong to have `straight-vc-git--validate-head` quietly pull in changes. It would also affect `straight-normalize-package`, which might not be what you'd want. It's also a nice "ask the user what we should do" function.

I considered adding a `ff-ok` argument to make it optional, but that would cascade through a lot of functions.

[...]

PR #245, opened 2018-03-01
Automatically fast-forward and reattach head

xendk:
As discussed in #158 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -1666,6 +1666,10 @@ You can customize the following user options:
   operations will offer to re-set your remote URLs from HTTPS to SSH
   or vice versa, depending on the value of
   `straight-vc-git-default-protocol`. This is nil by default.
+* `straight-vc-git-auto-fast-forward`: if this is non-nil, pulling will
+  quietly do fast-forward, to suppress asking for instructions on each
+  package with updates, unless they're not trivial. Set to nil if
+  you'd prefer to inspect all changes.
 
 ### Recipe lookup
 
--- a/straight.el
+++ b/straight.el
@@ -1620,7 +1620,9 @@ confirmation, so this function should only be run after
          (when (and ref ref-ahead-p straight-vc-git-auto-fast-forward)
            ;; Local is behind, catch up.
            (straight--get-call "git" "reset" "--hard" ref-name)
-           (cl-return-from straight-vc-git--ensure-head t))
+           ;; Return nil to signal that we're not quite done. In some
+           ;; cases a reset might leave untracked files.
+           (cl-return-from straight-vc-git--ensure-head nil))
          (straight-vc-git--popup-raw
           (concat
            (format "In repository %S, " local-repo)
--- a/straight.el
+++ b/straight.el
@@ -1442,7 +1442,7 @@ Do not suppress unexpected errors."
 ;;;;;; Validation functions
 
 (cl-defun straight-vc-git--ensure-remote (local-repo remote desired-url)
-  "Validate that LOCAL-REPO has REMOTE set to DESIRED-URL or equivalent.
+  "Ensure that LOCAL-REPO has REMOTE set to DESIRED-URL or equivalent.
 All three arguments are strings. The URL of the REMOTE does not
 necessarily need to match DESIRED-URL; it just has to satisfy
 `straight-vc-git--urls-compatible-p'."
@@ -1515,7 +1515,7 @@ but recipe specifies a URL of
       "git" "remote" "add" remote desired-url))))
 
 (cl-defun straight-vc-git--ensure-remotes (recipe)
-  "Validate that repository for RECIPE has remotes set correctly.
+  "Ensure that repository for RECIPE has remotes set correctly.
 RECIPE is a straight.el-style plist.
 
 This means the primary and upstream remotes, if configured, have
@@ -1540,7 +1540,7 @@ to satisfy `straight-vc-git--urls-compatible-p'."
                     desired-url))))))))
 
 (defun straight-vc-git--ensure-nothing-in-progress (local-repo)
-  "Validate that no merge conflict is active in LOCAL-REPO.
+  "Ensure that no merge conflict is active in LOCAL-REPO.
 LOCAL-REPO is a string."
   (let ((conflicted-files
          (string-remove-suffix
@@ -1561,7 +1561,7 @@ LOCAL-REPO is a string."
             (straight--get-call "git" "merge" "--abort")))))))
 
 (cl-defun straight-vc-git--ensure-worktree (local-repo)
-  "Validate that LOCAL-REPO has a clean worktree.
+  "Ensure that LOCAL-REPO has a clean worktree.
 LOCAL-REPO is a string."
   (let ((status (straight--get-call-raw
                  "git" "-c" "status.branch=false"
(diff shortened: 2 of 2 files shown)


SkySkimmer:
Please gate automatic fast-forward with an option, I like getting prompted so I can look at what commits are new. (I don't care if the option is on by default)

straight.el
```diff
@@ -1598,56 +1598,37 @@ confirmation, so this function should only be run after
...
                            "git" "merge-base" "--is-ancestor"
                            branch ref-name))
              (ref-behind-p (straight--check-call
                             "git" "merge-base" "--is-ancestor"
                             ref-name branch)))
-         (when (and ref ref-behind-p)
+         (when ref-ahead-p
+           ;; Local is behind, catch up.
+           (straight--get-call "git" "reset" "--hard" ref-name)
            (cl-return-from straight-vc-git--validate-head t))
```
raxod502: I don't think we should be returning non-nil here. Usually there will be more work to do with cleanup after a reset, no?

straight.el
```diff
@@ -1598,56 +1598,37 @@ confirmation, so this function should only be run after
...
        (let ((ref-ahead-p (straight--check-call
                            "git" "merge-base" "--is-ancestor"
                            branch ref-name))
              (ref-behind-p (straight--check-call
                             "git" "merge-base" "--is-ancestor"
                             ref-name branch)))
-         (when (and ref ref-behind-p)
+         (when ref-ahead-p
+           ;; Local is behind, catch up.
+           (straight--get-call "git" "reset" "--hard" ref-name)
```
raxod502: This should only happen automatically in the case that `ref` is non-nil, I think.

straight.el
```diff
@@ -1598,56 +1598,37 @@ confirmation, so this function should only be run after
...
            (cl-return-from straight-vc-git--validate-head t))
+         (when ref-behind-p
+           (if ref (cl-return-from straight-vc-git--validate-head t)
+             (straight--get-call "git" "checkout" branch)))
```
raxod502: This definitely should never happen automatically, as it's not a fast-forward.

straight.el
```diff
@@ -1598,56 +1598,37 @@ confirmation, so this function should only be run after
...
-                                branch quoted-ref-name)
-                   ,(lambda ()
-                      (straight--get-call
-                       "git" "reset" "--hard" ref-name)))))
-            ,@(when (and ref-behind-p (null ref))
-                `(("f" ,(format "Fast-forward HEAD to branch %S" branch)
-                   ,(lambda ()
-                      (straight--get-call
-                       "git" "checkout" branch)))))
-            ,@(unless (or ref-ahead-p ref-behind-p)
```
raxod502: These changes should be conditional on a new option, as discussed above.

straight.el
xendk: There is? `git reset --hard` reverts the index and the worktree, it's always been my get-out-of-jail card when I've messed up, so I don't know what cleanup you're referring to?

straight.el
xendk: Well, if `ref` is nil, `ref-name` is `"HEAD"`, so it basically checks if the branch is even with itself (detached heads and other branches are dealt with earlier), so the reset should be a no-op in most cases (dirty worktrees should have been dealt with in `straight-vc-git--validate-worktree`).

So yeah, you're right, it should depend on `ref`, at least for efficiency.

straight.el
xendk: I... Have no idea what I was thinking.

straight.el
raxod502: Take a look at this example:

```
example-git-repo % git init
Initialized empty Git repository in /private/tmp/example-git-repo/.git/
example-git-repo[master] % git commit --allow-empty -m "Initial commit"
[master (root-commit) 352e524] Initial commit
example-git-repo[master] % echo 'ignored-file' > .gitignore
example-git-repo[master*] % touch ignored-file
example-git-repo[master*] % git add .gitignore
example-git-repo[master*] % git commit -m "Add .gitignore"
[master e5420a0] Add .gitignore
 1 file changed, 1 insertion(+)
 create mode 100644 .gitignore
example-git-repo[master] % git status
On branch master
nothing to commit, working tree clean
example-git-repo[master] % git reset --hard @~
HEAD is now at 352e524 Initial commit
example-git-repo[master*] % git status
On branch master
Untracked files:
        ignored-file

nothing added to commit but untracked files present
```

README.md
```diff
@@ -1666,6 +1666,10 @@ You can customize the following user options:
   operations will offer to re-set your remote URLs from HTTPS to SSH
   or vice versa, depending on the value of
   `straight-vc-git-default-protocol`. This is nil by default.
+* `straight-vc-git-force-protocol`: if this is non-nil, pulling will
```
raxod502: Do you mean `straight-vc-git-auto-fast-forward`?

xendk:
OK, take two, now less ambitious and with an option.

I've renamed all the validate functions, as they all potentially change the repository.

straight.el
xendk: So we should do a `git clean -df` afterwards?

README.md
xendk: Doh... Rewriting history...

straight.el
raxod502: No, just return nil. If you look at how I've structured the validation functions, each one returns nil if some previously established state (such as a clean worktree) might have been destroyed. Then, I chain them together by running the sequence of validation functions until all of them return non-nil. In that case, the desired validation has been effected.

This should probably be documented somewhere, oops.

xendk: Oooh, right.

Fixed.

straight.el
```diff
@@ -1433,7 +1441,7 @@ Do not suppress unexpected errors."
 
 ;;;;;; Validation functions
 
-(cl-defun straight-vc-git--validate-remote (local-repo remote desired-url)
+(cl-defun straight-vc-git--ensure-remote (local-repo remote desired-url)
   "Validate that LOCAL-REPO has REMOTE set to DESIRED-URL or equivalent.
```
raxod502: If we are renaming `validate` to `ensure`, should we also change the docstring to reflect that terminology?

straight.el
```diff
@@ -1596,13 +1604,10 @@ confirmation, so this function should only be run after
...
        (error "Branch %S does not exist" ref))
       ((and (null ref) (string= branch cur-branch))
-       (cl-return-from straight-vc-git--validate-head t))
+       (cl-return-from straight-vc-git--ensure-head t))
       ((and (null ref) head-detached-p)
-       (straight-vc-git--popup
-         (format "In repository %S, HEAD is even with branch %S, but detached."
-                 local-repo branch)
-         ("a" (format "Attach HEAD to branch %S" branch)
-          (straight--get-call "git" "checkout" branch))))
```
raxod502: Can we make this only happen automatically if `straight-vc-git-auto-fast-forward` is non-nil? It's a nontrivial change to the state of the repository on disk.

xendk:
Good comments, I've fixed.

raxod502:
Looks like there is a long line somewhere, according to the build failure.

straight.el
```diff
@@ -1284,6 +1284,15 @@ one specified in `straight-vc-git-default-protocol'."
   :type 'boolean
   :group 'straight)
 
+(defcustom straight-vc-git-auto-fast-forward t
+  "Whether to quietly fast-forward when pulling packages.
+This suppresses popups for trivial remote changes (i.e. the
+current HEAD is an ancestor to the remote HEAD).
+Also re-attaches detatced heads quietly when non-nil.
```
raxod502: s/detatced/detached

xendk:
Aaah, the joys of late night pushes... Fixed.

raxod502:
Thanks! This is great.

Issue #251, opened 2018-03-12
Don't use elint


raxod502:
After half a year of usage, I'm unconvinced that factoring out testing functionality into [elint](https://github.com/raxod502/elint) was worth it. I'd like to port this back into `straight.el` directly, to reduce complexity and coupling.

See also:
* https://github.com/raxod502/elint/issues/4
* https://github.com/raxod502/el-patch/issues/22

PR #256, opened 2018-03-24
[MERGED] Drop elint

xuchunyang:
An attempt to solve #251.

- For Emacs 24, `make checkdoc` will simply pass
- `make all` won't run `make toc`. It requires `markdown-toc` thus node/npm.


Makefile
```diff
@@ -1,32 +1,39 @@
...
-.PHONY: travis
-travis:
-	@elint/elint checkdoc compile longlines
-	@mkdir -p ~/.emacs.d/straight/repos/
-	@ln -s $(PWD) ~/.emacs.d/straight/repos/
-	@emacs --batch -l ~/.emacs.d/straight/repos/straight.el/bootstrap.el \
-	--eval "(straight-use-package 'use-package)" \
-	--eval "(use-package el-patch :straight t)"
+ELS = $(wildcard *.el)
+ELCS = $(ELS:.el=.elc)
```
raxod502: I believe best practice says to use `:=` rather than `=` (see [manual](https://www.gnu.org/software/make/manual/html_node/Flavors.html#Flavors)).

Makefile
```diff
@@ -1,32 +1,39 @@
...
-	@elint/elint compile
+%.elc: %.el
+	$(EMACS) -Q --batch -L . -f batch-byte-compile $<
 
-.PHONY: longlines
-longlines:
-	@elint/elint longlines
+checkdoc:
+	@$(EMACS) -Q --batch \
+	--eval "(or (fboundp 'checkdoc-file) (kill-emacs))" \
```
raxod502: Given that Emacs 24.4 is supported by `straight.el`, I think it would be best to allow checkdoc to be run using such versions.

Makefile
```diff
@@ -1,32 +1,39 @@
...
 toc:
-	@elint/elint toc
+	markdown-toc -i README.md
 
-.PHONY: clean
 clean:
-	@rm -f *.elc
+	rm -f *.elc
+
+travis: checkdoc compile longlines
```
raxod502: I think `compile` should come before `checkdoc` here. That way, when the developer just runs `make`, it is more likely to produce a problem they can fix right away (probably they will want to fix bugs in the code before updating the documentation).

Makefile
xuchunyang: Fixed.

Makefile
xuchunyang: Fixed.

I didn't add it because it is not very easy to embedded a long sexp in Makefile.

Makefile
xuchunyang: Fixed.

Makefile
```diff
@@ -1,32 +1,49 @@
...
-longlines:
-	@elint/elint longlines
+%.elc: %.el
+	$(EMACS) -Q --batch -L . -f batch-byte-compile $<
+
+checkdoc:
+	@$(EMACS) -Q --batch \
+	--eval "$$checkdoc_file" \
+	--eval "(setq sentence-end-double-space nil)" \
+	--eval '(checkdoc-file "straight.el")'
```
raxod502: Previously, the build failed if any checkdoc warnings were emitted. That is no longer the case, so the Travis build won't catch these errors. I think that needs to be fixed. (The behavior of elint was to fail if any output was emitted; the annoyance of doing that in a Makefile was part of the reason for making a general framework for the task.)

Makefile
```diff
@@ -1,32 +1,49 @@
...
+
+longlines: $(ELS) README.md Makefile .travis.yml
+	@echo "$^" | xargs sed '/<!-- toc -->/,/<!-- tocstop -->/d' \
+	| sed '/longlines-start/,/longlines-stop/d' \
+	| grep -E '.{80}'                           \
+	| grep -E -v '\[.+\]: (#|http)'             \
+	| sed 's/^/long line: /'
```
raxod502: Does this return non-zero if long lines are found?

Makefile
```diff
@@ -1,32 +1,49 @@
...
+	--eval "(use-package el-patch :straight t)"
+
+define checkdoc_file
+(unless (fboundp 'checkdoc-file)
+  (defun checkdoc-file (file)
+    "Check FILE for document, comment, error style, and rogue spaces."
+    (with-current-buffer (find-file-noselect file)
+      (let ((checkdoc-diagnostic-buffer "*warn*"))
+        (checkdoc-current-buffer t)))))
+endef
```
raxod502: Why not put the other two `--eval` forms from above down here as well? I think that would improve readability.

Makefile
```diff
@@ -1,32 +1,49 @@
...
-.PHONY: compile
-compile:
-	@elint/elint compile
+compile: $(ELCS)
 
-.PHONY: longlines
-longlines:
-	@elint/elint longlines
+%.elc: %.el
+	$(EMACS) -Q --batch -L . -f batch-byte-compile $<
```
raxod502: Similarly to my comment on checkdoc below, we need to fail the build if byte-compilation warnings are emitted.

Makefile
xuchunyang: It seems checkdoc doesn't support batch mode. I don't know easy way to fix this..

Personally I am fine without treating such warnings as errors. Developers should always run `make` locally to test their changes, and when they do, they will be aware of warnings and errors. I am not willing to convert warnings to errors especially when it is not very easy.

Makefile
xuchunyang: No, it still returns zero.

Makefile
xuchunyang: Good point.

However, I am not sure if simply defining `checkdoc-file` for Emacs 24.5 will work

```el
(defun checkdoc-file (file)
  "Check FILE for document, comment, error style, and rogue spaces."
  (with-current-buffer (find-file-noselect file)
    (let ((checkdoc-diagnostic-buffer "*warn*"))
      (checkdoc-current-buffer t))))
```

it seems the buffer name `"*warn*"` is special for Emacs 25's checkdoc.el.

Makefile
xuchunyang: Fixed.

Makefile
raxod502: Sorry, but checkdoc warnings must fail the build. Experience has shown me that if warnings don't _have_ to be fixed, then people won't fix them.

xuchunyang:
I have no clue why Travis is failing. I can't understand the error message and also am unable to reproduce locally.

> https://travis-ci.org/raxod502/straight.el/jobs/357995663#L861
>
> In toplevel form:
> straight-x.el:92:8:Error: ‘(recipe (pop straight-x-waiting))’ is a malformed function
> make: *** [straight-x.elc] Error 1

Makefile
raxod502: I did some testing and you are indeed right that checkdoc was simply not working with Emacs 24.5 and previously. In light of that, I'd be fine with just doing `(kill-emacs)` as you suggested, since we still fail the build if checkdoc emits errors/warnings on newer versions.

raxod502:
You can reproduce it locally if you use Emacs version 25.3 or previously, and make sure to byte-compile `straight.el` before `straight-x.el`. The issue is that when `straight.el` is compiled, it evaluates compatibility definitions of `when-let*` and other functions, but those definitions persist only for the duration of that Emacs session due to the use of `eval-when-compile`. Hence when `straight-x.el` is byte-compiled separately, `when-let*` is undefined, but we don't see an undefined function error because of eager macroexpansion, which causes the arguments of the function-that-isn't-known-to-be-a-macro to be investigated as regular function calls, which they aren't. Replacing the `eval-when-compile` forms relevant to `when-let*` with `eval-and-compile` solves the problem, and I have no objection to this.

See also simplifying the whole matter in #259.

xuchunyang:
I see, thanks for the explanation. When I am trying to reproduce, `straight-x.el` is being compiled before `straight.el` unlike Travis.

```
~/.emacs.d/straight/repos/straight.el $ make compile --dry-run
emacs -Q --batch -L . --eval "(setq byte-compile-error-on-warn t)" \
	-f batch-byte-compile bootstrap.el
emacs -Q --batch -L . --eval "(setq byte-compile-error-on-warn t)" \
	-f batch-byte-compile install.el
emacs -Q --batch -L . --eval "(setq byte-compile-error-on-warn t)" \
	-f batch-byte-compile straight-x.el
emacs -Q --batch -L . --eval "(setq byte-compile-error-on-warn t)" \
	-f batch-byte-compile straight.el
```

xuchunyang:
Now if `checkdoc`/`longlines` gives output, `make` will fail. And for `compile`, warnings are treated as errors, so if there is any warning, `make` will fail.

Makefile
```diff
@@ -1,32 +1,40 @@
...
+all: compile checkdoc longlines
+
+compile: $(ELCS)
 
-.PHONY: compile
-compile:
-	@elint/elint compile
+%.elc: %.el
+	$(EMACS) -Q --batch -L . --eval "(setq byte-compile-error-on-warn t)" \
+	-f batch-byte-compile $<
```
raxod502: The build still passes with byte-compile warnings (e.g. unused lexical argument) for me.

xuchunyang: @raxod502 It looks like a bug of `byte-compile-error-on-warn`. The following uses Emacs 27 and gets the expected result. But If I uses 25.3, it does pass.

```
~/.emacs.d/straight/repos/straight.el $ git diff | cat
diff --git a/straight.el b/straight.el
index 4853213..c53fd07 100644
--- a/straight.el
+++ b/straight.el
@@ -4687,3 +4687,5 @@ is loaded, according to the value of
 ;; checkdoc-symbol-words: ("top-level")
 ;; outline-regexp: ";;;;* "
 ;; End:
+
+(defun foo (x))
~/.emacs.d/straight/repos/straight.el $ make compile
emacs -Q --batch -L . --eval "(setq byte-compile-error-on-warn t)" \
	-f batch-byte-compile straight.el

In toplevel form:
straight.el:4691:1:Error: Unused lexical argument ‘x’
make: *** [straight.elc] Error 1
~/.emacs.d/straight/repos/straight.el $
```

xuchunyang:
> I think we can also delete .projectile, as its only purpose was to ignore elint.

Done.

> This will also need to rebased to take straight-x.el explicitly into account (e.g. disable checkdoc).

Currently `checkdoc` checks just `straight.el`.

Makefile
raxod502: Well, in that case we need to work around the bug, at least in one of the versions that is tested. Otherwise, warnings won't fail the build.

raxod502:
I made a few adaptations to your work and merged it. Thanks for the help!

PR #257, opened 2018-03-24
Fetch multi repos asynchronously

xuchunyang:
For #103.

This provides a new  command `straight-x-fetch-all`, it calls `git fetch` within all repos asynchronously. The interface is still rough.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,1 @@
-/bootstrap.elc
-/install.elc
-/straight.elc
+*.elc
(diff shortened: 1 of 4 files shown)


raxod502:
Thanks! I added a package header and requires, fixed a typo, and updated the recipe for `straight.el` to include this new file.

I like the idea of having an extensions file where things can be shared and tested before being fully documented, tested, linted, etc.

Issue #258, opened 2018-03-25
Simplify function backports


raxod502:
Currently, we have 163 lines of function definitions copied from new Emacs versions, split into six different `eval-when-compile` forms so that all the dependencies work correctly during byte-compilation. This is very complex and brittle. Instead, we could use much simpler definitions, given that we know how the functions are used in `straight.el` (no need to account for all the weird edge cases, nor provide exactly the same feature-set). For example, instead of the huge setf definition for `alist-get`, we could just use something like [`radian-alist-set`](https://github.com/raxod502/radian/blob/2f900981077d8c2365a9f0f854c54ec0155e957a/radian-emacs/radian-util.el#L9-L31).

dieggsy:
@raxod502 I could look into this. I think it may be worth just getting rid of some of the backports (for example, `thread-first` is _nice_ but pretty much completely unnecessary, maybe even `if-let*` as well). What do you think?

dieggsy:
I feel more strongly about `thread-first`, which is only used in one place.

raxod502:
See also https://github.com/raxod502/straight.el/commit/98b587979baa5738fd9fd9806650411549eb23b7.

PR #259, opened 2018-03-25
Simplify backports

dieggsy:
Addresses #258. As I mentioned there, we could also do without `if-let*` and `when-let*`, but those are used much more extensively, so waiting for your opinion on that.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -147,14 +147,9 @@ VARLIST is the same as in `if-let*'."
   (when (version< emacs-version "25.1")
 
     ;; Definition from Emacs 25.3, subr.el
-    (defun alist-get (key alist &optional default remove)
+    (defun alist-get (key alist &optional default)
       "Return the value associated with KEY in ALIST, using `assq'.
-If KEY is not found in ALIST, return DEFAULT.
-
-This is a generalized variable suitable for use with `setf'.
-When using it to set a value, optional argument REMOVE non-nil
-means to remove KEY from ALIST if the new value is `eql' to DEFAULT."
-      (ignore remove) ;;Silence byte-compiler.
+If KEY is not found in ALIST, return DEFAULT."
       (let ((x (assq key alist)))
         (if x (cdr x) default)))))
 
--- a/straight.el
+++ b/straight.el
@@ -102,36 +102,7 @@
                 bindings)))))
 
 (eval-when-compile
-  (when (version< emacs-version "25.1")
-
-    ;; Definition from Emacs 25.3, subr.el
-    (gv-define-expander alist-get
-      (lambda (do key alist &optional default remove)
-        (macroexp-let2 macroexp-copyable-p k key
-          (gv-letplace (getter setter) alist
-            (macroexp-let2 nil p `(assq ,k ,getter)
-              (funcall do (if (null default) `(cdr ,p)
-                            `(if ,p (cdr ,p) ,default))
-                       (lambda (v)
-                         (macroexp-let2 nil v v
-                           (let ((set-exp
-                                  `(if ,p (setcdr ,p ,v)
-                                     ,(funcall setter
-                                               `(cons (setq ,p (cons ,k ,v))
-                                                      ,getter)))))
-                             (cond
-                              ((null remove) set-exp)
-                              ((or (eql v default)
-                                   (and (eq (car-safe v) 'quote)
-                                        (eq (car-safe default) 'quote)
-                                        (eql (cadr v) (cadr default))))
-                               `(if ,p ,(funcall setter `(delq ,p ,getter))))
-                              (t
-                               `(cond
-                                 ((not (eql ,default ,v)) ,set-exp)
-                                 (,p ,(funcall setter
-                                               `(delq ,p ,getter)))))))))))))))
-
+  (when (version< emacs-version "25.3")
     ;; Defined by Emacs 25.3 in C source code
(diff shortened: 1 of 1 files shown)


straight.el
```diff
@@ -3970,10 +3909,10 @@ This function also adds the recipe repository to
...
 This puts it in `straight-recipe-overrides', depending on the
 value of `straight-current-profile'."
-  (setf (alist-get
-         (car melpa-style-recipe)
-         (alist-get straight-current-profile straight-recipe-overrides))
-        (cdr melpa-style-recipe)))
+  (straight--alist-set
+   (car melpa-style-recipe)
+   (cdr melpa-style-recipe)
+   (alist-get straight-current-profile straight-recipe-overrides)))
```
raxod502: `straight--alist-set` does not actually mutate the original alist. You have to re-assign it to the original place. That will probably be a little inconvenient here because of the nested `alist-get` usage, but not too bad.

raxod502: Also, if the keys are symbols (which I believe they are in this case), it's more efficient to pass `'symbol` as the fourth argument to `straight--alist-set`, so that `eq` is used for comparison.

straight.el
```diff
@@ -4627,7 +4566,10 @@ is loaded, according to the value of
        (fmakunbound 'use-package-handler/:straight)
        (when (and (boundp 'use-package-defaults)
                   (listp use-package-defaults))
-         (setf (alist-get :straight use-package-defaults nil 'remove) nil)))))
+         (unless (alist-get :straight use-package-defaults)
+           (setq use-package-defaults
+                 (delq (assoc :straight use-package-defaults)
+                       use-package-defaults)))))))
```
raxod502: I think it's better to use `assq-delete-all` for this.

straight.el
```diff
@@ -4670,8 +4612,10 @@ is loaded, according to the value of
            #'straight-use-package--straight-handler)
          (when (and (boundp 'use-package-defaults)
                     (listp use-package-defaults))
-           (setf (alist-get :straight use-package-defaults)
-                 '('(t) straight-use-package-by-default))))))))
+           (straight--alist-set
+            :straight
+            '('(t) straight-use-package-by-default)
+            use-package-defaults)))))))
```
raxod502: Same as above, this needs an extra `setq`.

raxod502: We can also use `'symbol` here.

raxod502:
This doesn't have to be addressed in this pull request, but I think the compatibility definitions of `alist-get`, `if-let*`, and `when-let*` could be simplified because we don't use many of the features, including the `REMOVE` argument to `alist-get`, the ability of `if-let*` and `when-let*` to take nonstandard binding vectors, etc.

Plus it would probably be smarter to condition on `fboundp` rather than checking the version of Emacs.

dieggsy:
I've rebased the appropriate commit with your suggestions (the `straight-override-recipe` one is a bit convoluted, not sure if I got it right) and added an alist-get simplification. Haven't looked in depth into the `internal--*` functions and `if-get*` logic yet though.

straight.el
```diff
@@ -208,14 +147,13 @@ VARLIST is the same as in `if-let*'."
...
 
     ;; Definition from Emacs 25.3, subr.el
-    (defun alist-get (key alist &optional default remove)
+    (defun alist-get (key alist &optional default)
       "Return the value associated with KEY in ALIST, using `assq'.
 If KEY is not found in ALIST, return DEFAULT.
 
 This is a generalized variable suitable for use with `setf'.
 When using it to set a value, optional argument REMOVE non-nil
 means to remove KEY from ALIST if the new value is `eql' to DEFAULT."
```
raxod502: Probably remove reference to `REMOVE` from the docstring too.

straight.el
```diff
@@ -4627,7 +4571,9 @@ is loaded, according to the value of
        (fmakunbound 'use-package-handler/:straight)
        (when (and (boundp 'use-package-defaults)
                   (listp use-package-defaults))
-         (setf (alist-get :straight use-package-defaults nil 'remove) nil)))))
+         (unless (alist-get :straight use-package-defaults)
```
raxod502: Why do we have an `unless` here? It seems like it would cause the `setq` below to never be executed.

straight.el
dieggsy: Good point. Removing the setf bit as well, since we removed the gv- definition

straight.el
dieggsy: Because of this, I think:
```
When using it to set a value, optional argument REMOVE non-nil
means to remove KEY from ALIST if the new value is eql to DEFAULT.
```
So as far as I understand the previous code was simply removing that key if the value of `:straight` was `nil`. The unless makes sure the value isn't something other than nil before removing it.

raxod502: No, I think "new value" in `if the new value is eql to DEFAULT` means the value that's provided to `setf`, not the value in the alist. Certainly I did not intend to condition removal on the value being nil (since it was never supposed to be nil).

dieggsy: Ah, ok, I think I misunderstood the docs. So to be clear, you basically whant to always remove the `:straight` key from `use-package-defaults` within the enclosing `when` statement? So removing the `unless` should be enough.

raxod502: Yes, your assessment is correct.

dieggsy: Ok, I've made the change.

raxod502:
Thanks!

PR #261, opened 2018-03-27
Don't use when-let* in straight-x.el

xuchunyang:
`when-let*` is not available in `straight-x.el`, so don't use it, otherwise `make compile` will fail for Emacs versions less than 26.


raxod502:
Better solution is https://github.com/raxod502/straight.el/commit/98b587979baa5738fd9fd9806650411549eb23b7, I think.

Issue #276, opened 2018-05-15
Windows - full rebuild on every init due to "changes" in Emacs version


sivertsenstian:
Not sure why this is happening, my current emacs version stays the same when I print it:
![image](https://user-images.githubusercontent.com/1849292/40047442-c5cfeee6-582f-11e8-85cf-6ad468b339fe.png)

Not sure how I check what straight gets as `last-emacs-version ` when it compares the versions:
https://github.com/raxod502/straight.el/blob/ac4cfb80fc0ce59ee3a2e8d6aebf100731a44e0a/straight.el#L2528

In messages I get (on every startup)

>Rebuilding all packages due to change in Emacs version
>Building straight...done
>Building use-package...
>Building use-package → Building bind-key...done
>Building use-package...done
>Building evil...
>Building evil → Building undo-tree...done
>Building evil → Building goto-chg...done
>Building evil...done
>Building evil-leader...done
>Building evil-escape...done
>Building evil-exchange...done
>Building evil-matchit...done
>Building evil-snipe...done
.... etc.

> init completed in 114424.79ms 😢 

I tried adding
```
(setq straight-cache-autoloads t
      straight-check-for-modifications 'never)
```
To the top of my init.el (after the straight bootstrap snippet), to try and work around it, but did not seem to help.

Using emacs from https://ftp.gnu.org/gnu/emacs/windows  
the latest .zip

Not sure if this is an issue with my emacs. straight or my config.. If you have any idea why this is happening or what I can do to investigate further I would appreciate it.

janoc:
I see this same problem in windows too. No such issues on Mac or Linux.

janoc:
I have just done a test and this is what I get in my messages window:

Rebuilding all packages due to change in Emacs version

And the last-emacs-version is (sans quotes):

"GNU Emacs 26.0.90 (build 1, x86_64-w64-mingw32)
 of 2017-11-30"

I.e. there is CRLF between the two lines.

And here is my build-cache.el:
https://pastebin.com/AsJg9uEZ

raxod502:
Are you sure they are the same? If they were, then presumably the `equal` check would not fail. I think that one of them has a CRLF, and the other only has an LF.

I looked at your build cache, and indeed it has the CRLF as you said. I guess we need to decide whether the build cache should have a CRLF or an LF, and change the other string that is compared accordingly.

janoc:
I am sure it was CRLF and not just LF because the Messages window was showing the ^M character there. 

[...]

PR #277, opened 2018-05-17
Newline normalization hack for #276

janoc:
This is a quick fix for the windows newline issue, at least until a better solution could be found.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -263,6 +263,11 @@ that may contain `straight--not-present' as a value."
 
 ;;;;; Strings
 
+(defun straight--strip-newlines (string)
+  "Strips CR, LF, CRLF newlines from STRING.
+It is useful for normalizing strings before comparison."
+  (replace-regexp-in-string "[\n\r]" "" string 'fixedcase 'literal))
+
 (defun straight--split-and-trim (string &optional indent max-lines)
   "Split the STRING on newlines, returning a list.
 Remove any blank lines at the beginning or end. If INDENT is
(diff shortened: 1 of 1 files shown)


straight.el
```diff
@@ -2453,6 +2453,11 @@ straight.el knows to regenerate the whole cache.")
 If this is unchanged between loading and saving the build cache,
 then the saving step is skipped for efficiency.")
 
+(defun straight--strip-newlines (str)
+  "Strips CR, LF, CRLF newlines from strings, useful for
+normalizing strings before comparison."
+  (replace-regexp-in-string "[?\n|?\r]" "" str t t))
+
```
raxod502: Could you put this function in the section `Utility functions > Strings`, around line 307? I think it would make more sense there.

raxod502: This regexp also strips question marks from strings. Perhaps you meant to use `[\n\t]` instead.

raxod502: Instead of `t t`, I think the style `'fixedcase 'literal` is more clear, since it indicates what the purpose of the arguments is.

janoc: Sure, moved it. 
Re regexp - *facepalm*. Yup, you are right, also the bar shouldn't be there. Got confused by the Emacs regexp syntax, I am not used to Lisp so much yet.

janoc: Good point, fixed.

janoc:
I have updated the PR with the fixes, as requested.

raxod502:
This should not have been merged to master. I force-pushed and cherry-picked to develop instead.

This mistake has happened a number of times, so I have changed the default branch of this repository to develop.

raxod502:
Anyway, thanks for your contribution!

janoc:
Ah sorry, I didn't realize you had a develop branch. Thanks for merging it, though.

raxod502:
Agreed. While I usually refrain from pushing things to master until they have been in develop for long enough to spot regressions, it seems like in this case the improvement in usability is worth the risk. I added the commit to master.

bboytian:
Was this hot fix removed in future commits? I cant seem to find it in master anymore

raxod502:
It's https://github.com/radian-software/straight.el/commit/62c9ee7583070514320edc6d78be9e7e56ef2822, which the GitHub interface indicates is included in master and by extension develop.

However, the variable is no longer needed because in https://github.com/radian-software/straight.el/commit/cb7679cee0ea05b589607de9080e9d1b144c92bb we moved from checking `(emacs-version)` to `emacs-version`, and the latter does not have any newlines in it.

PR #284, opened 2018-06-15
Support for git clone --depth

renatofdds:
This adds support for the git clone --depth argument, limiting the history size fetched on initial repository clone, helping keep repositories with huge history small on disk. The depth can be by package with ```:depth``` on recipe and through the default custom variable ```straight-vc-git-default-clone-depth``` (which is set to full history by default). Travis CI uses [50](https://docs.travis-ci.com/user/customizing-the-build#Git-Clone-Depth) for example.

I was able to reduce disk usage from 700MB to 250MB on my configuration setting default depth to 10.

Note that github doesn't support advertized fetch refs (```git fetch --depth=<depth> <commit>```), so although i added support for checking if commit is in history for locked version checkout, this doesn't work for github.

I believe setting depths only for repos which contribute the most to total size  to avoid unfound commits is a good way to manage total size. I made a little snippet to investigate them if you should consider putting it on help:

```sh
(
	for REPO in ~/.emacs.d/straight/repos/*; do
		SIZES=$(git --git-dir=$REPO/.git count-objects -vH |
					grep 'size' |
					grep -oE '[0-9.]+.*' |
					tr '\n' '\t' |
					sed 's/ bytes//g; s/ KiB/K/g; s/ MiB/M/g; s/ GiB/G/g; s/\t$//');
		TOTAL=$(numfmt -d $'\t' --from=auto --to=none --field=1,2,3 $SIZES |
					tr $'\t' '+' |
					bc);
		printf "%s\t%s\t%s\n" $REPO $(numfmt --to=si $TOTAL) $SIZES;
	done
) | sort -t $'\t' -k2rh | column -t -s $'\t' -N repository,total,size,size-pack,size-garbage
```


straight.el
```diff
@@ -654,7 +654,8 @@ If the command fails, throw an error."
   "Call COMMAND with ARGS, returning its stdout and stderr as a string.
 Return a string with whitespace trimmed from both ends. If the
 command fails, throw an error."
-  (string-trim (apply #'straight--get-call-raw command args)))
+  (string-trim
+   (apply #'straight--get-call-raw command (delete nil (delete "" args)))))
```
raxod502: This smells like a hack to me. We should not be passing nil or empty-string arguments to commands in the first place.

README.md
```diff
@@ -507,6 +507,7 @@ set of additional directives. For example, the `git` backend accepts:
 * `:branch`
 * `:nonrecursive`
 * `:upstream`
+* `:depth`
```
raxod502: This should be documented further, in [the section on the Git backend](https://github.com/renatofdds/straight.el/blob/ada6ca9f89a5f11869e7eeeb83e57288800b558d/README.md#git-backend).

straight.el
```diff
@@ -1078,6 +1079,20 @@ This method simply delegates to the relevant
   "The default value for `:branch' when `:type' is symbol `git'."
   :type 'string)
 
+(defcustom straight-vc-git-default-clone-depth 'full
+  "The default value for `:depth' when `:type' is symbol `git'.
+
+The history depth to use when cloning remote repository. Setting
+this to a number reduces repository size by limiting initial
+cloned history. Using values too small may result in straight
```
raxod502: I prefer to style the name of this package as `straight.el` rather than just `straight`.

straight.el
```diff
@@ -1078,6 +1079,20 @@ This method simply delegates to the relevant
...
+being unable to fetch the locked commit for git servers such as
+github which don't support advertised fetch refs.
+
+See git clone --depth manual."
```
raxod502: The allowable values for this variable should be mentioned in the docstring, i.e. a positive integer or the symbol `full`.

straight.el
```diff
@@ -1078,6 +1079,20 @@ This method simply delegates to the relevant
   "The default value for `:branch' when `:type' is symbol `git'."
   :type 'string)
 
+(defcustom straight-vc-git-default-clone-depth 'full
+  "The default value for `:depth' when `:type' is symbol `git'.
+
+The history depth to use when cloning remote repository. Setting
```
raxod502: "when cloning a remote repository"

straight.el
raxod502: Maybe "See the `git-clone` manual page for more information about shallow clones."

straight.el
```diff
@@ -1627,55 +1642,72 @@ COMMIT is a 40-character SHA-1 Git hash. If it cannot be checked
...
 specified in RECIPE instead. If that fails, signal a warning."
   (straight--with-plist recipe
-      (local-repo repo host branch upstream nonrecursive)
+      (local-repo repo host branch upstream nonrecursive depth)
     (let ((success nil)
           (repo-dir (straight--repos-dir local-repo))
           (url (straight-vc-git--encode-url repo host))
-          (branch (or branch straight-vc-git-default-branch)))
+          (branch (or branch straight-vc-git-default-branch))
+          (history-depth (or depth straight-vc-git-default-clone-depth)))
```
raxod502: It seems less fragile to use `depth` as the name of the variable rather than `history-depth`.

straight.el
```diff
@@ -1627,55 +1642,72 @@ COMMIT is a 40-character SHA-1 Git hash. If it cannot be checked
...
+					(straight--get-call
+					 "git" "remote" "add"
+					 straight-vc-git-upstream-remote url)
+					(straight--get-call
+					 "git" "fetch" straight-vc-git-upstream-remote)))))
+			(setq success t))
+		;; Make cloning an atomic operation.
+		(unless success
+		  (when (file-exists-p repo-dir)
+			(delete-directory repo-dir 'recursive)))))))
```
raxod502: Please turn off `indent-tabs-mode`. I pushed a new commit that adds a file-local variable for this.

I'll have more comments on this section when the diff renders without whitespace changes.

raxod502:
See also #2.

raxod502:
This issue is being closed automatically since it has a "waiting on response" label but hasn't had any activity in 90 days. But don't worry: if you have some information to help move the discussion along, just leave a comment and I'll re-open the issue.

*(Comment generated automatically by [NO CARRIER](https://github.com/twbs/no-carrier) via [tidier](https://github.com/raxod502/tidier).)*

Issue #287, opened 2018-06-25
delete-directory: modified directory doesn't exist


kaiwk:
After recent updating, restart emacs gives:

> delete-directory: Opening directory: No such file or directory, /my-home-path/.emacs.d/straight/modified

environment:
emacs 25

I also tested it on emacs24.5 and emacs26.1, it seems it would give this error on emacs 24.5 and emacs 25, but it works fine on emacs 26.

raxod502:
Great! Thanks for the report and solution.

PR #288, opened 2018-06-25
[#287] Fix modified directory doesn't exist

kaiwk:


raxod502:
Thanks for the contribution. I removed the `file-exists-p` check in favor of a `condition-case` to suppress the error: that should speed up startup by a few precious nanoseconds :)

PR #296, opened 2018-07-20
el-get can also specify git revisions

vincentbernat:
`el-get` allows to specify a git revision, a branch or a tag. For example:

```lisp
(setq el-get-sources
      '((:name org-mode
               :checkout "release_9.1.13")
        (:name cider
               :checkout "v0.17.0")
        (:name clojure-mode
               :checkout "5.8.1")
        (:name bbdb
               :checkout "c951e15cd01d")
        (:name magit
               :checkout "2.13.0")))
```

Note I may have misunderstood the point. As a long time el-get user, other points seem valid to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1155,10 +1155,6 @@ offer corrections for this section.
 
 #### Advantages of `straight.el`
 
-* `straight.el` has integrated support for selecting particular Git
-  revisions of packages. This process is more manual in el-get, as it
-  requires placing the commit hash into the recipe, which disables
-  updates.
 * `straight.el` uses your init-file as the sole source of truth for
   package operations. el-get has additional metadata stored outside
   the init-file, although specifying all packages in your init-file is


raxod502:
It seems that what I wrote in the README was entirely correct: in el-get, you have to place the commit hash into the recipe, which disables updates.

Perhaps what was confusing was how the functionality of `straight.el` differs? The difference is that `straight.el` natively supports this specific use case, and allows you to automatically write the revisions of all packages to an external lockfile with no changes to your init-file (and then load those revisions next time packages are installed on a different machine).

The reason that this difference is important is that it means that with `straight.el` you can get perfect reproducibility just by writing a lockfile, whereas with el-get it seems that you have to manually record a recipe for every package so you can write down the revisions, including dependencies. If this is erroneous, please correct me—I never used el-get.

No objections to making the text more clear, but removing that bullet point doesn't seem quite right to me.

raxod502:
Actually, never mind. I see that the things I said above are already covered in the third bullet point, so the first bullet point is just redundant and confusing.

Thanks for your contribution.

Issue #274, opened 2018-05-08
Configurable Paths


nickgarber:
I'd like to be able to specify the locations for the 

1. build dir, 
2. repo dir, and 
3. versions dir 

... ideally via the customize system.

raxod502:
I was wondering when somebody would ask about this. I have no objection to it, except it will take a little bit of work to make it work correctly with the bootstrap system.

raxod502:
Out of curiosity, what's your use case?

nickgarber:
There are a few reasons this would make a difference for me, but the one that is front of mind is to 

... It'd support moving from my manual submodule-based config to use-package + straight.

In general it's a blend of aesthetics with data separation (to isolate personal info from my emacs.d). If it helps to illustrate, I'd set the repo-dir to where my submodules already live, the build-dir to the dir I already have setup as an `init.d` style directory, and the versions-dir to the directory where I store entirely personal info, like my custom-file, which I like to use extensively.

raxod502:
Interesting. As I said, no objection here to supporting this directly. As a workaround, of course, you can simply symlink `~/.emacs.d/straight/repos`, `~/.emacs.d/straight/build`, etc. to your desired paths. I think that should work just as well.

aqua36:
This would be useful for me as well. My use case is the following: I have my init.el and every other config file that is not a package in a separate directory plus a make file that creates an .emacs.d/init.el if it doesn't exist with a single line loading the init.el from my other dir. That way I keep only my custom dir under version control and it remains clean of anything emacs creates automatically. If you implemented this enhancement I would be able to set the versions dir to be inside my custom dir instead of messing with symlinks

nickgarber:
I hacked-up a basically working (though messy) example here, https://github.com/nickgarber/straight.el/commit/b3f7838c8afc4acc1cb167cd32dbd349a6ae88ef

@raxod502 Any advice on how to clean this up or generalize it into something that'd make for a decent PR?

Biggest surprise in this process - I anecdotally found that Github's caching system can take up to 30 minutes to reflect updates. This was inspired great head-scratching until properly understood.

raxod502:
Thanks. I'll take a look at this as soon as I have time (this may be a little while because I'm traveling to New York for the week). Why don't you go ahead and make it a pull request right away, so that the code review features work better? (We can always rebase it later to fix up the hacky parts.)

[...]

PR #300, opened 2018-08-05
[wip] configurable paths

nickgarber:
relates to https://github.com/raxod502/straight.el/issues/274

* Goals:

- [ ] determine optimal location and handling for the override variables `straight--project-base-url` and `straight--base-dir`
- [x] enable use of a package like https://github.com/emacscollective/no-littering
- [ ] support configurable paths, (without requiring the project be forked)
- [ ] (optionally) allow for different paths per profile. [1]

[1] I'm interested in having a full-featured 'workstation' profile, a 'minimal' profile, and a 'presentation' profile. By separating the paths it would support very stable configurations even while switching between profiles, even potentially running multiple profiles simultaneously.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1344,17 +1344,30 @@ care of all these details for you:
 
 <!-- longlines-start -->
 
-    (let ((bootstrap-file
-           (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
-          (bootstrap-version 4))
-      (unless (file-exists-p bootstrap-file)
-        (with-current-buffer
-            (url-retrieve-synchronously
-             "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
-             'silent 'inhibit-cookies)
-          (goto-char (point-max))
-          (eval-print-last-sexp)))
-      (load bootstrap-file nil 'nomessage))
+    ;; optional customization
+    ;(defvar straight--project-base-url "https://raw.githubusercontent.com/<example>/straight.el/")
+    ;(defvar straight--base-dir (concat user-emacs-directory "some/alternative/path"))
+
+    ;; bootstrap
+    (let*
+      ((straight--project-base-url (if (bound-and-true-p straight--project-base-url)
+                                         straight--project-base-url
+                                       "https://raw.githubusercontent.com/raxod502/straight.el/"))
+       (straight--base-dir (if (bound-and-true-p straight--base-dir)
+                               straight--base-dir
+                             (concat user-emacs-directory "straight/")))
+       (bootstrap-file (concat straight--base-dir "repos/straight.el/bootstrap.el"))
+       (bootstrap-version 4))
+    (unless (file-exists-p bootstrap-file)
+      (progn
+        (require 'load-mode__straight-install)
+      (with-current-buffer
+          (url-retrieve-synchronously
+           (concat straight--project-base-url "develop/install.el")
+           'silent 'inhibit-cookies)
+        (goto-char (point-max))
+        (eval-print-last-sexp)))
+    (load bootstrap-file nil 'nomessage))
 
 <!-- longlines-stop -->
 
--- a/bootstrap.el
+++ b/bootstrap.el
@@ -112,7 +112,7 @@
 
 ;; Then we register (and build) straight.el itself.
 (straight-use-package `(straight :type git :host github
-                                 :repo "raxod502/straight.el"
+                                 :repo "nickgarber/straight.el"
                                  :files ("straight*.el")
                                  :branch ,straight-repository-branch))
 
--- a/install.el
+++ b/install.el
@@ -65,6 +65,19 @@
 (progn
   (message "Bootstrapping straight.el...")
 
+  (defcustom straight--project-base-url
+    (if (bound-and-true-p straight--project-base-url)
+        straight--project-base-url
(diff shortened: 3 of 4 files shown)


raxod502:
There are a lot of things to talk about here. Before getting into all of them, I'd like to make sure we have the right overall approach. Here are some of my questions:

* Why do we have `straight--project-base-url`? What use case does it provide for which is not already served by the existing bootstrap mechanism plus `straight-recipe-overrides`?
* I think the base `straight.el` directory must default to `~/.emacs.d/straight`; changing it would break literally everyone's configuration. (I use `no-littering` myself, actually, and follow its conventions in the other Elisp packages I maintain—but I don't think it makes sense for a package manager to do the same; note that `package.el` uses `~/.emacs.d/elpa` and that `no-littering` doesn't attempt to change `package-user-dir`.)
* We should have user options of the form `straight-dir` (or `straight-base-dir` perhaps), `straight-repos-dir`, `straight-build-dir`, etc. The default values should be relative paths (`straight`, `repos`, `build`, etc.), since we can rely on `expand-file-name` to expand them relative to `user-emacs-directory`, etc.

And sorry about the `straight-compat` thing. I understand what's going on, and how to fix it. (I think the easiest solution is to just merge `straight-compat.el` directly into `straight.el`.) The issue was introduced when I first split out `straight-compat`, but I haven't had to bump the bootstrap version since then, so we didn't see the error.

> I'm interested in having a full-featured 'workstation' profile, a 'minimal' profile, and a 'presentation' profile.

This should be easily possible by just setting the `straight-dir` user option differently before bootstrap, the way I see it. This "profile" switching would be entirely different from the existing profile system in `straight.el`, by the way—and for good reason; the existing profile system is intended for use of multiple profiles simultaneously in the same Emacs session.

raxod502:
`straight-compat.el` has been removed, so changes to the installation bootstrap should now work as expected.

nickgarber:
Just a note to say I'm still around, just extra busy and expect to be able
to check in on this in early October.

raxod502:
This thread is being closed automatically since it has a "waiting on response" label but hasn't had any activity in 133 days. But don't worry: if you have some information to help move the discussion along, just leave a comment and I'll re-open the thread.

*(Comment generated automatically by [NO CARRIER](https://github.com/twbs/no-carrier) via [tidier](https://github.com/raxod502/tidier).)*

nickgarber:
> * Why do we have `straight--project-base-url`? What use case does it provide for which is not already served by the existing bootstrap mechanism plus `straight-recipe-overrides`?

`straight--project-base-url` isn't intended to stick around. It was just a ladder to allow me to make sure I was using the configurations from my fork. Though I am curious if a person wanted to maintain their own distinct fork of `straight.el`, would the functionality of `straight-recipe-overrides` be sufficient?


> * I think the base `straight.el` directory must default to `~/.emacs.d/straight`; changing it would break literally everyone's configuration. (I use `no-littering` myself, actually, and follow its conventions in the other Elisp packages I maintain—but I don't think it makes sense for a package manager to do the same; note that `package.el` uses `~/.emacs.d/elpa` and that `no-littering` doesn't attempt to change `package-user-dir`.)

That makes sense. I agree with you about the defaults, no change is requested or desired there. 


> * We should have user options of the form `straight-dir` (or `straight-base-dir` perhaps), `straight-repos-dir`, `straight-build-dir`, etc. The default values should be relative paths (`straight`, `repos`, `build`, etc.), since we can rely on `expand-file-name` to expand them relative to `user-emacs-directory`, etc.

For those that want to stray from the default, it seems that adopting something like `straight-base-dir` might be the single biggest change needed to make the `straight.el` directory relocatable. Does that seem true from your perspective and would you be interested in accepting a PR for that change?

This PR is in a very rough state, but I'll chip away at it in hopes it could be distilled into something worth adopting.

raxod502:
> if a person wanted to maintain their own distinct fork of `straight.el`, would the functionality of `straight-recipe-overrides` be sufficient?

Yes, see [the documentation on this](https://github.com/raxod502/straight.el/tree/330de6ad9d2fb9eb48778ab4211fc5e11153fc82#overriding-the-recipe-for-straightel). The only exception (not documented before; now documented) is if you want to change the way in which your own recipe for `straight.el` will be interpreted (this is totally different and way more advanced than simply provided a different recipe from the default). This use case is also supported, but some additional steps are required.

You should be able to test your pull request without modifying any of the code in `straight.el` relating to recipes or bootstrap. The only change necessary is the `bootstrap-file` value in the bootstrap snippet in your init-file—and of course to the `install.el` URL on GitHub, if you are changing the bootstrap snippet. (I think it is okay and preferable to make the user change these things manually, instead of adding more complex logic to the bootstrap snippet.)

> adopting something like `straight-base-dir` might be the single biggest change needed to make the `straight.el` directory relocatable

Yes. I think it would be preferable to add all of the user options I mentioned, but this would be the most complex change (still quite simple, I think).

> would you be interested in accepting a PR

Very much so.

> I'll chip away at it in hopes it could be distilled into something worth adopting.

Cool :)

raxod502:
This thread is being closed automatically by [Tidier](https://github.com/raxod502/tidier) because it is labeled with "waiting on response" and has not seen any activity for 90 days. But don't worry—if you have any information that might advance the discussion, leave a comment and the thread may be reopened :)

Issue #9, opened 2017-01-21
It's slow


raxod502:
My startup time was 0.5s before using `straight.el`; now it's 0.9s.

raxod502:
I think it got slower after the recent refactor. :'(

therockmandolinist:
Can confirm - it's around 1-2 seconds slower for me. (My init time went from 2.5-ish seconds to 4.2-ish)

raxod502:
It's great to hear that it's not actually that much slower. My config has also gotten slower for other reasons that I haven't bothered to fix yet, so I haven't done profiling on `straight.el` specifically since I first added the build caching system.

See also #41, which should eliminate even the 100–300ms slowdown.

By the way, how many packages do you have? (You can get that number by evaluating `(length (hash-table-keys straight--success-cache))`.)

raxod502:
I'm pretty sure the symlinks are negligible. The slowdown should be almost entirely due to the fact that `straight.el` uses `find(1)` to check all of your packages for modifications on init. (With that in mind, you may be impressed that it is so fast!)

therockmandolinist:
I have 173 packages, apparently. (Huh, I should probably revise this and get rid of the cruft) I `:defer t` most of that, I think, with probably around 10-ish packages actually loaded on init, 20-ish deferred with some time.

raxod502:
I believe that you are not in fact imagining things, and `straight.el` really is 1–2 seconds slower. I think the reason you are having trouble attributing it to `straight.el` is that the package modifications check (the part we expect to be slow, and the part that happens all at once) is about 300ms, but every other `straight-use-package` invocation is ever-so-slightly slower than it should be, which adds up to 1–2 seconds.

I'm finally going to get my Emacs config back to a <0.5s init, so I'm investigating this issue now.

raxod502:
Some profiling.

```
(benchmark 10 '(straight-use-package 'cider))

"Elapsed time: 4.143111s (0.600769s in 13 GCs)"

(benchmark 10000000 '(straight--convert-recipe 'cider))

"Elapsed time: 1.680602s"

(let ((recipe (straight--convert-recipe 'cider)))
  (benchmark 100000 '(straight--register-recipe recipe)))

"Elapsed time: 2.212280s (0.308168s in 8 GCs)"

(let ((recipe (straight--convert-recipe 'cider)))
  (benchmark 10000 '(straight--repository-is-available-p recipe)))

[...]

PR #310, opened 2018-09-21
Document symlink setup on MS Windows

qwerbzuio:
Actually symlinks are supported on windows; it's just a bit weird that you have
to assign permission to the user. This patch detects if the user has the ability
to create symlinks and then conditionally activates their usage in straight.

The rant about MS Windows would have to fall victim to this patch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1587,10 +1587,34 @@ recipes which do not explicitly specify a `:no-autoloads` attribute.
 Usually, `straight.el` uses symbolic links ("symlinks") to make
 package files available from the build directory. This happens when
 `straight-use-symlinks` is non-nil, the default. On Microsoft Windows,
-however, there is no support for symlinks, so the default value of
-`straight-use-symlinks` is nil on that platform. That causes copying
-to be used instead, and an advice is placed on `find-file` to cause
-the copied files to act as symlinks if you try to edit them.
+however, support for symlinks is not always available, so the default
+value of `straight-use-symlinks` is nil on that platform. That causes
+copying to be used instead, and an advice is placed on `find-file` to
+cause the copied files to act as symlinks if you try to edit them.
+
+If you want to activate symlink support on MS Windows 7, 8, or 10, the
+following information may be helpful:
+
+* `straight-use-symlinks` has to be set to non-nil manually
+
+* Your user-account needs to be assigned the right to create symbolic
+  links. To do so, run "secpol.msc" and in "Local Policies → User
+  Rights Assignment" assign the right to "Create symbolic links" to
+  your user-account.
+
+* If you have User Account Control (UAC) enabled and your user-account
+  belongs to the the _Administrators_ group you'll need to run Emacs
+  in elevated mode to be able to create symlinks (see
+  [here](https://community.perforce.com/s/article/3472) and
+  [here](https://stackoverflow.com/questions/29063916/win32api-symlink-creation-issue-with-uac-enabled#29065060)
+  and, for an official reference, section Access Token Changes [in
+  this
+  document](https://msdn.microsoft.com/en-us/library/bb530410.aspx#vistauac_topic8)
+
+* [Windows Creators
+  Update](https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/)
+  supports symlink creation without any special permission setup
+
 
 #### Customizing how packages are made available
 
--- a/README.md
+++ b/README.md
@@ -1592,10 +1592,10 @@ value of `straight-use-symlinks` is nil on that platform. That causes
 copying to be used instead, and an advice is placed on `find-file` to
 cause the copied files to act as symlinks if you try to edit them.
 
-If you want to activate symlink support on MS Windows 7, 8, or 10, the
-following information may be helpful:
+If you want to activate symlink-support on MS Windows 7, 8, or 10, you
+should ensure the following requirements:
 
-* `straight-use-symlinks` has to be set to non-nil manually
+* `straight-use-symlinks` has to be set to non-nil manually.
 
 * Your user-account needs to be assigned the right to create symbolic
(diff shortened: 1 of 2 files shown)


raxod502:
This patch requires several disk IO operations at startup to support an infrequently exercised use case only relevant on Windows. I would prefer to avoid this, especially as the workaround (change the value of `straight-use-symlinks` in your init-file) is trivial, intuitive, and well-documented.

We could save this data in the build cache, but then the user would have to manually clear the build cache after updating permissions, which is undesirable.

We could instead check whether symlinks are supported at the time we try to create one, but this has the disadvantage that we can't tell whether to enable `straight-symlink-emulation-mode` at startup.

How about a combination of both, where the value is cached but we recompute it the first time we try to create a symlink in any given session?

If this seems like premature optimization, it may be relevant to note that `straight.el` is highly optimized enough that the disk IO for loading 100 packages (without changes, using live modification detection) is approximately only reading one file from disk, listing one directory, and checking for the existence of 100 other directories. See also https://github.com/raxod502/straight.el/issues/9#issuecomment-337435499 for how expensive disk IO can be.

qwerbzuio:
Thank you for the response. I wasn't aware of the performance considerations involved.

Setting the variable manually is absolutely no problem, of course. The solution you described is maybe a bit too complex, considering that Windows is an infrequently used host system. But I'll let you decide on this as I don't know the code.

Anyway, I find it confusing that the docstring of straight-use-symlinks says that on Windows "you will have to use copying". This is not true. Instead, I think the docu could mention that, on Windows, the user needs to grant link-creation permissions and then set the variable manually. I think this makes it a bit easier for newcomers to get that setting right.

What do you think?

raxod502:
> I find it confusing that the docstring of straight-use-symlinks says that on Windows "you will have to use copying". This is not true. Instead, I think the docu could mention that, on Windows, the user needs to grant link-creation permissions and then set the variable manually. I think this makes it a bit easier for newcomers to get that setting right.

This sounds fine to me, and it's probably the best solution—trying to do the right thing is great, but there is a point when it just adds needless complexity.

I would gladly accept a change to the documentation which makes it clear that symlinks don't work by default, but that you can make them work via the permissions change that you mentioned.

I recall reading somewhere that you also have to disable UAC in order for symlinks to work on Windows 10—is that true? If so, it should be documented.

One reason that I didn't document the possibility of symlinks on Windows is that I am not familiar with the exact circumstances under which they can be used, so I appreciate your help.

qwerbzuio:
On MS Windows the situation is quite diverse - I'm not sure if we have to document the entire mess in the docstring. 

- Symlinks are not available before Windows Vista and on FAT (ok...)
- With UAC disabled it suffices to set the appropriate security policies
- With UAC enabled:
  - users _in the administrators group_ need to run apps in elevated mode to create symlinks (see [here](https://community.perforce.com/s/article/3472) and [here](https://stackoverflow.com/questions/29063916/win32api-symlink-creation-issue-with-uac-enabled#29065060) and, for an official reference, section Access Token Changes [in this document](https://msdn.microsoft.com/en-us/library/bb530410.aspx#vistauac_topic8)
  - users not in the admin group can create symlinks if the security policies are set accordingly
- [Windows Creators Update](https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/) seems to support symlink creation out of the box

I've drafted a reduced docstring to give hints for to the most important steps. Do you think it's enough information?

qwerbzuio:
I've updated the docu. I decided to put the part on setting up MS
Windows mainly into the User Manual and simply link to it from the
docstring.

README.md
```diff
@@ -1587,10 +1587,34 @@ recipes which do not explicitly specify a `:no-autoloads` attribute.
...
-`straight-use-symlinks` is nil on that platform. That causes copying
-to be used instead, and an advice is placed on `find-file` to cause
-the copied files to act as symlinks if you try to edit them.
+however, support for symlinks is not always available, so the default
+value of `straight-use-symlinks` is nil on that platform. That causes
+copying to be used instead, and an advice is placed on `find-file` to
+cause the copied files to act as symlinks if you try to edit them.
+
+If you want to activate symlink support on MS Windows 7, 8, or 10, the
+following information may be helpful:
```
raxod502: How about instead of "… the following information may be helpful", "… you will need to follow these steps" or something like that?

README.md
```diff
@@ -1587,10 +1587,34 @@ recipes which do not explicitly specify a `:no-autoloads` attribute.
...
+
+* `straight-use-symlinks` has to be set to non-nil manually
```
raxod502: I think these bullet-point sentences should all end in periods, to be consistent.

README.md
```diff
@@ -1587,10 +1587,34 @@ recipes which do not explicitly specify a `:no-autoloads` attribute.
...
+  in elevated mode to be able to create symlinks (see
+  [here](https://community.perforce.com/s/article/3472) and
+  [here](https://stackoverflow.com/questions/29063916/win32api-symlink-creation-issue-with-uac-enabled#29065060)
+  and, for an official reference, section Access Token Changes [in
+  this
+  document](https://msdn.microsoft.com/en-us/library/bb530410.aspx#vistauac_topic8)
+
+* [Windows Creators
+  Update](https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/)
+  supports symlink creation without any special permission setup
```
raxod502: Would you mind using the `[link][label]` style, and then putting the URLs down at the bottom with the others? There's an alphabetized list of external links. (This is what caused the Travis build to fail.)

README.md
qwerbzuio: Ok, I chose a different formulation, but I have no strong opinion on that. I'm not a native speaker, so feel free to change it.

README.md
qwerbzuio: There's more than one link list. Hope I got the link position right.

Issue #313, opened 2018-10-01
Hydra for straight.el


aqua36:
I made a simple [hydra](https://github.com/abo-abo/hydra) for this package and [included it](https://github.com/abo-abo/hydra/wiki/straight.el) in the hydra wiki. (It doesn't include absolutely every straight interactive function, because I didn't think some functions are used often enough to justify being included, but you may not share my view on this)

I figured I should let you know, because you might want to include a link to it in the documentation, or utilize it in some other way

raxod502:
Neat, thanks. I don't have time right now to make code changes but I would happily merge a pull request that either links to this from an appropriate place in the documentation (new section for integrations with other packages, possibly near the section on integration with `use-package`?) or includes the code directly in `straight.el`, analogously to how `use-package` integration is handled (in this case some additional review would be necessary), or includes the code in `straight-x.el`, in which case additional review isn't necessary.

raxod502:
Great, no rush :)

aqua36:
I chose to change the documentation. If you think that the code should be included in straight-x.el as well, I'd be happy to make a PR about that.

raxod502:
Where to put the code is up to you. I agree it would not be a good idea to have it in two places. Since we have a pull request now, I'll close this issue. Thanks for your help!

aqua36:
>  Where to put the code is up to you

I'll leave it as it is, then

PR #316, opened 2018-10-08
docs: add section relevant to integration with hydra

aqua36:
This new section was added after the section that mentions integration
with use-package as recommended in #313. I decided that instead of
having "Integration with use-package" and "Integration with other
packages" as two separate sections, one after the other, it would look
"cleaner" if I "promoted" "Integration with other packages", and
included "Integration with use-package" as a subsection of it


aqua36:
done

raxod502:
Thanks very much!

Issue #318, opened 2018-10-15
:fork doesn't work when cloning fresh


andschwa:
I discovered this when my Travis CI build failed after I added a `:fork` to some of my recipes. Such as:

```
(use-package dtrt-indent
  :straight (dtrt-indent :host github :repo "jscheid/dtrt-indent"
                         :fork (:host github :repo "andschwa/dtrt-indent")) ...)
```

In `straight-vc-git-clone`, if `:fork` is present, it doesn't `cd` into the cloned Git repo, and so instead executes `git remote` in `straight/repos` (which is in my `.emacs.d` repo). The first time it does this, it doesn't fail, since you can add the `fork` remote to the repo; but the second time (for another recipe with a fork), it finally fails because `fork` already exists as a remote. Both of these executions are incorrect anyway, because it should be done in the cloned repo of the package.

I think I have a fix, but I'm not having fun figuring out how to use my own fork of `straight.el` without hacks...

andschwa:
Ah ha, figured out how to use my [fork](https://github.com/raxod502/straight.el-support/blob/8805112d6846411920595e7643112d21c7e039f9/benchmark/init.el#L76)... maybe we should put that example in the readme.

dieggsy:
I also just ran into this exact issue a few minutes ago, don't have much to add.

andschwa:
@dieggsy Tests show my fix is working... opening a PR in a minute here 😄

raxod502:
Sorry about this; should be fixed now thanks to your pull request.

PR #319, opened 2018-10-15
Fix bug when cloning with :fork

andschwa:
The shell command to `git remote add fork` needs to happen in the
context of the package's repo, and previously this was happening in
`straight/repos`, which is usually the user's `.emacs.d` repo. The first
time it does this, it doesn't fail, since you can add the fork remote to
the repo; but the second time (for another recipe with a fork), it
finally fails because fork already exists as a remote. Both of these
executions are incorrect anyway, because it should be done in the cloned
repo of the package. Fortunately, fixing this is as easy as moving the
logic down, such that `default-directory` is set to `repo-dir`.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -1974,12 +1974,12 @@ specified in RECIPE instead. If that fails, signal a warning."
             (straight--get-call
              "git" "clone" "--origin" upstream-remote
              "--no-checkout" url local-repo)
-            (when fork-repo
-              (let ((url (straight-vc-git--encode-url fork-repo fork-host)))
-                (straight--get-call "git" "remote" "add" fork-remote url)
-                (straight--get-call "git" "fetch" fork-remote)))
             (let ((straight--default-directory nil)
                   (default-directory repo-dir))
+              (when fork-repo
+                (let ((url (straight-vc-git--encode-url fork-repo fork-host)))
+                  (straight--get-call "git" "remote" "add" fork-remote url)
+                  (straight--get-call "git" "fetch" fork-remote)))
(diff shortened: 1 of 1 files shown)


andschwa:
This resolves #318. It could probably be covered with tests, but I am too tired right now 😩

andschwa:
[Failed build demonstrating bug](https://travis-ci.org/andschwa/.emacs.d/builds/440406841) and [successful build with fix](https://travis-ci.org/andschwa/.emacs.d/jobs/441496452).

andschwa:
For @dieggsy and anyone else that wants to test, add this before your straight.el bootstrap logic:

```
(customize-set-variable
 'straight-recipe-overrides
 '((nil . ((straight
	    :files ("straight*.el") :host github
            ;; Note that `install.el' does not support `:fork'.
            :repo "andschwa/straight.el" :branch "develop")))))
```

raxod502:
Thank you very much for the fix, and I apologize for the fairly atrocious logic error. I have merged it immediately to `master` as well, since it's clearly an important improvement.

raxod502:
Note that I have also pushed a fix to `develop` and `master` which makes `:fork` function as expected during initial installation bootstrap. The fact that it didn't previously was a simple oversight on my part—I forgot to update the recipe schema version counter when I added `:fork` support. (To test, update to the latest version and write a new lockfile.)

I suggest using simply `setq` instead of `customize-set-variable`, but it is up to you. See also #320.

andschwa:
Thanks @raxod502 😄

PR #371, opened 2019-04-06
Fix org-version == "N/A" error

sboehler:
I tried to install org-journal using straight (current develop branch with the org hack), but installation did not succeed due to the following error:

```
Error (use-package): org-journal/:catch: Invalid version syntax: ‘N/A’ (must start with a number)
```

The issue seems to be that org-mode assigns the result of `(org-version)` statically to [the constant with the same name as the function](https://code.orgmode.org/bzg/org-mode/src/master/lisp/org.el#L350). The result of `(org-version)` at that point is "N/A". org-journal [reads that constant](https://github.com/bastibe/org-journal/blob/bca2c39c692f7c58dbbe03698748187f87822043/org-journal.el#L68) and fails to get the proper version number.

The attached commit fixes the issue for me. However, elisp is not within my usual comfort zone, so there might be better ways to do it.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -5356,6 +5356,7 @@ Inserted by installing org-mode or when a release is made."
                "--abbrev=0"
                "HEAD")))))))
 
+    (defconst org-version (org-version))
     (provide 'org-version)))
 
 (if straight-fix-org


raxod502:
Thanks for the code links. I don't think this change should be necessary, though. With the new Org hack, the function `org-version` ought to be defined before Org is even byte-compiled, so this variable ought to be set correctly by Org in the first place.

The following code produces no error for me:

```elisp
(setq straight-repository-branch "develop")

(defvar bootstrap-version)
(let ((bootstrap-file
       (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
      (bootstrap-version 5))
  (unless (file-exists-p bootstrap-file)
    (with-current-buffer
        (url-retrieve-synchronously
         "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
         'silent 'inhibit-cookies)
      (goto-char (point-max))
      (eval-print-last-sexp)))
  (load bootstrap-file nil 'nomessage))

(straight-use-package 'org)
(straight-use-package 'org-journal)
(require 'org-journal)
```

Perhaps you simply need to `M-x straight-rebuild-package RET org RET`?

sboehler:
Thank you for your minimal example, it prompted me to search along a different route and helped me track down a bug in my init which caused the initial issue! 

I keep my init in an init.org file which gets automatically tangled to init.el on save. While all packages are being configured with use-package, there was a single defun which actually implements the auto-tangling-function at the very bottom of my huge org file and which gets added to after-save-hook. It references `org-babel-tangle`, which probably caused some built-in org-mode code to load, resulting all kinds of funny side effects.

Needless to say, after moving that function into org-mode's use-package stanza, everything works as intended!

Thank you for your reply, and even more for providing straight.el. Before, I used nix to manage my emacs packages, but with straight and its git-based approach it is much easier to just go to a specific piece of package code and modify it.

Issue #294, opened 2018-07-14
straight-freeze-versions + straight-thaw-versions is not no-op


antifuchs:
As suggested in https://github.com/raxod502/straight.el/issues/292#issuecomment-404868441, I made a reproducer repo for the issue I'm seeing: https://github.com/antifuchs/straight.el-repro-case-1 - it's an init file pulling in ace-window, and a lock file locking that to an older version.

To reproduce:

1. Clone the repo above
2. `emacs -Q init.el`
3. `C-M-x` to evaluate the init file (this will bootstrap straight, so it's going to take a while)
4. `M-x straight-thaw-versions`
5. `M-x straight-freeze-versions`
6. In the clone from above, run `git diff`

On my machine, `git diff` says that the `ace-window` version changed to the latest. I'd expect there to be no differences. This also reproduces for me if I restart emacs between steps 4 and 5.

I'm testing this on `GNU Emacs 26.1 (build 1, x86_64-apple-darwin13.4.0, Carbon Version 157 AppKit 1265.21) of 2018-04-15` - the latest bottle from https://github.com/railwaycat/homebrew-emacsmacport

antifuchs:
I just stepped through the freeze process, and it looks like `straight-vc-git--ensure-head-pushed` (via `straight-vc-push-to-remove`, invoked from the freeze function) is what resets the checkout for `ace-window` to the `master` branch. I'll step through this some more & hopefully will find the exact trace of git things that happen when it does.

raxod502:
Oh. Yes, that is the intended behavior, except that it causes this horrible effect because #66 has not been fixed yet. I think that #66 is the correct way to address this problem—you do want to make sure that the default branch is checked out (otherwise how can you be sure that the commit will be accessible when the configuration is cloned on another machine?) but _of course_ `M-x straight-thaw-versions` followed by `M-x straight-freeze-versions` should be a no-op, whereas currently it is not. Thawing versions should reset the default branch rather than detaching HEAD.

antifuchs:
Ah, that makes a ton of sense - I think resetting the branch to the previously checked-out rev would be quite reasonable (if that rev was an ancestor of the remote's pushed branch).

raxod502:
This should be fixed by https://github.com/raxod502/straight.el/pull/375.

Issue #66, opened 2017-07-07
Move away from decapitation


raxod502:
Currently `straight-thaw-versions` works by detaching your HEAD at the appropriate commit. I'd like to move towards instead doing a normalization operation followed by resetting master. This would make `straight-normalize-all` a no-op when performed after `straight-thaw-versions`, and vice versa.

raxod502:
Should be fixed by https://github.com/raxod502/straight.el/pull/375.

Issue #58, opened 2017-06-28
straight-thaw-versions requires fetch


raxod502:
You may need to update your repositories or remotes in order to check out the correct commit.

raxod502:
To clarify, I mean that fetching should be done automatically, since I think that would be an improvement in UX. See also #280.

raxod502:
Should be fixed by https://github.com/raxod502/straight.el/pull/375.

PR #375, opened 2019-04-20
straight-thaw-versions ergonomics improvements: auto-fetch, no decapitation

antifuchs:
This PR is kinda a catch-all for the things that I found that were ... not ideal with thawing:

* #294 - freezing after thawing is not a no-op
* #66 - thawing leaves detached HEADs around
* #58 - thawing requires fetching beforehand

I think the attached change solves all of these in a fairly neat way,
and I hope you think so too (and that I didn't miss some subtle
interaction!).

`straight-vc-git-check-out-commit` now normalizes the local repo when checking out a commit.

`straight-thaw-versions` now:

* Tries to ensure that the commit that we're thawing to is present by checking it's there, and if not, fetching.
* Uses a new and revamped check-out-commit function that `resets --hard` instead of running `checkout`.

With this change, I managed to thaw and re-freeze the versions (with an updated repro case from #294), and they were stable!

I've also updated the README to not mention having to fetch before a freeze, as freeze does that automatically now.

Hope this is reasonable!

### DIFF
--- a/README.md
+++ b/README.md
@@ -2668,8 +2668,7 @@ UX of this situation.
 ### The interactive version-control operations are confusing
 
 This part of `straight.el` still needs some work; see [#54] about the
-UX of pushing and pulling, and [#58] about commits not being available
-when thawing a lockfile.
+UX of pushing and pulling.
 
 ### How do I pin package versions or use only tagged releases?
 
@@ -2686,6 +2685,14 @@ with Emacs, rather than cloning the upstream repository:
 [Read more.][#user/recipes]
 
 ## News
+### May 1, 2019
+
+`straight-thaw-versions` now fetches in a repository if a commit in
+your lockfile can not be found and normalizes the repository to the
+recipe's default branch. This should ensure that versions you have
+frozen can be quickly restored and that they can also be saved back to
+the version lock file. This addresses issues [#58], [#66], and [#294].
+
 ### March 15, 2019
 
 `straight.el` [now installs a hack for Org by
--- a/README.md
+++ b/README.md
@@ -2240,11 +2240,9 @@ make sure to check out the specified revisions of each package when
 cloning them for the first time.
 
 To install the versions of the packages specified in your version
-lockfiles, run `M-x straight-thaw-versions`. You may need to run `M-x
-straight-fetch-all` first to ensure that you have those versions
-available. Thawing will interactively check for local changes before
-checking out the relevant revisions, so don't worry about things
-getting overwritten.
+lockfiles, run `M-x straight-thaw-versions`. Thawing will
+interactively check for local changes before checking out the relevant
+revisions, so don't worry about things getting overwritten.
 
 #### The profile system
 
--- a/README.md
+++ b/README.md
@@ -1743,6 +1743,9 @@ a backend API method. The relevant methods are:
 
 * `clone`: given a recipe and a commit object, clone the repository
   and attempt to check out the given commit.
+* `commit-present-p`: given a recipe and a commit object, return
+  whether the commit can be checked out offline, i.e., without
+  fetching from the remote.
 * `normalize`: given a recipe, "normalize" the repository (this
   generally means reverting it to a standard state, such as a clean
   working directory, but does not entail checking out any particular
@@ -1813,6 +1816,8 @@ the version-control backend API:
   your revision lockfile, or the `:branch` (from the `:fork`
   configuration, if given), or `origin/HEAD`. If a `:fork` is
   specified, also fetches from the upstream.
+* `commit-present-p`: checks if the commit SHA is among the revisions
+  that are present locally.
 * `normalize`: verifies that remote URLs are set correctly, that no
   merge is in progress, that the worktree is clean, and that the
   primary `:branch` (from the `:fork` configuration, if given) is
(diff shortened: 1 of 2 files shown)


antifuchs:
Phew, took me 2 tries to get lints to work out, but I think this should do it!

straight.el
```diff
@@ -1212,6 +1212,14 @@ repository directory and delegates to the relevant
...
     (straight-vc 'check-out-commit type local-repo commit)))
 
+(defun straight-vc-commit-present-p (recipe commit)
+  "Check in RECIPE's repo if COMMIT can be checked out without fetching it.
+Returns non-nil if the commit is available."
+  (straight--with-plist recipe
+      (local-repo type)
+    (let ((straight--default-directory (straight--repos-dir local-repo)))
+      (straight-vc 'commit-present-p type local-repo commit))))
+
```
raxod502: This needs a bullet point in the README section that lists VC backend operations, as well as one in the section that describes the Git backend's implementations of them.

straight.el
```diff
@@ -2066,9 +2074,17 @@ is a 40-character string identifying a Git commit."
     (while t
       (and (straight-vc-git--ensure-nothing-in-progress local-repo)
            (straight-vc-git--ensure-worktree local-repo)
-           (straight--get-call "git" "checkout" commit)
+           (straight--get-call "git" "reset" "--hard" commit)
            (cl-return)))))
 
+(cl-defun straight-vc-git-commit-present-p (local-repo commit)
```
raxod502: This should have a docstring.

straight.el
```diff
@@ -4987,6 +5003,9 @@ according to the value of `straight-profiles'."
              ;; We can't use `alist-get' here because that uses
              ;; `eq', and our hash-table keys are strings.
              (when-let ((commit (cdr (assoc local-repo versions-alist))))
+               (straight-vc-normalize recipe)
+               (unless (straight-vc-commit-present-p recipe commit)
+                 (straight-vc-fetch-from-upstream recipe))
```
raxod502: I think we should also have a `straight-vc-fetch-from-remote`, no?

straight.el
```diff
@@ -2066,9 +2074,17 @@ is a 40-character string identifying a Git commit."
...
       (and (straight-vc-git--ensure-nothing-in-progress local-repo)
            (straight-vc-git--ensure-worktree local-repo)
-           (straight--get-call "git" "checkout" commit)
+           (straight--get-call "git" "reset" "--hard" commit)
            (cl-return)))))
 
+(cl-defun straight-vc-git-commit-present-p (local-repo commit)
+  (cl-block nil
+    (while t
+      (and (straight-vc-git--ensure-nothing-in-progress local-repo)
```
raxod502: Do we need this?

antifuchs: I wasn't sure - it seemed like a good idea to prevent concurrent operations that the user might have kicked off inadvertently, but given that this is non-destructive, it should be extremely ok leave the looping/waiting to functions that are destructive. I'll remove it.

straight.el
antifuchs: Oh myyyy, yes it should be that, I missed that!

antifuchs:
Thanks so much for the review! I have gone through your suggestions & think I addressed them! 

I'd be excited for this to be part of the NEWS - we should probably do this as part of another commit, since I don't yet know when it'll go in (:

straight.el
raxod502: I think my position is that if you run Git commands concurrently with `straight.el` VC operation, then you are perfectly well entitled to shoot yourself in the foot. It would be hopeless to try to fix all the cases that would result from such unpredictable state.

straight.el
```diff
@@ -2066,9 +2074,14 @@ is a 40-character string identifying a Git commit."
     (while t
       (and (straight-vc-git--ensure-nothing-in-progress local-repo)
            (straight-vc-git--ensure-worktree local-repo)
-           (straight--get-call "git" "checkout" commit)
+           (straight--get-call "git" "reset" "--hard" commit)
```
raxod502: Now that I'm looking at this again, it seems to me that the normalization logic should be included here. When would we ever want to check out a commit by resetting a branch other than the primary branch?

The fetch, though, should probably happen where you have it.

antifuchs:
Alright - I moved this to the `develop` branch, made that News entry (dating it May 1; feel free to adjust if you need), and added the normalization logic to the check-out-commit function. That needed an incompatible change to the vc interface: check-out-commit only took the local repo, not the recipe (which is needed to normalize). 

I adjusted the interfaces, but didn't do anything else about it - please let me know if there's anything else that I should do about it (:

README.md
```diff
@@ -2682,6 +2685,14 @@ with Emacs, rather than cloning the upstream repository:
 [Read more.][#user/recipes]
 
 ## News
+### May 1, 2019
```
raxod502: Run `make toc` to update the table of contents.

straight.el
```diff
@@ -1203,8 +1203,8 @@ repository directory and delegates to the relevant
       (let ((straight--default-directory (straight--repos-dir local-repo)))
         (straight-vc 'push-to-remote type recipe)))))
 
-(defun straight-vc-check-out-commit (type local-repo commit)
-  "Using VC backend TYPE, in LOCAL-REPO, check out COMMIT.
+(defun straight-vc-check-out-commit (recipe commit)
+  "Normalize the repo for RECIPE and check out COMMIT.
 TYPE is a symbol like symbol `git', etc. LOCAL-REPO is a string
```
raxod502: This docstring needs updating.

antifuchs: While I was there, I adjusted the function's behavior to not do anything if there's no local-repo, symmetrical to the other vc functions that take a recipe.

antifuchs:
Updated again (:

antifuchs:
\o/ - thanks for the thoughtful reviews! I'm super excited to use this going forward (:

PR #376, opened 2019-04-24
Fix symlink unpriviledged in windows

kiennq:
>* [Windows Creators Update](https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/) supports  symlink-creation without any special permission setup.

This is not entirely correct for `Emacs` implementation of `make-symbolic-link`.
It seems that when calling `CreateSymbolicLink`, `Emacs` needs to add a new flag `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` for symlink can be created unpriviledged in Developer mode.

I've created a mitigation by using `mklink` command which works independently from `Emacs` implementation and supported in Windows.

### DIFF
--- a/install.el
+++ b/install.el
@@ -186,14 +186,14 @@
                    ;; This is a relative symlink. It won't break if you
                    ;; (for some silly reason) move your
                    ;; `user-emacs-directory'.
-                   (target (concat "repos/" local-repo "/bootstrap.el"))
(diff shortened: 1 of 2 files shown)


kiennq:
Fixed CI

raxod502:
Thanks!

Issue #246, opened 2018-03-03
:type git :branch vs :ref?


nrvale0:
:branch works with other refs like a tag name. Which is great. But perhaps it should be renamed to :ref?

raxod502:
This requires some more thought. The repository management algorithms are not actually designed to work with any refs other than branch names, since it's expected that you use version lockfiles to specify exact revisions or tags, rather than putting them in the recipes directly.

nrvale0:
Ah, ok. I've just gotten started with straight.el so I've not yet explored lockfiles. Entirely possible this is better approach.

raxod502:
If your goal in specifying a tag or revision in the recipe is that your configuration will use that revision of the package, then indeed version lockfiles are designed to achieve exactly that purpose.

dieggsy:
That said, it'd be interesting to have a `:ref` that supports arbitrary refs, with the end goal being that the _only_ source of truth is the init file. This is probably not as intuitive as lockfiles since they're derived from whatever ref you `git checkout` to, but it might be a nice advanced feature.

raxod502:
> it'd be interesting

Sure, and I wouldn't be opposed to this...

> with the end goal being that the _only_ source of truth is the init file

... but I'm honestly not sure why you'd want to do it. You'd have to edit your init-file every time you update a package.

dieggsy:
Well, if your workflow is one in which you update a lot (which I actually do) then it'd be annoying, but some people might prefer to "pin" packages in that way.

Regardless, it'd be an additional feature, so you wouldn't be required to 'pin' every package with a specific ref throughout. But if a package came up that you did want to pin or revert, you'd have the convenience of simply specifying the ref in the init file, without pinning everything to it's current state through `straight-freeze-versions`

nrvale0:
> ... but I'm honestly not sure why you'd want to do it. You'd have to edit your init-file every time you update a package.

I can speak to this a bit. I went down the path of straight.el to allow me to install the latest version of Org and associated packages without conflicts with RPM-packaged Org on Fedora (not possible with use-package). All other packages are still installed with use-package. 

So, in my case, having a lockfile separate from the package specification in my init-file is just one more thing to track in the repo and in my head with not much payout. 

Though I agree, having worked with similar setups in other environments (requirements.txt, Gemfile.lock, etc) that specification in the init-file probably does not scale beyond a handful of packages under straight.el management.

raxod502:
> Regardless, it'd be an additional feature

[...]

PR #380, opened 2019-05-19
Add package pinning to straight-x.el.

DonHugo69:
Dear @raxod502 ,

in reply to your comment in [this issue](https://github.com/raxod502/straight.el/issues/246#issuecomment-472702266) I would like to ask for inclusion of my temporary solution for package pinning. 

I have updated the code to work with the changes introduced by commit [c42edf91249a3610a872402082e977c61cf65dc9](https://github.com/raxod502/straight.el/commit/c42edf91249a3610a872402082e977c61cf65dc9) and written a short tutorial on how to use it. 

I hope the code's quality is good enough for inclusion. If you require any changes please let me know and I will try to take care of them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2728,10 +2728,16 @@ After straight's install procedure you will need to add
 `straight-x.el` and load the required commands.
 
     (use-package straight-x
-      :straight nil 
-      :commands
-      (straight-x-pull-all
-       straight-x-freeze-versions))
+      :straight (:host github
+                 :repo "raxod502/straight.el"
+                 :branch "develop"
+                 :local-repo "straight.el"
+                 :files ("straight-x.el"))
+      :commands (straight-x-pull-all
+                 straight-x-freeze-versions))
+                 
+Please take care that `straight-x.el`'s `:branch` is the same as for
+`straight.el`. By default, this will be `master`.
 
 A variable called `straight-x-pinned-packages` has been defined in
 `straight-x.el` and will contain your list of pinned packages.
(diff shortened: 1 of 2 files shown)


raxod502:
It looks great to me. Only problem is the Travis build (which _should_ be linked here, but isn't for some reason): https://travis-ci.org/raxod502/straight.el/jobs/534460538

I think the solution to the error there would be to include a `(defvar straight-x-pinned-packages)` to satisfy the byte-compiler. (Or just define the variable in `straight-x.el` directly.)

README.md
```diff
@@ -2733,11 +2733,8 @@ the required commands.
       (straight-x-pull-all
        straight-x-freeze-versions))
 
-Furthermore, you will need to define a variable called
-`straight-x-pinned-packages`.
-
-    (defvar straight-x-pinned-packages nil
-      "List of pinned packages.")
```
raxod502: Actually, this is still necessary if you are trying to autoload `straight-x.el` (i.e. not load it eagerly), right? Otherwise the list variable will not be defined when you try to append to it.

There's no need to work out the details in this pull request, but I think it would be possible to declare autoloads in `straight-x.el`, exclude that file from the default recipe, and then allow `(straight-use-package '(straight-x :local-repo "straight.el" :files ("straight-x.el")))` as the way to access them, or some such.

DonHugo69: You are right, the variable will not exist that way. 
I did not use autoload cookies here because my understanding was, that `straight-x.el` is experimental and a user should cherry pick the desired functions after looking at them. If it is fine for you, I will add the autoload cookie to the variable (but not to the functions?).

And try to fix the longline errors.

DonHugo69: I have checked the instructions again and made the necessary modifications. The variable gets autoloaded now together with the specified commands.

raxod502:
Thanks for your contribution.

DonHugo69:
Thank you for creating straight :-)

PR #387, opened 2019-06-14
Don't call pre-build functions unless building

### LINKED ISSUE #386: What to do with package that requires autotools and configure and make to build...
I hesitate to ask, but....  I've been reading my mail in `vm` for a very long time.  Its definitely moribund, if not mostly dead, but it works.   

I've switched nearly all of my emacs set up to use `use-package` (aka `straight-use-package`), but I'm still wiring up my vm by hand from a copy of it's 8.2.0b release.

While this might be a job for Miracle Max the Wizard ("mostly dead", obligatory Princess Bride ref), I wanted to see what I could do about it.

I was fascinated to discover that straight would happily 'use-package' it, but not in the least surprised that it didn't work.

It turns out that it's mirrored here:  https://github.com/emacsmirror/vm and that's what straight's picking up.  That tree, however, seems to need to have autotools called on it, then configure (with a bit of extra info to tie it into bbdb in my case).

And then, there are a couple bits that don't actually work; they removed the bundled vcard.el for reasons their commit doesn't specify, but didn't commit it out of the Makefile.in.  Nor the vm-vcard.el that was using it.....    There's also a use of `&optional &rest ...` that irritates emacs and that I've fixed before by just removing the `&optional`).

So, at a minimum seems like I should fork the repo, fix those bits, then have straight use my fork (and branch).

For experimenting, I've cloned the repo, run autoconf and then configure.  It makes successfully, albeit without the bbdb bits.  So in addition to asking about a straight-approved way of calling autoconf and then make, I suppose I should also ask if there's a non-ugly way to get at the path where straight has built bbdb for me?

**Thanks**

hartzell:
Don't call the things on the `straight-use-package-pre-build-functions` hook unless we're actually going to do a build.

I use the pre-build-functions hook to call `autoconf/configure/make` for the vm package before the build happens and w/out this change it was happening on every startup.  With this change it only happens when necessary.

Closes #386

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -4665,14 +4665,14 @@ otherwise (this can only happen if NO-CLONE is non-nil)."
                                (and table (gethash 'version table))
                                (funcall func))))
                  (remhash (intern package) straight--recipe-lookup-cache)))
-             (run-hook-with-args
-              'straight-use-package-pre-build-functions package)
              (unless no-build
                ;; Multi-file packages will need to be on the
                ;; `load-path' in order to byte-compile properly. So we
                ;; do this before `straight--build-package'.
                (straight--add-package-to-load-path recipe))
              (when (and modified (not no-build))
+               (run-hook-with-args
+                'straight-use-package-pre-build-functions package)
                (straight--build-package recipe cause))
              ;; We need to do this even if the package wasn't built,
              ;; so we can keep track of modifications.


raxod502:
Looks great, thanks for the help.

hartzell:
You're welcome.  Thanks for the mentoring!

PR #396, opened 2019-07-08
symlinks ARE useable OOTB on cygwin

vzell:
No need to set "(setq straight-use-symlinks t)".  Tested on a couple of systems with Cygwin installed.


raxod502:
Sounds good, thanks :)

PR #404, opened 2019-08-16
Implement configurable base path

darth10:
This is a simple implementation of a configurable base path based on #300. 
There's a couple other changes/hacks that I've mentioned in comments.

This should not break existing installations as there are no changes to `bootstrap.el`.

Related to #274.

### DIFF
--- a/install.el
+++ b/install.el
@@ -88,7 +88,9 @@
         (version :pluto)
         (straight-profiles (if (boundp 'straight-profiles)
                                straight-profiles
-                             '((nil . "default")))))
+                             '((nil . "default"))))
+        (straight-install-dir (or (bound-and-true-p straight-base-dir)
+                                  user-emacs-directory)))
     ;; The only permissible error here is for a lockfile to be absent
     ;; entirely. Anything else triggers an abort so that we don't
     ;; accidentally do something the user doesn't expect (like if they
@@ -96,7 +98,7 @@
     ;; the recipe specification, and forgot to update which repository
     ;; their init-file downloaded install.el from).
     (dolist (lockfile-name (mapcar #'cdr straight-profiles))
-      (let ((lockfile-path (concat user-emacs-directory
+      (let ((lockfile-path (concat straight-install-dir
                                    "straight/versions/"
                                    lockfile-name)))
         (when (file-exists-p lockfile-path)
@@ -184,7 +186,7 @@
                    ;; (for some silly reason) move your
                    ;; `user-emacs-directory'.
                    (link-target (concat "repos/" local-repo "/bootstrap.el"))
-                   (link-name (concat user-emacs-directory
+                   (link-name (concat straight-install-dir
                                       "straight/bootstrap.el")))
               (ignore-errors
                 ;; If it's a directory, the linking will fail. Just let
--- a/install.el
+++ b/install.el
@@ -85,7 +85,7 @@
 
   (let (;; This needs to have a default value, just in case the user
         ;; doesn't have any lockfiles.
-        (version :pluto)
+        (version :alpha)
         (straight-profiles (if (boundp 'straight-profiles)
                                straight-profiles
                              '((nil . "default"))))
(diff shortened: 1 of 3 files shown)


install.el
```diff
@@ -83,6 +83,9 @@
   (defvar url-http-end-of-headers)
   (defvar url-http-response-status)
 
+  (defvar straight-base-dir)
+  (defvar straight-repository-branch)
```
darth10: Need these `defvar` forms for linting.

install.el
```diff
@@ -123,10 +126,12 @@
...
     (with-current-buffer
         (url-retrieve-synchronously
+	 ;; HACK Maybe use the install/* branch here?
+	 ;; TODO Revert github user to raxod502
          (format
           (concat "https://raw.githubusercontent.com/"
-                  "raxod502/straight.el/install/%s/straight.el")
-          (substring (symbol-name version) 1))
+                  "darth10/straight.el/%s/straight.el")
+          straight-repository-branch)
```
darth10: Instead of using the `install/` branches, I've made a change to use the `straight-repository-branch` branch.
I'm more than happy to change this back.
We would need to decide a new version name then am I'm not really sure what comes after `:pluto`. Maybe Alpha Centauri? :smile:

darth10: I've changed the github user for testing purposes only.
I'll revert that once this is ready to land.

README.md
```diff
@@ -174,13 +174,17 @@ First, place the following bootstrap code in your init-file:
...
       (unless (file-exists-p bootstrap-file)
         (with-current-buffer
             (url-retrieve-synchronously
-             "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
+              (format
+               "https://raw.githubusercontent.com/raxod502/straight.el/%s/install.el"
+               straight-repository-branch)
              'silent 'inhibit-cookies)
           (goto-char (point-max))
           (eval-print-last-sexp)))
```
raxod502: I'd prefer to keep this as simple as possible. I don't think it's unreasonable to expect people to tweak the bootstrap snippet a little if they want to do something weird.

install.el
raxod502: The `install/` branches are there for a reason. They make it so that you can install old versions of `straight.el` that interpreted recipes differently. You can't just use the latest version. (I suggest looking through the comments at the top of `install.el`.)

install.el
```diff
@@ -96,7 +99,7 @@
     ;; the recipe specification, and forgot to update which repository
     ;; their init-file downloaded install.el from).
     (dolist (lockfile-name (mapcar #'cdr straight-profiles))
-      (let ((lockfile-path (concat user-emacs-directory
+      (let ((lockfile-path (concat straight-base-dir
```
raxod502: Don't rely on this variable being defined. It breaks backwards compatibility. Just use `(or (bound-and-true-p ...) ...)`.

install.el
```diff
@@ -147,7 +152,7 @@
           ,@(let* ((vars nil)
                    (regexps '("bootstrap-version"
                               "straight-[a-z-]+"
-                              "user-emacs-directory"))
+                              "straight-base-dir"))
```
raxod502: I think we still need to pass `user-emacs-directory` in. And we don't need an entry for `straight-base-dir`; that's already covered by the previous regexp.

install.el
raxod502: As for continuing the schema, I propose using the Greek alphabet: `:alpha`, `:beta`, `:gamma`, `:delta` etc.

darth10:
Thanks for the feedback, @raxod502! 
I've made the relevant changes.

darth10:
Thanks for the feedback, @raxod502! 
I've made the relevant changes and bumped the version to `:alpha`.

Please let me know if there are any more improvements I can make.

hlissner:
I'm getting an `Unknown recipe version: :alpha` error when bootstrapping the latest version. I think an install/alpha tag is missing.

raxod502:
My bad -- should be fixed now.

Issue #409, opened 2019-09-07
[non-Windows OS] [straight--symlink-recursively] [cmd] no symlinks created


rprimus:
Sat Sep  7 13:26:44 BST 2019

Based on https://github.com/hlissner/doom-emacs/issues/1715, this was caused by having `cmd` in my `$PATH`.  The [code](https://github.com/raxod502/straight.el/blob/3794237a251c759262fb0b1a2579344aba4caf9a/straight.el#L706) could be made more 'robust' by 
1. only check for `cmd` when the OS is `Windows`.
2. using the absolute path to the location of the native `Windows` `cmd` - in case a user may have another `cmd` in their `PATH` ahead of the native one.

Thoughts?

raxod502:
That is a _big_ oof.

* Is there any way you can be on Windows but _not_ have `cmd`? If not, then let's just replace the `if` condition with a conditional on `system-type`.
* Otherwise, let's do your suggestion (1).

raxod502:
@hlissner I didn't know until just now that Doom uses `straight.el`! This is really super cool and I'm so happy the project has been helpful for you. I don't have much bandwidth, but I would be very excited to work with you to make sure `straight.el` can be maximally helpful for embedding in other projects like Doom. Also, I was extremely impressed at the good design when looking through your commit log. I may actually be interested in adding my configuration development efforts to Doom and migrating from [Radian](https://github.com/raxod502/radian).

rprimus:
Sat Sep  7 17:52:17 BST 2019

> Is there any way you can be on Windows but not have cmd? If not, then let's just replace the if condition with a conditional on system-type.

From the [docs](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/windows-commands):

```
All supported versions of Windows (server and client) have a set of Win32 console commands built in.
```

On the list is `cmd`.   I'd go with your suggested fix!

Cheers.

rprimus:
Sat Sep  7 18:14:59 BST 2019

Might be worth creating a `windows-p` function - as a check here would be the fourth one in `straight.el` - with only one instance (thus far) checking additionally for `cgywin`.

rprimus:
Sat Sep  7 18:49:17 BST 2019

PR: https://github.com/raxod502/straight.el/pull/410 created.

rprimus:
Mon Sep 16 16:44:40 BST 2019

https://github.com/raxod502/straight.el/pull/413  has been merged.

Closing issue.

PR #413, opened 2019-09-16
[https://github.com/raxod502/straight.el/issues/409] Replace executab…

rprimus:
Mon Sep 16 15:26:46 BST 2019

…le-find "cmd" with check for Windows os

make checkindent passes


raxod502:
The CI problem should be fixed.

PR #414, opened 2019-09-17
Add user-facing function to visit a package's website

swaroopch:
Useful when package is not already installed.

----

#

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -4538,6 +4538,20 @@ action, just return it)."
                (message "Copied \"%S\" to kill ring" recipe))
         (_ recipe)))))
 
+;;;;; Jump to package website
+
+;;;###autoload
+(defun straight-visit-package-website ()
+  "Interactively select a recipe, and visit the package's website."
+  (interactive)
+  (let* ((melpa-recipe (straight-get-recipe))
+         (recipe (straight--convert-recipe melpa-recipe)))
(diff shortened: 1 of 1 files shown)


straight.el
```diff
@@ -4538,6 +4538,21 @@ action, just return it)."
...
+;;;;; Jump to package website
+
+;;;###autoload
+(defun straight-visit-package-website ()
+  "Interactively select a recipe, and visit the package's website."
+  (interactive)
+  (let* ((melpa-recipe (straight-get-recipe))
+         (recipe (straight--convert-recipe melpa-recipe))
+         (host (plist-get recipe :host))
+         (repo (plist-get recipe :repo)))
```
raxod502: This could use `straight--with-plist`, but not a big deal.

straight.el
```diff
@@ -4538,6 +4538,21 @@ action, just return it)."
...
+    (pcase host
+      ('github (browse-url (format "https://github.com/%s" repo)))
+      ('gitlab (browse-url (format "https://gitlab.com/%s" repo)))
```
raxod502: These need to be backquotes for compatibility with Emacs 24. Check the Travis build.

straight.el
swaroopch: done 👍

straight.el
swaroopch: done 👍

PR #415, opened 2019-09-18
Fix magit deprecation warning

swaroopch:
Not sure what is straight.el's policy to support older magit versions...

```
‘magit-status-internal’ is an obsolete function (as of Magit 2.91.0);
use ‘magit-status-setup-buffer’ instead.
```

----

#

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -112,7 +112,7 @@ They are still logged to the *Messages* buffer.")))
 (defvar package--builtins)
 
 ;; `magit'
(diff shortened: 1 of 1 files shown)


raxod502:
You need to update the `declare-function` definitions though, to fix the build.

swaroopch:
> You need to update the `declare-function` definitions though, to fix the build.

Fixed it. Builds pass now 👍

raxod502:
Thanks for your contribution!

PR #417, opened 2019-09-28
Make emacsmirror parsing deal with CRLF files (fix #416)

### LINKED ISSUE #416: straight-recipes-emacsmirror-mirror-retrieve doesn't work if mirror file is in CRLF encoding
aae8a57dc70077ba4afeaab6febe1f7bd2e63572 changed `straight-emacsmirror-use-mirror` to `t`. When launching [my configuration](https://github.com/drrlvn/.emacs.d/blob/b7a01affd97ae1741e6a97425cf5dde917f51c0d/init.el#L5) in Windows I get:

```
error: Could not find package syntax-subword in recipe repositories: (org-elpa melpa gnu-elpa-mirror emacsmirror-mirror)
```

This seems to happen only on Windows. What's even stranger is that setting the variable back to `nil` with `setq` after `(require 'straight)` doesn't fix it, only when calling `setq` before `require` fixes it.

r-darwish:


raxod502:
Thanks!

PR #422, opened 2019-10-13
Generate dir file for pre generated info files

dajva:
This was only done for info files generated from texinfo files but
some packages provide pre generated info files. This makes sure the
latter are added into the main info directory menu.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -4131,7 +4131,8 @@ repository."
                         (straight--repos-dir local-repo)
                         (straight--build-dir package)
                         flavor))
-          (when (string-match-p ".texi\\(nfo\\)?$" repo-file)
+          (when (or (string-match-p ".info$" repo-file)
+                    (string-match-p ".texi\\(nfo\\)?$" repo-file))
             (let ((texi repo-file)
                   (info
                    (concat (file-name-sans-extension build-file) ".info")))
--- a/straight.el
+++ b/straight.el
@@ -4132,8 +4132,7 @@ repository."
                         (straight--build-dir package)
                         flavor))
           (cond
-           ((string-match-p ".info$" repo-file)
-            ;; info file has already been symlinked so no more action needed.
+           ((string-match-p ".info$" build-file)
             (push build-file infos))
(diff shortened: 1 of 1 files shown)


straight.el
```diff
@@ -4131,7 +4131,8 @@ repository."
                         (straight--repos-dir local-repo)
                         (straight--build-dir package)
                         flavor))
-          (when (string-match-p ".texi\\(nfo\\)?$" repo-file)
+          (when (or (string-match-p ".info$" repo-file)
+                    (string-match-p ".texi\\(nfo\\)?$" repo-file))
             (let ((texi repo-file)
                   (info
                    (concat (file-name-sans-extension build-file) ".info")))
```
raxod502: doesn't this also need to change, otherwise we are passing info files to makeinfo rather than just install-info?

dajva: My tests show that we will not invoke `makeinfo` because of the `(file-exists-p info)` check further down. AFAICT, it's because the info file has already been symlinked into the build directory by some earlier stage in the build. Might be something I am missing though and I only tested it on a package with the info file in the main repo dir.

I admit my change make the code a bit clunky and harder to understand. Could be improved by separating the info and texinfo case since if the info file is already existing it only needs to be pushed to the list of info files.

dajva: Please advise me how to proceed with this issue. Thanks.

raxod502: Oops, I apologize. I thought I had submitted a comment, but actually GitHub left it on the floor as a "pending review". I think it is definitely a good idea to refactor the function so that we explicitly separate the texinfo and info cases. It's confusing and fragile the way the code is written currently.

dajva: Thanks, I'll have a look at that.

dajva: Done

dajva:
Great. Before submitting this, I checked what `package-build` does and I am pretty sure I just copied the regexp from there without thinking about more about it. Looking it up now, the same bug is indeed there. Quite interesting how small issues like this just propagagates without anyone noticing.
Except you did in this case. :)

PR #427, opened 2019-10-28
Allow recipes to inherit from defaults

### LINKED ISSUE #116: Recipe inheritance
Inherit recipe keywords from the formula in the recipe repositories by default; provide user option and keyword to disable.

CeleritasCelery:
`:host`, `:repo`, and `:files` are inherited from the default recipe. Also `:fork` will inherit its host from the upstream recipe. These can be overridden like any recipe component.

closes #116 

For example, we can do the following simplification
### example 1
before:
```lisp
(package :host gitlab :repo "other-user/repo" 
         :files ("lisp.el")
         :fork (:host :gitlab :repo "my-user/repo"))
```
after:
```lisp
(package :fork "my-user/repo")
```
or 
```lisp
(package :fork (:repo "my-user/repo"))
```
### example 2
before:
```lisp
(package :host gitlab :repo "user/repo" :local-repo "local")
```

after:
```lisp
(package :local-repo "local")
```

Let me know what you think. If you give the green light I will add some documentation to the main README.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -208,6 +208,36 @@ profile (nil) will suffice without additional setup."
                 (alist :key-type symbol :value-type
                        (plist :key-type symbol :value-type sexp))))
 
+(defcustom straight-allow-recipe-inheritance t
+  "Non-nil allows partially overriding recipes.
+If you override a recipe, every component that is not explicitly
+overriden will be searched for in original recipe. If found, that
+value will be added to the overriden recipe. This allows you to
+only override the recipe components you are interested in,
+instead of being required to override them all. The supported
+components are the ones listed by `straight-vc-git-keywords' and
+`:files'. Note that enabling this feature has the side effect
+that all recipe repos (i.e. melpa, elpa) will always be cloned,
+even if you explicitly specify all your recipes.
+
+The `:fork' keyword is handled specially. If its value is a
+string instead of a list, then it is assigned as the `:repo' of
+the fork. Also the fork recipe will inherit its `:host' component
+from the default recipe.
+
+For example, the following are all equivalent with recipe
+inheritance enabled.
+
+\\='(package :host \\='gitlab :repo \"other-user/repo\"
+          :fork (:host \\='gitlab :repo \"my-user/repo\"))
+
+\\='(package :fork (:host \\='gitlab :repo \"my-user/repo\"))
+
+\\='(package :fork (:repo \"my-user/repo\"))
+
+\\='(package :fork \"my-user/repo\")"
+  :type 'boolean)
+
 (defcustom straight-safe-mode nil
   "Non-nil means avoid doing anything that modifies the filesystem.
 In safe mode, package modifications will still be detected
@@ -2912,6 +2942,26 @@ for dependency resolution."
           ;; override the default value (which is determined according
           ;; to the selected VC backend).
           ;;
+          (when straight-allow-recipe-inheritance
+            ;; To keep overridden recipes simple, Some keywords can be
+            ;; inherited from the original recipe. This is done by
+            ;; looking in original and finding all keywords that are
+            ;; not present in the override and adding them there.
+            (let ((fork (plist-get plist :fork)))
+              (when (stringp fork)
+                (straight--put plist :fork `(:repo ,fork))))
+            (let* ((default (cdr (straight-recipes-retrieve package)))
+                   (keywords (straight-vc-keywords (or (plist-get default :type) 'git))))
+              (dolist (keyword (cons :files keywords))
+                (if (eq keyword :fork)
+                    (dolist (keyword keywords)
+                      (let ((fork-plist (plist-get plist :fork))
+                            (value (plist-get default keyword)))
+                        (when (and value fork-plist (not (plist-member fork-plist keyword)))
+                          (straight--put plist :fork (plist-put fork-plist keyword value)))))
+                  (let ((value (plist-get default keyword)))
+                    (when (and value (not (plist-member plist keyword)))
+                      (straight--put plist keyword value)))))))
           ;; The normalized recipe format will have the package name
           ;; as a string, not a symbol.
           (let ((package (symbol-name package)))
--- a/README.md
+++ b/README.md
@@ -1428,6 +1428,24 @@ will be looked up by default (see the section on [recipe
 lookup][#user/lookup]). You can see the default recipe for a package
 by invoking [`M-x straight-get-recipe`][#user/interactive].
 
+If `straight-allow-recipe-inheritance` is non-nil, then you only need
+to specify the components of the recpie that you want to override. All
+other components will still be looked up in the default recipe. In the
+example above, we are only interested in changing the `:fork`
+component. Therefore if `straight-allow-recipe-inheritance` is set,
+the recipe could be simplifed as follows:
+
+    (straight-use-package
+     '(el-patch :fork (:repo "your-name/el-patch")))
+
+or even simpler:
+
+    (straight-use-package
+     '(el-patch :fork "your-name/el-patch"))
+
+The `:files` keyword and all version control keywords support
+inheritance.
+
 To learn more, see the section on [the recipe format][#user/recipes].
 
 #### Additional arguments to `straight-use-package`


vyp:
What about when someone has `straight-allow-recipe-inheritance` set to `t`, but doesn't want say `:files` to be inherited in one particular recipe? (Assuming that when a recipe doesn't have `:files`, straight uses some default `:files` value?)

In that spirit, have you considered maybe having some way to enable/disable inheritance in a recipe itself, like say `:inherit` keyword?

CeleritasCelery:
> What about when someone has straight-allow-recipe-inheritance set to t, but doesn't want say :files to be inherited in one particular recipe?

All of these properties can be overridden. So if someone does not want `:files` to be inherited they can set it explicitly in their recipe. 

The whole idea behind this PR is that you should only need to override the properties you actually care about and not need to add redundant information that straight already knows. This does not prevent you from completely overriding a recipe If you so choose.

I feel like an `:inherit` keyword is unnecessary because this is the sensible default behavior and if you don’t like it, you can override the all recipes components explicitly (which is what you have to do now). I considered not even adding a variable to control this because all `straight-allow-recipe-inheritance` does it requires you to be explicit about everything (leading to a longer more error prone configuration).

vyp:
Yeah I understand the reason for this PR, I've been wanting this feature for a while too. :tada: Anyway I found out that I can just specify `:files :defaults` to set the default value for `:files`, so yeah anything like `:override` wouldn't be needed. :+1:

raxod502:
This sounds really lovely, it's a feature that has been sorely lacking for quite some time now. I will get to the review as soon as I can.

straight.el
```diff
@@ -208,6 +208,27 @@ profile (nil) will suffice without additional setup."
...
+
+can be simplifed to
+
+'(package :fork \"my-user/repo\")
+
+if the original recipe was
+
+'(package :host gitlab :repo \"other-user/repo\")"
+  :type 'boolean)
+
```
raxod502: An important shortcoming should be documented: namely, that if recipe inheritance is enabled then all recipe repositories need to be cloned right away even if you're specifying a custom recipe for a package that doesn't exist in any of them.

straight.el
```diff
@@ -208,6 +208,27 @@ profile (nil) will suffice without additional setup."
                 (alist :key-type symbol :value-type
                        (plist :key-type symbol :value-type sexp))))
 
+(defcustom straight-allow-recipe-inheritance t
+  "Non-nil allows missing recipe components to be inherited.
+These missing components inherit from the default recipe. The
+supported components are `:host', `:repo', `:files', and `:fork'.
+`:fork' will inherit the `:host' component from the original
+recipe origin.
```
raxod502: I would like significantly more explicitness about exactly what this means. For example, it's not clear at all from this description that if you specify a string for `:fork` it gets translated into a list `(:repo "...")` and then additional attributes are copied over if non-nil (is that the behavior? whatever the behavior is should be made explicit and defended).

straight.el
```diff
@@ -208,6 +208,27 @@ profile (nil) will suffice without additional setup."
...
+
+For example, the following recipe
+
+'(package :host gitlab :repo \"other-user/repo\" :fork (:host
+:gitlab :repo \"my-user/repo\"))
```
raxod502: Avoid curly quotes: https://github.com/raxod502/straight.el/blob/6649858bdb76d0c557bf46c93b3a2eb539cec750/straight.el#L1209

Having this indented nicely would also be desirable, I think.

straight.el
```diff
@@ -2866,6 +2887,27 @@ for dependency resolution."
           ;; override the default value (which is determined according
           ;; to the selected VC backend).
           ;;
+          (when straight-allow-recipe-inheritance
+            ;; To keep overridden recipes simple, the `:host',
+            ;; `:repo', `:files' can be inherited from the original
+            ;; recipe. Also `:fork' can inherit its host from origin,
```
raxod502: Does "from origin" mean "from the original recipe"?

straight.el
```diff
@@ -2866,6 +2887,27 @@ for dependency resolution."
...
+            ;; meaning that :fork "user/repo" is equivalent to :fork
+            ;; (:repo "user/repo")
+            (when-let ((fork (plist-get plist :fork))
+                       (_stringp (stringp fork)))
```
raxod502: I would prefer just `let` with `when` inside.

straight.el
```diff
@@ -2866,6 +2887,27 @@ for dependency resolution."
...
+                (cdr (straight-recipes-retrieve package))
+              (when (and host (not (plist-member plist :host)))
+                (straight--put plist :host host))
+              (when (and repo (not (plist-member plist :repo)))
+                (straight--put plist :repo repo))
+              (when (and files (not (plist-member plist :files)))
+                (straight--put plist :files files))
+              (when-let ((host host)
+                         (fork-plist (plist-get plist :fork))
+                         (_no-fork-host (not (plist-member fork-plist :host))))
```
raxod502: Large `when-let` seems a bit hard to follow to me. Can we just split it into multiple nested conditions which are simpler?

straight.el
CeleritasCelery: Expand on this. What would a user have to do to make sure recipe repos were cloned at the start?

straight.el
CeleritasCelery: Not sure what you mean by "avoid curly quotes".  Are you saying that I should be using single quotes in the documentation even though that is invalid syntax?

CeleritasCelery: nevermind, I see what you mean.

straight.el
raxod502: I don't mean they need to be cloned by the user. I mean that enabling recipe inheritance will cause `straight.el` to automatically clone them all. This isn't otherwise necessary if you provide an explicit recipe (as no recipe needs to be looked up).

I do not think this is a big problem, but I think it is worth mentioning.

CeleritasCelery: How is this triggered? With `straight-recipes-retrieve`?

CeleritasCelery:
I addressed your feedback. Please take another look.

straight.el
raxod502: Yes.

straight.el
```diff
@@ -2912,6 +2942,24 @@ for dependency resolution."
...
+          (when straight-allow-recipe-inheritance
+            ;; To keep overridden recipes simple, the
+            ;; `straight-vc-git-keywords' and `:files' can be
+            ;; inherited from the original recipe. Also `:fork' can
+            ;; inherit its host from the original recipe.
+            (let ((fork (plist-get plist :fork)))
+              (when (stringp fork)
+                (straight--put plist :fork `(:repo ,fork))))
+            (let ((default (cdr (straight-recipes-retrieve package))))
+              (dolist (keyword (cons :files (straight-vc-git-keywords)))
```
raxod502: Don't use `straight-vc-git-keywords`, instead use `straight-vc-keywords` with the appropriate `:type` for this recipe.

straight.el
```diff
@@ -2912,6 +2942,24 @@ for dependency resolution."
           ;; override the default value (which is determined according
           ;; to the selected VC backend).
           ;;
+          (when straight-allow-recipe-inheritance
+            ;; To keep overridden recipes simple, the
+            ;; `straight-vc-git-keywords' and `:files' can be
+            ;; inherited from the original recipe. Also `:fork' can
+            ;; inherit its host from the original recipe.
```
raxod502: This comment is out of date.

straight.el
```diff
@@ -208,6 +208,36 @@ profile (nil) will suffice without additional setup."
...
+  "Non-nil allows partially overriding recipes.
+If you override a recipe, every component that is not explicitly
+overriden will be searched for in original recipe. If found, that
+value will be added to the overriden recipe. This allows you to
+only override the recipe components you are interested in,
+instead of being required to override them all. The supported
+components are the ones listed by `straight-vc-git-keywords' and
+`:files'. Note that enabling this feature has the side effect
+that all recipe repos (i.e. melpa, elpa) will always be cloned,
+even if none of the recipes are used.
```
raxod502: I would say, "even if you explicitly specify all your recipes".

straight.el
```diff
@@ -208,6 +208,36 @@ profile (nil) will suffice without additional setup."
...
+
+The `:fork' keyword is handled specially. If its value is a
+string instead of a list, then it is assigned as the `:repo' of
+the fork. Also the fork recipe will inherit its `:host' component
+from the default recipe.
```
raxod502: Why should `:fork` only inherit `:host`, rather than all of `:repo`, `:branch`, and `:remote` as well?

CeleritasCelery:
Please review.

raxod502:
Thanks for the contribution!

CeleritasCelery:
Happy to contribute to one of my favorite packages.

PR #429, opened 2019-11-03
Do not display a prompt in non-interactive mode (fix #412)

### LINKED ISSUE #412: Support batch mode
I'm using [Topgrade](https://github.com/r-darwish/topgrade) to update my Emacs packages using `package.el`. I'm transitioning to straight and I'd love to integrate that into Topgrade. 

I think that it makes most sense that Topgrade will just run `(straight-thaw-versions)`. However, Topgrade runs Emacs in batch mode so it just hangs whenever straight asks the user for an action. I think that when `noninteractive` is `t` straight should just skip every repository that needs a user intervention and print a report saying what is skipped to some designated buffer.

r-darwish:


r-darwish:
This is probably not the ideal solution, but it is a solution. Perhaps we could do better in the future.

straight.el
```diff
@@ -408,6 +408,13 @@ function with the quoted name of the argument, or use t."
 
 ;;;;; Messaging
 
+(defun straight--output (&rest args)
+  (if noninteractive
+      (progn
+          (princ (apply #'format args))
+          (terpri))
+    (apply #'message args)))
+
```
raxod502: What does this change do?

straight.el
```diff
@@ -926,6 +933,8 @@ FUNC.
 
 ACTIONS later in the list take precedence over earlier ones with
 regard to keybindings."
+  (when noninteractive
+    (error prompt))
```
raxod502: I guess this is reasonable, but `prompt` should really be using a format of `"cannot display prompt in batch mode: %s"` or something.

straight.el
r-darwish: It prints to `stdout` in batch mode and prints to the minibuffer in interactive mode

r-darwish:
@raxod502 is there anything else that needs to be done?

raxod502:
Ack, please do not make pull requests against `master`. I have force-pushed to revert the change, and manually merged to the correct branch.

r-darwish:
Sorry. I'll open one for develop next time

raxod502:
No problem

Issue #442, opened 2019-12-06
changes to straight-format-timestamp lead to failed find command on start


cwhatley:
## What's wrong

Recent commit breaks on OSX.  See 

https://github.com/raxod502/straight.el/commit/dce8208f5d9d95127f0a47ff39d5a98eac1cfc7b

Contents of `*straight-process*`:

```
$ cd /
$ find /dev/null -newermt 2018-01-01\ 12\:00\:00

/dev/null

[Return code: 0]

$ cd /Users/cwhatley/.emacs.d/straight/repos/melpa/
$ find . -name .git -prune -o -newermt 2019-12-05\ 18\:54\:12.656686 -print

find: Can't parse date/time: 2019-12-05 18:54:12.656686

[Return code: 1]
```

## Directions to reproduce

Run with straight.el off develop branch.

### Version information

* Emacs version: 27.0.50 built from head on 12/3/2019
* Operating system: MacOS 10.15.2 beta

asf-stripe:
I've run into the same issue. I thought you could work around it for now by using this adjusted find-flavor determining function:

```elisp
(defun asf--straight--determine-find-flavor ()
  "Determine the best default value of `straight-find-flavor'.
This uses -newermt if possible, and -newer otherwise."
  (if (straight--check-call
       "find" "/dev/null" "-newermt" "2018-01-01 12:00:00.000000")
      `(newermt)
    nil))
(setq straight-find-flavor '(asf--straight--determine-find-flavor))
```

But unfortunately, straight then switches to `touch`, but using a GNU fileutils option, `-d`.

`brew install findutils` and overriding the PATH to pick up that gnu find worked, though: 

```elisp
;;; Use GNU find for https://github.com/raxod502/straight.el/issues/442:
(push "/usr/local/opt/findutils/libexec/gnubin" exec-path)
(setenv "PATH" "/usr/local/opt/findutils/libexec/gnubin:/usr/bin:/bin:/usr/local/bin")
```

raxod502:
Is there any syntax for specifying milliseconds on the BSD find? If not I'm going to be pretty sad. (In that case I'll just round up to the next second for the timestamps we save in the build cache.)

asf-stripe:
I believe neither `find` nor `touch` support millisecond timestamps: I couldn't find any docs on the time format in the `find` manpage, but `touch`'s has this:

     -t      Change the access and modification times to the specified time
             instead of the current time of day.  The argument is of the form
[...]

PR #443, opened 2019-12-07
Move "find" out to customizable variable. Re: #442

cwhatley:
Not a full solution for #442, but makes the workaround on MacOS/BSD systems less invasive.


manuel-uberti:
Does this mean an alternative solution (such as [fd](https://github.com/sharkdp/fd)) can be used?

Issue #393, opened 2019-06-25
Invalid switch


aviikc:
"git" "clone" "--origin" "origin" "--no-checkout" "<url to package.git" "~/.emacs.d/straight/repos/package/"

Packages are not downloading automatically 
i have to manually git clone each one.

raxod502:
Please provide more information. Otherwise, I have no way of knowing what is wrong. In particular, please provide an `init.el` file that I can put in an empty `user-emacs-directory` and reproduce the error, as well as versions of all relevant software.

That being said, perhaps your version of Git is too old.

aviikc:
[init file link](https://github.com/aviikc/emacs_settings/blob/master/code/emacs/.emacs)

The exact error is error: Failed to run "git"     
I'm on windows

*straight-process* buffer shows
`
$ cd "c:/emacs/emacs-26.2-x86_64/bin/"
$ "find" "/dev/null" "-newermt" "2018-01-01 12:00:00"

FIND: Invalid switch

[Return code: 2]

$ cd "c:/Users/Avik/.emacs.d/straight/repos/"
$ "git" "clone" "--origin" "origin" "--no-checkout" "https://github.com/emacs-straight/seq.git" "c:/Users/Avik/.emacs.d/straight/repos/seq/"

[File error while searching for program]
`

raxod502:
It sounds like you need to install a `find` program which has support for `-newermt`. It also sounds like you need to install Git, or perhaps add it to your PATH. How one does these things on Windows, I have no idea. Perhaps somebody who does use Emacs on Windows can help you with that.

garyo:
I can help with this -- just got it working on Win10, Emacs27 nightly (2019-10-10).

1. Install msys2 (this can work with older cygwin too, just adjust `exec-path` below)
2. Make sure msys2's `usr/bin` is in `exec-path` (usually `c:/msys2/usr/bin` but may be different depending on how you install msys2)
   1. Check `(executable-find "find")` returns the proper `find`, *not* `/Windows/system32/find.exe`
   2. Make sure you do this _before_ bootstrapping `straight.el` 
3. Because `find` is still run via `cmd.exe` which has weird quoting rules, you may need this patch:

```patch
diff --git a/straight.el b/straight.el
index 1dd3632..4e9c7bd 100644
--- a/straight.el
+++ b/straight.el
@@ -3510,7 +3510,7 @@ modified since their last builds.")
                           (append (list "-o"
                                         "-path"
                                         (format
[...]

PR #473, opened 2020-01-31
Use the appropriate quote in Windows (#393)

r-darwish:
This is probably not the ideal solution, but it makes stuff work until someone come up with a better one


r-darwish:
Will do. I’ll probably get to that in Sunday

r-darwish:
I pushed another fix. Seems to be okay now

r-darwish:
Sorry for bugging, but is there anything else that I should do before we can merge?

PR #477, opened 2020-02-10
Fix Package-Requires header parsing

### LINKED ISSUE #476: Regression prevents straight from resolving dependencies
## What's wrong

Straight fails to resolve (and thus, install) dependencies due to https://github.com/raxod502/straight.el/commit/260b388a96bb20af430e2935c92529a8bbde14cc. This new regexp leaves the cursor on the ~~colon~~ whitespace just after the colon in the `Package-Requires:` header. The old regexp would place the cursor directly on the opening parentheses of the package list (assuming the header is well-formed).

Then `(looking-at "(")` returns nil, and no dependencies are computed.

The following works around it:

```diff
                  (let ((case-fold-search t))
                    (re-search-forward "^;; *Package-Requires *:"))
-                 (when (looking-at "(")
+                 (when (looking-at " *(")
                    (straight--process-dependencies
                     (read (current-buffer)))))))))
```

Relevant to: https://github.com/hlissner/doom-emacs/issues/2517

EDIT: Corrected my diagnosis.

### Version information

* Emacs version: 26.3 (on Doom Emacs)
* Operating system: NixOS

hlissner:
Fixes #476

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -4088,7 +4088,7 @@ this run of straight.el)."
                  ;; has no dependencies.
                  (let ((case-fold-search t))
                    (re-search-forward "^;; *Package-Requires *:"))
-                 (when (looking-at "(")
+                 (when (looking-at " *(")
                    (straight--process-dependencies
                     (read (current-buffer)))))))))
     (straight--insert 1 package dependencies straight--build-cache)))


raxod502:
I installed an alternate fix that I think is more semantically correct; let me know if it doesn't work as expected.

PR #481, opened 2020-02-20
Allow for custom build directory

ThibautVerron:
Hello,

I use the same `.emacs.d` across different operating systems and emacs versions. With package.el, I was using customized elpa directories (like e.g. `elpa-windows-26.3`) to avoid conflicts, essentially downloading and compiling the packages as many times as I had environments.

With `straight` so far, I could get a satisfactory result using a custom directory of the form `straight-windows-26.3` in which `straight` itself was placed.

However, I came to realize that with the way `straight` separates repositories and builds, it could be made more efficient by just separating the `build` subfolders, not the whole `straight` folder. 

For example, if I want to apply modifications to a package, I would need to do it only once, not once per environment. And it would allow (I guess, I didn't test it) bootstrapping the configuration on a new environment, even without a good internet connection to download the packages. 

However, unlike `straight-base-dir`, there isn't a variable customizing where the builds go, so it requires changing a few functions after bootstrapping, recompiling `straight` in the new build directory, and reloading the new functions.

The proposed patch aims at making the process more straightforward. It adds a variable `straight-build-subdirectory` which is used instead of `"build"` in the two functions `straight--build-dir` and `straight--build-file`, and patches `straight--compute-dependencies` to use those functions instead of a hard-coded path (I believe that was a minor bug regardless).

Thanks!


ThibautVerron:
My system seriously messed the commits, I'm closing and will reopen after cleaning.

raxod502:
You might be interested in `git push --force`, which will allow you to amend an existing pull request instead of needing to creating a new one.

ThibautVerron:
I don't have a lot of rules, but I try to stick with "don't amend a pushed commit". Especially when it concerns a repo which is not mine and when I don't understand how the breakage happened in the first place. (I don't know if simply amending the request would have worked, I would have really needed to undo at least the last commit which deleted half the content of the repository.)

That being said, there was probably a better solution than nuking both the repo and the working copy and starting afresh.

Sorry for the spam with the duplicate pull requests.

raxod502:
> I try to stick with "don't amend a pushed commit"

That is a good rule to follow in most cases, but I do not think it makes sense to follow it for pull requests, because it requires creating additional pull requests when you want to change something.

> it concerns a repo which is not mine

Okay, but you have no push access to this repository and therefore cannot possibly break anything.

> Sorry for the spam with the duplicate pull requests.

No worries. I have been there before.

Issue #484, opened 2020-02-21
Making different versions of emacs cohabitate


ThibautVerron:
(Follow-up to https://github.com/raxod502/straight.el/pull/482)

It turns out that having a different "build" directory for each target version is not enough, because the file `build-cache.el` also depends on the emacs version.

An easy fix would be to replace the definition of `straight--build-cache-file` with

```
(defun straight--build-cache-file ()
  "Get the file containing straight.el's build cache."
  (straight--build-file "build-cache.el"))
```

so that the cache itself is in the "build" directory.

But changing it like that would make all installations recompile all their packages when updating straight, is it acceptable?

Otherwise, what would be the best course of action? Making it a setting with the current behavior the default? Or adding some code copying the current cache if it exists?

raxod502:
> the cache itself is in the "build" directory.

That sounds reasonable; however, I'd prefer to avoid placing the build cache directly in the `build` directory, because it will then be very hard to find among all the package directories (and, in an edge case, could even conflict with a package called `cache.el`).

What about simply adding a user option analogous to `straight-build-dir`, called `straight-build-cache-file`? Then if you wish to enable version isolation, you can set both of those variables. We can later add a feature to `straight.el` to do this automatically if it proves a popular use case.

> changing it like that would make all installations recompile all their packages when updating straight, is it acceptable?

With my proposed course of action above, there wouldn't be any migration necessary. However, since you asked, my general policy is to not worry about migration code, as it quickly gets very complicated for a relatively high-velocity project such as `straight.el`. There are many reasons the build cache might need to get rebuilt (check [`straight--load-build-cache`](https://github.com/raxod502/straight.el/blob/ed54b626b0faec36fdaff30c72351d5fb41a6786/straight.el#L3286-L3321)), and in all of them we just chuck it out and start over, even if the case could be handled by a migration.

ThibautVerron:
> That sounds reasonable; however, I'd prefer to avoid placing the build cache directly in the build directory, because it will then be very hard to find among all the package directories (and, in an edge case, could even conflict with a package called cache.el).

That's a very good point. 

> What about simply adding a user option analogous to straight-build-dir, called straight-build-cache-file? Then if you wish to enable version isolation, you can set both of those variables. We can later add a feature to straight.el to do this automatically if it proves a popular use case.

Another possibility, without placing the cache file in the build directory, would be to create the name of the cache file based on the name of the build directory. 

[...]

PR #485, opened 2020-02-21
Variable for the name of the build cache

ThibautVerron:
Cf https://github.com/raxod502/straight.el/issues/484


ThibautVerron:
Is it okay like this?

Issue #492, opened 2020-03-04
Use gccemacs to compile packages in parallel


jjzmajic:
Something like `(native-compile-async "~/.emacs.d/straight/build" 4 t)` would be great. Plus it solves the problem of async compilation for people building from source. Once it lands in master (and it should around Emacs 28, the feature branch seems very mature) it will be the default anyway.

PS: [This](https://akrl.sdf.org/gccemacs.html) is the feature branch I am talking about.

raxod502:
Just to make sure I am understanding:

* this feature provides a way to compile Elisp files to a form that can be executed more efficiently, so it's desirable to make it available as an option for `straight.el`
* although we could already implement asynchronous byte-compilation, using this feature would make that implementation simpler in `straight.el`

Is that right?

If so, I'd point out that there are two levels:

* byte-compile the files of a single package asynchronously, which is not a problem
* byte-compile the files of multiple packages in parallel, which requires the implementation of a new interface by which multiple packages can have their recipes specified at the same time; otherwise `straight.el` doesn't know ahead of time what packages will need to be built or what their recipes are

alphapapa:
Basically, something like this:
```el
(use-package comp
  :config
  (defcustom native-compile-async-jobs
    (or (ignore-errors
	  (string-to-number (shell-command-to-string "nproc")))
	1)
    "How many jobs to use."
    :type 'integer)

  (setf native-compile-async-jobs 2)

  (defun ap/package-native-compile-async (package &optional all)
    "Compile PACKAGE natively, or with prefix ALL, all packages."
    (interactive (list (unless current-prefix-arg
			 (completing-read "Package: " (mapcar #'car package-alist)))))
    (let* ((directory (if package
			  (file-name-directory (locate-library package))
			package-user-dir)))
      (native-compile-async directory native-compile-async-jobs t)
      (load-library (symbol-name package)))))
```
However, adding such a feature to Straight is probably a little premature.  I hope the branch is merged for Emacs 28, but let's wait and see.  :)

raxod502:
Is this solved by #507?

jjzmajic:
Yes, sir!

Issue #334, opened 2018-12-15
Prompt for SSH/HTTPS passphrase in minibuffer


qleguennec:
```
(use-package private-repo
  :straight (:host github :repo "qleguennec/private-repo"))
```

Results in:

```
$ cd /home/void/.emacs.d/straight/repos/
$ git clone --origin origin --no-checkout https\://github.com/qleguennec/private-repo.git private-repo

Cloning into 'private-repo'...
fatal: could not read Username for 'https://github.com': No such device or address

[Return code: 128]
```

Not sure if that's a bug or something that could be implemented.
Either way, that would be pretty useful to have.

raxod502:
If you

```elisp
(setq straight-vc-git-default-protocol 'ssh)
```

then you can take advantage of `ssh-agent` authentication. With HTTPS, you can use `git-credential-cache`.

Unfortunately, if the cache has timed out, you still get an error like this. I agree that it's not desirable behavior. Probably the right way to handle this is to do as Magit does and just connect stdin for all Git commands to the terminal. That way the user can provide a passphrase when one is needed.

raxod502:
:sparkles: :sparkles: :sparkles:

raxod502:
Another regression, this one even more incomprehensible. This makes me extremely sad.

bestlem:
> If you
> 
> ```emacs-lisp
> (setq straight-vc-git-default-protocol 'ssh)
> ```
> 
> then you can take advantage of `ssh-agent` authentication. With HTTPS, you can use `git-credential-cache`.
> 
> Unfortunately, if the cache has timed out, you still get an error like this. I agree that it's not desirable behavior. Probably the right way to handle this is to do as Magit does and just connect stdin for all Git commands to the terminal. That way the user can provide a passphrase when one is needed.

When you do this chnage to straight-vc-git-default-protocol not you have to also change the [remote] entries in the .git/config from https to ssh for those repos you already have.

raxod502:
You can also accomplish this by running `M-x straight-normalize-all` with `straight-vc-git-force-protocol` non-nil.

PR #507, opened 2020-04-16
Support native compilation of packages

flatwhatson:
This adds support for native compilation (aka. "gccemacs"), where
byte-compiled elisp is compiled to native code. Modules will be native
compiled when they are byte-compiled, unless `:no-native-compile` or
`straight-disable-native-compilation` are set.

Has no effect on emacs versions without the `native-compile` function.

Possibly addresses #492.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,10 @@ chat][gitter-badge]][gitter]
         * [`check-on-save`](#check-on-save)
         * [`watch-files`](#watch-files)
     + [Customizing how packages are built](#customizing-how-packages-are-built)
+        * [Autoload generation](#autoload-generation)
+        * [Byte compilation](#byte-compilation)
+        * [Native compilation](#native-compilation)
+        * [Symbolic links](#symbolic-links)
     + [Customizing how packages are made available](#customizing-how-packages-are-made-available)
     + [Hooks run by `straight-use-package`](#hooks-run-by-straight-use-package)
   * [The recipe format](#the-recipe-format)
@@ -1633,6 +1637,8 @@ which do not bundle executable Lisp code. (Make sure to use
 [`straight-use-recipes`][#user/lookup/repos] for registering recipe
 repositories.)
 
+##### Autoload generation
+
 By specifying a non-nil value for the `:no-autoloads` attribute in a
 package's recipe, you may prevent any autoloads provided by the
 package from being generated and loaded into Emacs. This is mostly
@@ -1643,6 +1649,8 @@ you *really* know what you're doing). You can also customize the
 variable `straight-disable-autoloads` to effect this change on all
 recipes which do not explicitly specify a `:no-autoloads` attribute.
 
+##### Byte compilation
+
 By specifying a non-nil value for the `:no-byte-compile` attribute in
 a package's recipe, you may inhibit byte-compilation. See [this
 issue][#357] for discussion of why this might be useful. You can also
@@ -1650,14 +1658,25 @@ customize the variable `straight-disable-byte-compilation` to effect
 this change on all recipes which do not explicitly specify a
 `:no-byte-compile` attribute.
 
+##### Native compilation
+
+Experimental support for native compilation of Emacs Lisp code is
+currently under development in the `feature/native-comp` branch of the
+official Emacs repository (see [gccemacs][gccemacs]). When running on
+this version of Emacs, `straight.el` will perform native compilation
+of packages.
+
 By specifying a non-nil value for the `:no-native-compile` attribute
 in a package's recipe, you may inhibit native compilation. You can
 also customize the variable `straight-disable-native-compilation` to
 effect this change on all recipes which do not explicitly specify a
-`:no-native-compile` attribute. Native compilation requires
-byte-compilation, so `:no-byte-compile` and
-`straight-disable-byte-compilation` also have the effect of inhibiting
-native compilation.
+`:no-native-compile` attribute.
+
+Native compilation requires byte-compilation, so `:no-byte-compile`
+and `straight-disable-byte-compilation` will also disable native
+compilation.
+
+##### Symbolic links
 
 Usually, `straight.el` uses symbolic links ("symlinks") to make
 package files available from the build directory. This happens when
@@ -2964,6 +2983,7 @@ the version lock file. This addresses issues [#58], [#66], and [#294].
 [epkg]: https://github.com/emacscollective/epkg
 [epkgs]: https://github.com/emacsmirror/epkgs
 [forge]: https://github.com/magit/forge
+[gccemacs]: http://akrl.sdf.org/gccemacs.html
 [git]: https://git-scm.com/
 [git-credential-cache]: https://git-scm.com/docs/git-credential-cache
 [gitter-badge]: https://badges.gitter.im/raxod502/straight.el.svg
--- a/README.md
+++ b/README.md
@@ -1650,6 +1650,15 @@ customize the variable `straight-disable-byte-compilation` to effect
 this change on all recipes which do not explicitly specify a
 `:no-byte-compile` attribute.
 
+By specifying a non-nil value for the `:no-native-compile` attribute
+in a package's recipe, you may inhibit native compilation. You can
+also customize the variable `straight-disable-native-compilation` to
+effect this change on all recipes which do not explicitly specify a
+`:no-native-compile` attribute. Native compilation requires
+byte-compilation, so `:no-byte-compile` and
+`straight-disable-byte-compilation` also have the effect of inhibiting
+native compilation.
+
 Usually, `straight.el` uses symbolic links ("symlinks") to make
 package files available from the build directory. This happens when
 `straight-use-symlinks` is non-nil, the default. On Microsoft Windows,
--- a/straight.el
+++ b/straight.el
@@ -4284,7 +4284,7 @@ repository."
             (message-log-max nil))
         (native-compile-async
          (straight--build-dir package)
-         'recursively)))
+         'recursively 'late)))
     (straight--wait-for-async-jobs)))
 
 (defun straight--pending-async-jobs ()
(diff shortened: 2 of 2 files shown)


flatwhatson:
The main challenge with native compilation is that it's *slow*.  The `comp` module uses a job queue and a pool of workers to maximize concurrency, but this is only effective if you keep the queue topped-up.  Any approach where `straight-use-package` blocks until the package is compiled won't perform well.

I've provided a "low-level" solution for this with `straight--wait-for-async-jobs`, so you can do something like the following:
```lisp
;; install the packages without blocking
(let ((straight--wait-for-async-jobs nil))
  (straight-use-package 'foo)
  (straight-use-package 'bar)
  ...)
;; block until compilation has finished
(straight--wait-for-async-jobs)
```

This could potentially be wrapped up into a nicer API:
```lisp
(straight-with-async-compilation
  (straight-use-package 'foo)
  (straight-use-package 'bar)
  ...)
```

Maybe this could even be extended to support async byte compilation?

I'm interested in any and all suggestions!

straight.el
```diff
@@ -4252,6 +4257,61 @@ repository."
          (straight--build-dir package)
          0 'force)))))
 
+;;;;; Native-compilation
```
raxod502, line 4260:
```suggestion
;;;;; Native compilation
```

straight.el
```diff
@@ -4252,6 +4257,61 @@ repository."
...
+      (let ((inhibit-message t)
+            (message-log-max nil))
+        (native-compile-async
+         (straight--build-dir package)
+         'recursively)))
+    (straight--wait-for-async-jobs)))
+
+(defun straight--pending-async-jobs ()
+  "How many async compilation jobs are queued or in-progress.
+Returns nil if there are no pending jobs."
```
raxod502, line 4292: Why does this return nil instead of zero when there are no jobs?

straight.el
```diff
@@ -4252,6 +4257,61 @@ repository."
...
+
+(defun straight--wait-for-async-jobs ()
+  "Block until all async compilation jobs have completed (maybe).
+Blocking can be suppressed by setting `straight--wait-for-async-jobs' to nil."
+  (when (and (fboundp 'comp-async-runnings)
+             straight--wait-for-async-jobs)
+    (let ((inhibit-message t)
+          (message-log-max nil))
+      (while (not (zerop (comp-async-runnings)))
+        (sleep-for 0.1)))))
```
raxod502, line 4311: Have you found this to be reliable and not too bad for performance? I have tried to write code in `straight.el` before that synchronously waits for an async job to complete, and it's been quite difficult. If this works robustly, maybe I can steal the approach for my other work too (specifically #334).

README.md
```diff
@@ -1650,6 +1650,15 @@ customize the variable `straight-disable-byte-compilation` to effect
...
 
+By specifying a non-nil value for the `:no-native-compile` attribute
+in a package's recipe, you may inhibit native compilation. You can
+also customize the variable `straight-disable-native-compilation` to
+effect this change on all recipes which do not explicitly specify a
+`:no-native-compile` attribute. Native compilation requires
+byte-compilation, so `:no-byte-compile` and
+`straight-disable-byte-compilation` also have the effect of inhibiting
+native compilation.
+
```
raxod502, line 1661: I think this section of the README should include enough info to explain what gccemacs is, how people get it, and what its relation to GNU Emacs is. Otherwise, how does one tell how to use the feature?

straight.el
flatwhatson: This is an attempt to make the function more ergonomic, ie.
```lisp
(if (straight--pending-async-jobs) ...)
```
vs.
```lisp
(if (not (zerop (straight--pending-async-jobs))) ...)
```
It could be considered "too cute", I'm happy to just always return the count if you'd prefer.

straight.el
flatwhatson: I can't make any great claims, but it seems to do the job!

straight.el
flatwhatson: Fixed!

straight.el
flatwhatson: I've changed this routine to always return a number.

README.md
flatwhatson: Good point, I've added some additional info, at least it should point people in the right direction.

straight.el
```diff
@@ -4252,6 +4257,60 @@ repository."
...
+      (let ((inhibit-message t)
+            (message-log-max nil))
+        (native-compile-async
+         (straight--build-dir package)
+         'recursively 'late)))
+    (straight--wait-for-async-jobs)))
+
+(defun straight--pending-async-jobs ()
+  "How many async compilation jobs are queued or in-progress.
+Returns nil if there are no pending jobs."
```
raxod502, line 4292:
```suggestion
  "How many async compilation jobs are queued or in-progress."
```

straight.el
raxod502: Yeah I do like the ergonomics, but it does violate the "principle of least surprise" rather badly so I think your revised version is better.

alphapapa:
@raxod502 @flatwhatson  Be advised that, AFAIK, Andrea is still developing the deferred compilation features.  For example, he recently added a feature that defers compilation until a library is loaded; when that happens, compilation is started, and as each unit is compiled, its native form is loaded to replace the loaded, byte-compiled form.

Having done that, it would probably be only a few lines of code to provide for the case in which:

1.  Package is installed and byte-compiled.
2.  Package's native compilation is begun.
3.  Package is loaded.
4.  Package's native compilation is finished (this step would be per-unit, i.e. per Elisp file).

In other words, it's probably not necessary to do this in Straight.  If it currently is, it would probably be better to talk to Andrea about supporting this in `comp.el` (or I might be able to help with that, as I contributed some small changes to it).  And it may be inadvisable to try to manage the queue in Straight, especially since these features are still under development (i.e. this code might break as the native-comp branch is developed).

flatwhatson:
Understood, I think it's clear that this is very much "experimental" and will probably need review as the feature development progresses.

The deferred compilation feature is very nice, and you're right that straight needs to do very little (if anything) when used that way.  The functionality I've added here is for a different use-case; pre-compiling packages so they're *not* building in the background.

alphapapa:
> The functionality I've added here is for a different use-case; pre-compiling packages so they're _not_ building in the background.

That's what I'm talking about.  The difference between starting compilation at package-load time and starting compilation at package-install time is negligible.  Either way, we want the files to be compiled in a non-blocking way, and when each compilation unit is finished compiling, if its byte-compiled form is already loaded, we want its native-compiled form to be loaded instead.  Support for that should be done in `comp.el` and `package.el`, not in Straight.

flatwhatson:
We use the `native-compile-async` function provided by `comp.el` to request compilation of all elisp files in the package directory.  This ensures that all features provided by that package will be compiled up-front.  This approach also allows us to explicitly disable compilation of a package using its recipe.

In contrast, relying on the deferred compilation system would *not* guarantee that everything was compiled, and does *not* provide a way to blacklist compilation of a package.  Many packages are split into a number of modules and use autoloads to load only when needed, so for infrequently-used modules package-load time could be much later than package-install time.

I don't think anything has been added here which belongs upstream, but if you can point to something specific I'd be very happy to contribute!

alphapapa:
> In contrast, relying on the deferred compilation system would _not_ guarantee that everything was compiled, and does _not_ provide a way to blacklist compilation of a package. Many packages are split into a number of modules and use autoloads to load only when needed, so for infrequently-used modules package-load time could be much later than package-install time.

There seems to be a communication problem here.  Please read my comments carefully.  I am aware of how the deferred compilation system currently works, and I am aware of how you want to compile packages at install-time rather than waiting for load-time.  My point is that a few simple changes to `comp.el` could support both use cases, which would also allow AOT compilation using other package managers like `package.el`, and it would mean that Straight would not need to maintain this code itself.

flatwhatson:
OK, that makes more sense.  Thanks for your patience!

I believe that our use-case is already well-supported by `comp.el`, and that similar functionality could be added to `package.el` without too much effort.  I don't use `package.el`, so am probably not the right person to do this work.

In the simplest form, we're doing the following:
```lisp
(native-compile-async "/path/to/package1" 'recursively)
(native-compile-async "/path/to/package2" 'recursively)
...
(while (not (zerop (comp-async-runnings)))
  (sleep-for 0.1))
```

Perhaps `comp.el` could add a more convenient routine for blocking until complation has finished?  Otherwise I'm not sure how things could be improved from that end.

alphapapa:
Blocking until compilation is finished seems like a very bad idea, because that can mean many minutes of waiting.  That's why it's done asynchronously, and that's the point of deferred compilation: the user can continue using Emacs, including using the new package immediately, and the native-compiled units replace the byte-compiled ones when ready.

AFAIU, the only difference in the desired behavior here is to begin compilation immediately, rather than waiting until a package is loaded.  And that can be done very easily by calling `native-compile-async`.  The remaining issue is to ensure that, if a compilation unit's byte-compiled form is loaded before the native-compiled form is ready, the native-compiled form is loaded automatically when compilation finishes.

AFAIR, this is currently handled by setting a variable or argument when compilation begins.  But since the package's files would not already be loaded when Straight asks them to be compiled, that would not happen, so if the user loaded the package's features before compilation finishes, the native-compiled forms would not be loaded automatically.

So, again, I think this would be trivial to handle cleanly in `comp.el`.  I think Straight shouldn't be doing anything more than `(native-compile-async "/path/to/package1" 'recursively)`.  And it definitely shouldn't be blocking until compilation finishes.

alphapapa:
> AFAIR, this is currently handled by setting a variable or argument when compilation begins.

I see that you did this in https://github.com/raxod502/straight.el/pull/507/commits/130b8a24c99e8f633233083c7e58091bcc0ad880.  So:

1.  From what I can tell, doing that forces the file to load when compilation finishes, even if the user hasn't already loaded it.  That's not the end of the world, but it shouldn't be the case.  This is what should be handled by minor improvements in `comp.el`.
2.  Why would you ever want to block while waiting for compilation?  That seems to defeat the purpose of asynchronous compilation.

Since this PR is merged already, I'll open an issue.

Issue #511, opened 2020-04-24
Issues with new native-comp support


alphapapa:
As I mentioned in comments on https://github.com/raxod502/straight.el/pull/507, the changes merged there are somewhat inadvisable.

* It makes no sense to block while native compilation happens.  That defeats the purpose of the asynchronous compilation that Andrea has implemented.  It means minutes of waiting while Emacs is blocked for no good reason.
* A few minor improvements to `comp.el` would support the desired behavior in which packages would be compiled at install-time (rather than deferring compilation until load-time), while not forcing packages to be loaded when compilation finishes (unless the user has already loaded their byte-compiled forms while waiting for native compilation).
* As well, adding code for this may be premature, because symbols may be renamed at any time.  I already contributed a patch to `comp.el` that renamed some, and some other symbols, including some referenced in that PR, should also be renamed for clarity and consistency.  I would expect such renaming and refactoring to take place at least a few times before the branch is merged.

So I would generally recommend that we make the necessary, minor improvements to `comp.el` to support this use case (which would also benefit `package.el`), and that Straight should refrain from adding references to `comp.el` symbols until it is merged to Emacs `master`.

alphapapa:
I did not intend for this to be labeled `bug`, but that seems to have happened automatically, and I can't change it.

gagbo:
I don't totally understand the issue here. Is it not okay if straight provides a handle to compile AOT ?

I mean, *as a user of straight.el* you can also pass the `:no-native-compile` flag in the recipe, and then enable deferred compilation in order to enjoy asynchronous compilation happening as provided in `comp.el`if I understand #507 correctly

flatwhatson:
Blocking until compilation has finished is *optional*, though I'll grant that the API is currently awkward.  Setting `straight--wait-for-async-jobs` to `nil` will completely prevent blocking.  I agree that blocking might not be desirable in the common case, so maybe this should be the default?

I added blocking because I call straight from a command-line process so that package installation *never* impacts my regular use of emacs.  Without blocking, this process exits immediately and nothing is compiled.

As for depending on work-in-progress APIs, I think if there is breakage here it will only affect people who are a) using a bleeding-edge build of emacs and b) using the develop branch of straight.  These crazy folks deserve any breakage!

alphapapa:
> I don't totally understand the issue here. Is it not okay if straight provides a handle to compile AOT ?

[...]

PR #512, opened 2020-04-25
Don't block for native compilation

flatwhatson:
Blocking until native compilation has completed is not necessary or
desirable for standard usage of `straight.el`, so has been removed.

This has the added benefit of reducing our exposure to the API of
`comp.el` while it's under active development.

Related to #511 
Related to #507

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -4289,10 +4289,12 @@ individual package recipe."
                                  straight-disable-native-compilation))))
 
 (defun straight--native-compile-package (recipe)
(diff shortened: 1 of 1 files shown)


blak3mill3r:
Is there still an optional way to block until native comp finishes?

raxod502:
I'm guessing no, although I am not an authority on the matter. If it's desired, somebody could re-add the relevant functions to `straight-x.el`.

PR #513, opened 2020-04-26
Ensure directory doesn't exist before full clone

flatwhatson:
This is necessary to prevent a clone error when we try to perform a
shallow clone, fail, and recurse with depth `full`.  The shallow clone
attempt leaves a non-empty repo-dir which git refuses to clobber.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -2185,10 +2185,12 @@ clone of everything."
            "--depth" (number-to-string depth)
            "--branch" branch))
       ;; Fallback for dumb http protocol.
-      (error (straight-vc-git--clone-internal :depth 'full
-                                              :remote remote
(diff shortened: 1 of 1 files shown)


straight.el
```diff
@@ -2159,6 +2159,7 @@ clone of everything."
   (cond
    ((eq depth 'full)
     ;; Clone the whole history of the repository.
+    (delete-directory repo-dir 'recursive)
```
raxod502, line 2162: I feel like this would be better placed at the end of the `condition-case`, just before recursing on `straight-vc-git--clone-internal`. Thoughts?

flatwhatson: Agreed!

PR #539, opened 2020-06-20
Set load-in-progress to t when processing autoloads

xzz53:
Several packages (e.g. realgud [1]) expect `load-in-progress` variable
to be set to `t` when their autoloads are evaluated, but `straight`
fails to do so in `straight--activate-package-autoloads`, thus
breaking build of such packages. This patch addresses the issue by
explicitly setting `load-in-progress` to `t` in
`straight--activate-package-autoloads`.

[1] https://github.com/realgud/realgud/blob/24c02f07c4ae6610278da1f04a2479dc3d9b6bdf/realgud.el#L179


raxod502:
Thanks!

xzz53:
That was fast! :)

Issue #279, opened 2018-05-29
MELPA recipes whose repos have custom default branches


raxod502:
E.g. the recipe for `ensime` is

```
(ensime :fetcher github
	:repo "ensime/ensime-emacs")
```

from this there is no way to tell that the correct branch to check out is `2.0` and not `master`. Thus the user must override the recipe in their init-file, which is undesirable (all MELPA recipes should work out of the box).

I don't know how to fix this without introducing more problems.

See also https://github.com/sjrmanning/.emacs.d/issues/1.

hartzell:
I'm coming at this from a position of relative naïveté ...

If you clone the repository, the branch that is checked out by default is the default branch.  In the case of ensime, 2.0 is the default branch, so you're good to go.  Just don't try to check out something else unless you're told to.

If there's some repo out there that wants you to use a branch that they don't have set as their default, then there's more work to be done...

In theory you could hit API and ask for the default branch, but you'd probably need a token and etc...  https://developer.github.com/v3/repos/#get

Actually,  it seems I can ask for it without authenticating.  You just have to figure out whether to ask for an `/orgs/...` repo or a `/users/...` repo.  Not sure that's really viable.

```
curl https://api.github.com/orgs/ensime/repos
```

SkySkimmer:
`git ls-remote --symref "$remote" HEAD` will produce something like
```
ref: refs/heads/trunk   HEAD
12aa0caf5a4a4f595f3bc3af5631a674bc8cece4        HEAD
```
(on https://github.com/ocaml/ocaml)

So maybe you could have a special value `:branch :default`. The main issue is that you need a network call to figure out if you're on the right branch, or some special purpose cache.

raxod502:
Checking the remote `HEAD` is a good idea, @SkySkimmer. I think that will work. Updating the code will entail looking at everywhere `straight-vc-git-default-branch` is used and having it do this instead. (Although I think we still want to have `straight-vc-git-default-branch`, e.g. in the case of a local repository with no remotes.

xendk:
No way that it can be figured out for local repositories too? I foresee a imminent future where the name of the default branch is a pretty fluid concept.

raxod502:
Are you referring to this comment?

> e.g. in the case of a local repository with no remotes

What I meant by that was a package that is taken from a repository that only exists on your local machine, with nothing on the Internet anywhere. This is something supported by `straight.el`, so by definition there is no concept of a default branch for that configuration.

[...]

PR #543, opened 2020-07-12
Fetch name of default branch from remote

ahyatt:
This makes it possible to deal with an ecosystem that may be diverging
away from the "master" branch naming. This change provides a different
default branch name, "main", and provides a list of alternate branch
names that can be tried.

In the various git operations, these alternate branches are tried
until one works.


ahyatt:
Should fix https://github.com/raxod502/straight.el/issues/279.  Warning: I'm not that familiar with this package.  I tested this out somewhat, and it was able to clone & update emacs-websocket, only getting the `main` branch, but that's about it so far.

straight.el
```diff
@@ -1415,10 +1415,16 @@ For built-in packages, this is always nil."
 
 ;;;;; Git
 
-(defcustom straight-vc-git-default-branch "master"
+(defcustom straight-vc-git-default-branch "main"
   "The default value for `:branch' when `:type' is symbol `git'."
   :type 'string)
 
+(defcustom straight-vc-git-alternate-branches'("master")
```
raxod502, line 1422:
```suggestion
(defcustom straight-vc-git-alternate-branches '("master")
```

straight.el
```diff
@@ -2242,6 +2258,23 @@ argument is not part of the VC API."
...
+  (let ((default-directory (straight--repos-dir repo)))
+    (straight-vc-git--narrow-branch branches)))
+
+(cl-defun straight-vc-git--narrow-branch (branches)
+  "Narrow possible branches to a single choice.
+Return the real branch name, or nil if none were found. BRANCHES
+can be a string, in which case it is assumed to be correct
+already, or a list, in which case we try to find the correct one."
+  (cond
+   ((cl-typep branches 'string) branches)
```
raxod502, line 2272:
```suggestion
   ((stringp branches) branches)
```

straight.el
```diff
@@ -2242,6 +2258,23 @@ argument is not part of the VC API."
...
+   ((cl-typep branches 'list)
```
raxod502, line 2273:
```suggestion
   ((listp branches)
```

straight.el
```diff
@@ -1415,10 +1415,16 @@ For built-in packages, this is always nil."
 
 ;;;;; Git
 
-(defcustom straight-vc-git-default-branch "master"
+(defcustom straight-vc-git-default-branch "main"
```
raxod502, line 1418: Can we leave the default at master until at least a version of Git which defaults to main has shipped to, say, some release of Ubuntu? That would still allow repositories using `main` to be used out of the box. I fear that changing the default before everyone's software has changed will lead to confusion.

straight.el
```diff
@@ -2242,6 +2258,23 @@ argument is not part of the VC API."
 If RECIPE does not configure a fork, do nothing."
   (straight-vc-git-fetch-from-remote recipe 'from-upstream))
 
+(cl-defun straight-vc-git--narrow-branch-in-repo (branches repo)
+  "Narrow possible BRANCHES to a single choice, in given REPO."
```
raxod502, line 2262: I believe this should be "local-repo", not "repo" (the latter refers to a Git URL).

straight.el
ahyatt: The reason I did it this way is that repositories (like mine), which switched from `master` to `main`, had to recreate `master` for compatibility reasons.  But `main` is the main branch, and it'd be nice if we could have it take precedence.

I agree that it's confusing, though - perhaps better documentation on these variables might help.

Alternately, we can just deprecate the default-branch variable and replace it with the alternate-branches variable.  In runtime we can make sure the default-branch has the least precedence.  What do you think?

straight.el
ahyatt: Done.

straight.el
raxod502: The main thing I'm worried about is certain edge cases. For example, when shallow cloning is enabled (this will likely be the default in the future), only a single branch is fetched. If that single branch is wrong (e.g. `main` when the repository still uses `master`), then we have a problem. In principle we could then fetch additional branches if the first one isn't found, but that kind of defeats the point of shallow cloning a bit.

I think that the _right_ solution to this problem, overall, is to implement the idea of https://github.com/raxod502/straight.el/issues/279#issuecomment-392834194 to detect the default branch automatically. This pull request is just a temporary workaround -- which is fine, but it shouldn't break existing use cases.

ahyatt: I think any ordering we have has the possibility of being wrong.  My guess is that `main` should take precedence over `master`, because I've seen that happen in real repositories, but haven't seen the reverse.  However, I'm happy to leave the default branch name as it was, if so I'd prefer to change it as I suggested in my last suggestion in my previous email, so that it's the lowest precedence.  That will make it so that my repository is correctly retrieved.

raxod502: It's just that this approach cannot possibly solve the problem properly. What we actually need is for `straight.el` to determine the default branch by checking the remote. Without that capability, any repository whose default branch appears later in the list will receive degraded functionality (e.g. for shallow clones). I don't think it's acceptable to introduce such degraded functionality for the majority of current repositories.

I also understand that by keeping master at the head of the list, it somewhat defeats the point because then backwards-compatibility master branches will be picked up instead of newer main branches. So perhaps we shouldn't implement this at all, and instead insist on a proper solution that will not introduce degraded functionality for any repository. It is up to you.

ahyatt: That makes sense, thanks.  Let me think about the proper solution a bit more and see how easy it would be to implement.

ahyatt: Sorry for the delay, I'm looking at this again.  Perhaps an even better solution is to not even have the concept of a default branch name, and look up what `origin/HEAD` maps to, in order to find what the default name for the repository is?

What do you think of that solution?

raxod502: Absolutely. That is exactly what I suggest in https://github.com/raxod502/straight.el/issues/279#issuecomment-392834194, and I do not think it has any downsides aside from the inconvenience of the implementation and some more reliance on network connectivity.

ahyatt:
I'm in the middle of reworking this.  I'm not sure if you had something in mind, but I'm guessing the proper thing is to always default to using the remopte branch that corresponds to the locally checked out branch, regardless of what the default branch is.

Interestingly, I was playing around with this, and noticed that the org-roam package had `develop` as the default remote branch, not `master`.  Default branch is probably the best we can do for now, but it would for git to have a concept of release branch.  As is, you never know if you are going to get a branch that is stable or not.

ahyatt:
I'm not sure why the repository you are looking up doesn't have a `HEAD` reference.  Do you find this on many repositories? As I understand it, the purpose of `origin/HEAD` is to determine what branch is checked out when you clone the repository.  So when you clone this particular repository, what branch is checked out?  And I wonder how git makes that determination?

In this case, what does `git remote show origin` produce?   That's another way we can find the HEAD branch.  

About using whatever branch is checked out, right now not providing `:branch` will have that effect.  To me, that makes the most sense - the checked out branch is the one that will actually affect emacs when starting, so it seems like the one that people will want to manipulate.  Another alternative is finding the remote branch, and using the correspondingly named local branch when `:branch` is not set.  As you suggest, we can do both by using a special value of `:branch` to distinguish these behaviors.  However, I want to make sure which behavior makes the most sense to use for a default, and which one users have to specify explicitly.

raxod502:
Sorry for the delay. I have sampled the repositories I have in my `~/.emacs.d/straight/repos` and found that 454 have an `origin/HEAD` while 109 do not. Here is the list:

<details>
<p>

```
[BROKEN] 0x0
[WORKING] ace-window
[WORKING] ada-mode
[WORKING] a.el
[WORKING] alert
[BROKEN] all-the-icons.el
[WORKING] amx
[WORKING] anaconda-mode
[BROKEN] apache-mode
[WORKING] apel
[BROKEN] apheleia
[BROKEN] apples-mode
[WORKING] at-el
[BROKEN] atomic-chrome
[BROKEN] auctex
[WORKING] auctex-src
[WORKING] autodisass-llvm-bitcode
[WORKING] avy
[WORKING] bbdb
[WORKING] biblio.el
[WORKING] blacken
[BROKEN] blackout
[BROKEN] buffer-move
[WORKING] bui.el
[WORKING] cache
[WORKING] caliper
[WORKING] caps-lock
[BROKEN] cider
[WORKING] cl-generic
[BROKEN] clj-refactor.el
[WORKING] cl-lib
[BROKEN] clojure-mode
[WORKING] closql
[WORKING] col-highlight
[WORKING] company-auctex
[WORKING] company-flx
[WORKING] company-irony
[WORKING] company-irony-c-headers
[WORKING] company-lsp
[BROKEN] company-mode
[WORKING] company-tern
[BROKEN] contributor-guide
[BROKEN] counsel-projectile
[BROKEN] crontab-mode
[WORKING] crosshairs
[WORKING] crypt-plus-plus
[WORKING] csharp-mode
[WORKING] csound-mode
[WORKING] csound-x
[WORKING] csv-mode
[BROKEN] ctrlf
[WORKING] dap-mode
[WORKING] dart-mode
[BROKEN] dash.el
[WORKING] di
[BROKEN] diary-manager
[WORKING] dict-tree
[WORKING] diff-hl
[WORKING] dired-hacks
[WORKING] dockerfile-mode
[WORKING] docker-tramp.el
[WORKING] dumb-jump
[BROKEN] dumbparens
[WORKING] edit-indirect
[WORKING] edn.el
[WORKING] efs
[WORKING] egg
[WORKING] eglot
[WORKING] eldoc
[WORKING] elfeed-org
[WORKING] el-get
[BROKEN] elisp-refs
[WORKING] elisp-ruby-electric
[WORKING] elm-mode
[WORKING] elpa
[BROKEN] el-patch
[WORKING] elpy
[WORKING] emacs-anzu
[WORKING] emacs-async
[WORKING] emacs-deferred
[WORKING] Emacs-D-Mode
[WORKING] emacs-doom-themes
[WORKING] emacs-git-gutter
[WORKING] emacs-git-gutter-fringe
[BROKEN] emacs-hcl-mode
[WORKING] emacs-libvterm
[BROKEN] emacs-memoize
[WORKING] emacs-mini-frame
[WORKING] emacsmirror-mirror
[WORKING] emacs-monokai-pro-theme
[WORKING] emacs-multi
[WORKING] emacsql
[BROKEN] emacs-racer
[BROKEN] emacs-request
[WORKING] emacs-sbt-mode
[WORKING] emacs-scala-mode
[WORKING] emacs-terraform-mode
[WORKING] emacs-web-server
[BROKEN] emacs-websocket
[WORKING] Emacs-wgrep
[WORKING] emacs-which-key
[BROKEN] emtas
[WORKING] epkgs
[BROKEN] epl
[WORKING] ESS
[BROKEN] esup
[WORKING] esxml
[BROKEN] etunes
[WORKING] eval-sexp-fu.el
[WORKING] evil
[WORKING] evil-args
[WORKING] evil-easymotion
[WORKING] evil-ediff
[WORKING] evil-escape
[WORKING] evil-exchange
[WORKING] evil-iedit-state
[WORKING] evil-indent-plus
[WORKING] evil-lion
[WORKING] evil-lisp-state
[WORKING] evil-magit
[WORKING] evil-matchit
[WORKING] evil-mc
[WORKING] evil-nerd-commenter
[WORKING] evil-numbers
[WORKING] evil-org-mode
[WORKING] evil-quickscope
[WORKING] evil-search-highlight-persist
[WORKING] evil-surround
[WORKING] evil-tutor
[WORKING] evil-visual-mark-mode
[WORKING] evil-visualstar
[BROKEN] example-repository
[WORKING] excorporate
[WORKING] exec-path-from-shell
[WORKING] expand-region.el
[WORKING] exwm
[WORKING] eyebrowse
[WORKING] f3
[WORKING] fancy-battery.el
[BROKEN] f.el
[WORKING] files-plus
[WORKING] filladapt
[WORKING] find-file-in-project
[WORKING] find-test
[WORKING] flim
[WORKING] floobits-emacs
[WORKING] flx
[WORKING] flx-isearch
[BROKEN] flycheck
[WORKING] flycheck-bashate
[WORKING] flycheck-haskell
[WORKING] flycheck-irony
[WORKING] flycheck-pos-tip
[WORKING] flymake
[WORKING] flyspell-correct
[WORKING] font-lock-plus
[WORKING] forge
[WORKING] frecency.el
[BROKEN] fringe-helper.el
[WORKING] fsm
[WORKING] fuzzy-el
[WORKING] gas-mode
[BROKEN] geiser
[WORKING] general.el
[WORKING] ggtags
[WORKING] gh.el
[WORKING] gh-md.el
[BROKEN] ghub
[WORKING] ghub-plus
[WORKING] git.el
[WORKING] git-gutter
[WORKING] git-gutter-fringe
[WORKING] git-gutter-fringe-plus
[WORKING] git-gutter-plus
[BROKEN] git-link
[BROKEN] git-modes
[WORKING] git-timemachine
[WORKING] gntp.el
[BROKEN] gnu-elpa-mirror
[WORKING] gnuplot-mode
[WORKING] golden-ratio.el
[BROKEN] go-mode.el
[WORKING] google-calendar
[WORKING] google-maps.el
[WORKING] google-translate
[WORKING] goto-chg
[WORKING] graphql.el
[WORKING] graphviz-dot-mode
[WORKING] grapnel
[WORKING] haml-mode
[WORKING] happy-mode
[BROKEN] haskell-mode
[WORKING] heap
[WORKING] helm
[WORKING] helm-bibtex
[WORKING] helm-company
[WORKING] helm-css-scss
[WORKING] helm-c-yasnippet
[WORKING] helm-dash
[WORKING] helm-descbinds
[WORKING] helm-dictionary
[WORKING] helm-flx
[WORKING] helm-gitignore
[WORKING] helm-itunes
[WORKING] helm-make
[WORKING] helm-swoop
[BROKEN] help-fns+
[WORKING] help-fns-plus
[BROKEN] helpful
[WORKING] hide-comnt
[BROKEN] highlight
[WORKING] highlight.el
[WORKING] Highlight-Indentation-for-Emacs
[WORKING] highlight-numbers
[WORKING] highlight-parentheses.el
[WORKING] highlight-thing.el
[BROKEN] hindent
[WORKING] historian.el
[WORKING] history
[WORKING] hjson-mode
[WORKING] hl-line-plus
[WORKING] hl-todo
[WORKING] homebrew-mode
[BROKEN] ht.el
[WORKING] hugo.el
[WORKING] hungry-delete
[BROKEN] hydra
[WORKING] hy-mode
[WORKING] hyperbole
[WORKING] ibuffer-projectile
[WORKING] ical-event
[WORKING] icicles
[WORKING] icomplete-vertical
[WORKING] idle-highlight-in-visible-buffers-mode
[WORKING] iedit
[WORKING] iflipb
[WORKING] igrep
[WORKING] imenu-list
[WORKING] imp.el
[WORKING] indent-guide
[WORKING] info-plus
[BROKEN] inf-ruby
[WORKING] ini-mode
[WORKING] insert-shebang
[BROKEN] irony-eldoc
[WORKING] irony-mode
[BROKEN] ishikk
[WORKING] itail
[WORKING] ivy-explorer
[WORKING] ivy-purpose
[WORKING] ivy-rich
[WORKING] ivy-xref
[WORKING] ivy-youtube
[WORKING] ix.el
[WORKING] js2-mode
[WORKING] js2-refactor.el
[WORKING] js3-mode
[WORKING] js-doc
[BROKEN] json-mode
[BROKEN] json-reformat
[WORKING] jsonrpc
[BROKEN] json-snatcher
[WORKING] julia-emacs
[WORKING] jump.el
[WORKING] key-chord
[WORKING] kotlin-mode
[WORKING] langdoc
[WORKING] lein
[WORKING] less-css-mode
[WORKING] less-mode
[BROKEN] let-alist
[WORKING] libegit2
[WORKING] link-hint.el
[WORKING] linum-relative
[WORKING] lispy
[WORKING] live-py-plugin
[WORKING] livescript-mode
[WORKING] livid-mode
[WORKING] llvm
[WORKING] log4e
[WORKING] logito
[WORKING] lookup-word-on-internet
[BROKEN] loop.el
[WORKING] ls-lisp-plus
[WORKING] lsp-dart
[WORKING] lsp-elixir.el
[WORKING] lsp-haskell
[WORKING] lsp-ivy
[WORKING] lsp-java
[WORKING] lsp-julia
[WORKING] lsp-latex
[WORKING] lsp-mode
[WORKING] lsp-php
[BROKEN] lsp-python-ms
[WORKING] lsp-scala
[WORKING] lsp-treemacs
[WORKING] lsp-ui
[WORKING] lua-mode
[BROKEN] macrostep
[WORKING] magit
[WORKING] magit-gh-pulls
[WORKING] magit-gitflow
[WORKING] magithub
[WORKING] magit-popup
[BROKEN] markdown-mode
[WORKING] markdown-toc
[WORKING] marshal.el
[WORKING] math-symbol-lists
[WORKING] matlab-mode
[WORKING] maven-mode
[WORKING] mbsync-el
[WORKING] mc
[WORKING] mediawiki-el
[BROKEN] melpa
[WORKING] mercury
[BROKEN] mmm-mode
[WORKING] mmt
[WORKING] mode-compile
[WORKING] monokai-emacs
[WORKING] move-text
[WORKING] mpg123
[WORKING] mu
[WORKING] mu4e
[BROKEN] multiple-cursors.el
[WORKING] multi-term
[WORKING] nasm-mode
[WORKING] navi2ch
[WORKING] nethack
[WORKING] nginx-mode
[WORKING] nhexl-mode
[WORKING] node-resolver.el
[BROKEN] no-littering
[WORKING] notmuch
[WORKING] ntlm
[WORKING] nusmv
[WORKING] ob-ipython
[WORKING] om.el
[WORKING] open-junk-file
[WORKING] openscad
[WORKING] orderless
[BROKEN] org
[WORKING] org-alert
[WORKING] org-brain
[WORKING] org-bullets
[WORKING] org-caldav
[WORKING] org-download
[WORKING] org-journal
[WORKING] org-mime
[WORKING] org-password-manager
[WORKING] org-pdfview
[WORKING] org-pomodoro
[WORKING] org-present
[WORKING] org-projectile
[WORKING] org-reveal
[WORKING] org-roam
[WORKING] org-tree-slide
[WORKING] org-trello
[WORKING] org-wild-notifier.el
[WORKING] osx-plist
[WORKING] osx-trash
[BROKEN] osx-trash.el
[WORKING] otp
[WORKING] outorg
[WORKING] outshine
[WORKING] ov
[WORKING] ox-gfm
[WORKING] ox-hugo
[WORKING] ox-pandoc
[WORKING] p4.el
[WORKING] package-build
[BROKEN] package-lint
[WORKING] packed
[WORKING] pacmacs.el
[WORKING] pandoc-mode
[WORKING] paradox
[WORKING] paredit
[WORKING] parent-mode
[WORKING] parsebib
[WORKING] parseclj
[WORKING] parseedn
[WORKING] password-store
[WORKING] patchutils
[WORKING] pbl
[WORKING] pcache
[WORKING] pcap-mode
[WORKING] pcmpl-git-el
[WORKING] pcre2el
[WORKING] pdf-tools
[WORKING] pdftools
[WORKING] peg.el
[WORKING] persp-mode.el
[WORKING] pfuture
[WORKING] PG
[WORKING] php-mode
[WORKING] pipenv.el
[WORKING] pip-requirements.el
[WORKING] pkgbuild-mode
[WORKING] pkg-info
[BROKEN] pkg-info.el
[WORKING] PlatformIO-Mode
[WORKING] polymode
[BROKEN] popup-el
[WORKING] popwin-el
[WORKING] posframe
[BROKEN] pos-tip
[WORKING] powerline
[WORKING] predd
[BROKEN] prescient.el
[WORKING] prettier-emacs
[WORKING] private-diary
[WORKING] profile-dotemacs
[WORKING] project
[WORKING] project-emacs--folding-mode
[BROKEN] projectile
[WORKING] protobuf
[WORKING] puppet-mode
[WORKING] pyenv-mode
[WORKING] py-isort.el
[BROKEN] pyrelight
[WORKING] pytest-el
[WORKING] python
[WORKING] python-docstring-mode
[WORKING] pythonic
[WORKING] python-mode
[WORKING] python-x.el
[BROKEN] pyvenv
[WORKING] quack
[WORKING] quelpa
[BROKEN] queue
[WORKING] rainbow-delimiters
[WORKING] rainbow-mode
[WORKING] raven
[WORKING] real-auto-save
[WORKING] recursive-narrow
[WORKING] reddit-mode
[WORKING] reformatter.el
[WORKING] refs.el
[WORKING] resize-window
[BROKEN] restart-emacs
[BROKEN] rg.el
[BROKEN] robe
[WORKING] robots-txt-mode
[WORKING] ruby-electric
[WORKING] ruby-electric.el
[BROKEN] rust-mode
[WORKING] sallet
[WORKING] sass-mode
[WORKING] sbt
[WORKING] scss-mode
[BROKEN] s.el
[BROKEN] selectrum
[WORKING] semantic-refactor
[WORKING] semi
[BROKEN] seq
[WORKING] seq.el
[WORKING] sergv-dotemacs
[BROKEN] sesman
[WORKING] shell-pop-el
[WORKING] showkey
[WORKING] shpec-mode
[BROKEN] shut-up
[WORKING] simple-call-tree
[WORKING] skewer-mode
[WORKING] sly
[BROKEN] smartparens
[WORKING] smex
[WORKING] sml-mode
[WORKING] smtpmail-multi
[WORKING] snails
[WORKING] soap-client
[WORKING] solarized-emacs
[WORKING] spaceline
[WORKING] spacemacs-theme
[BROKEN] spinner
[WORKING] split-root
[WORKING] sprunge.el
[WORKING] ssass-mode
[BROKEN] ssh-config-mode-el
[BROKEN] stfu-emacs
[BROKEN] straight.el
[WORKING] string-inflection
[WORKING] styleguide
[WORKING] sudo-edit
[WORKING] sudoku.el
[WORKING] su.el
[WORKING] sunrise-commander
[BROKEN] swift-mode
[WORKING] swiper
[BROKEN] switch-buffer-functions-el
[BROKEN] sx.el
[WORKING] symon
[WORKING] system-packages
[BROKEN] tabcrush
[WORKING] tablist
[WORKING] tagedit
[WORKING] telephone-line
[WORKING] tern
[WORKING] terraform-mode
[WORKING] tide
[WORKING] tiny
[WORKING] tldr.el
[WORKING] tNFA
[WORKING] toc-org
[BROKEN] toml-mode.el
[WORKING] tptp-el
[WORKING] transient
[WORKING] transpose-frame
[WORKING] treemacs
[WORKING] treepy.el
[WORKING] trie
[WORKING] tuareg
[WORKING] turing-machine
[BROKEN] typescript.el
[WORKING] typit
[BROKEN] undo-tree
[WORKING] uniquify-files
[WORKING] url-http-ntlm
[BROKEN] use-package
[BROKEN] utunes
[WORKING] uuidgen-el
[BROKEN] verilog-mode
[BROKEN] vimrc-mode
[WORKING] visual-basic-mode
[WORKING] visual-fill-column
[BROKEN] visual-regexp.el
[BROKEN] visual-regexp-steroids.el
[WORKING] vi-tilde-fringe
[WORKING] vline
[WORKING] vm
[WORKING] volatile-highlights.el
[WORKING] vr
[WORKING] vue-html-mode
[WORKING] vue-mode
[WORKING] w3
[WORKING] wanderlust
[WORKING] wc
[WORKING] web-beautify
[WORKING] web-completion-data
[BROKEN] web-mode
[BROKEN] webpaste.el
[WORKING] wisi
[BROKEN] with-editor
[BROKEN] with-feature
[BROKEN] word-count
[WORKING] word-counter
[WORKING] word-count-mode
[WORKING] worf
[WORKING] ws-butler
[WORKING] xcode-mode
[WORKING] xelb
[WORKING] xref
[WORKING] xterm-color
[BROKEN] yaml-mode
[WORKING] yapfify
[BROKEN] yasnippet
[WORKING] z3-mode
[WORKING] zeal-at-point
[WORKING] zenburn-emacs
[BROKEN] zerodark-theme
[WORKING] zone-matrix
[WORKING] zone-nyan
[WORKING] zoom
[WORKING] zoutline
```

</p>
</details>

Looking at the list, the broken ones seem to be all of the ones I actually use. Maybe it has something to do with shallow clones? I did recently change that in my configuration. If you can't get your hands on a repository that has this property, I can provide you with one to test, or try to bisect further.

> Another alternative is finding the remote branch, and using the correspondingly named local branch when `:branch` is not set.

This is indeed what I was suggesting. It's what would make the most sense, to me, in terms of compatibility. Currently, we have no concept of a local repository without a preferred branch. Even if you don't specify a branch, then it will use `master` by default, and enforce this in the repository management commands. The main objective of this change, in my view, is to make it so that we're enforcing the _correct_ preferred branch in the case of (primarily MELPA) recipes that expect the semantics of "use the branch that is marked as default on the remote".

> the checked out branch is the one that will actually affect emacs when starting, so it seems like the one that people will want to manipulate

I would say rather that if you're interested in using a different branch, you should update the recipe in your init-file. You already can use whatever branch you want for random testing, since `straight.el` doesn't validate anything on startup, but in my view it's helpful as a default to normalize to a single preferred branch when the user asks for local changes to be reverted interactively. Otherwise, it's very easy to forget that you're on some random feature branch in one of your repositories.

ahyatt:
Interesting, I haven't been able to reproduce not having that `HEAD` remote branch visible.  I've tried cloning with `git clone --depth 10 https://github.com/company-mode/company-mode.git --single-branch`, but I still see the `HEAD` with `git branch -r`.

I think I'll need to figure out how these are created, because without that, it's hard to say what to do when we can't find the `HEAD`.  Perhaps these repositories are special in some way that allows us to understand what the default remote directory is.  If not, I'm not sure what we can do without relying on guesses about what the default branch might be.

raxod502:
Once we understand how these repositories are being created, if they do not have enough info to get the remote branch, I think it would be fine to make it so that when `straight.el` goes to compute the default branch it will ask if it can automatically fix them (by fetching, for example).

Check this out, I think it may be the reason: https://stackoverflow.com/a/17639471/3538165

I took a look at that list above, this time sorted by mtime, and I think I understand what's going on. The broken repositories are the ones that I have authored (so local `git init` and then push, never cloned) and the ones that I've had around for so long that they predate the feature of `straight.el` to normalize remote protocols (so the original HTTPS clone remote was deleted and a new SSH remote was added).

ahyatt:
Thanks for that last post, which was very helpful.

Hopefully the new fix should work fine on all your repositories - except any that have no remotes at all, in which case failure is expected.  There was no need prompt the user.  

I verified the fix by deleting the HEAD link in the locally-stored remote repository refs, and we are still able to get the default branch name.

straight.el
```diff
@@ -2057,12 +2054,19 @@ used; it should be a string that is not prefixed with a remote
...
       (local-repo branch)
-    (while t
-      (and (straight-vc-git--ensure-local recipe)
-           (or (straight-vc-git--ensure-head
-                local-repo branch (format "%s/%s" remote remote-branch))
-               (straight-register-repo-modification local-repo))
-           (cl-return-from straight-vc-git--merge-from-remote-raw t)))))
+    (let ((branch (or branch (straight-vc-git--default-branch-in-local-repo
+                              local-repo)))
+          (remote-branch (or remote-branch
```
raxod502, line 2059: Why does this work? It seems like this binding of `remote-branch` depends on `remote-branch` already being bound.

raxod502 (commented):
So what happens when you run `M-x straight-normalize-package` for a package that only has a local repository, and no remote configured? Which branch gets checked out?

straight.el
ahyatt: `remote-branch` is passed into the function.  So if it is `nil` (in other words, unspecified), we'll give it the value of the default branch.  So in the body of the `let` statement, we should have a valid `remote-branch` variable.

ahyatt:
Thanks for calling my attention to normalize.  If there are no remotes, I believe that `(straight-vc-git--ensure-remotes recipe)` in `straight-vc-git--ensure-local` will return false.  I tested this out, and indeed normalize did nothing in this case.  I didn't add the call to `straight-vc-git--ensure-remotes` so I believe this is what happened before as well.

When I looked into this, though, and started testing normalize more, I found a few issues which I have now fixed.  

However, I noticed that there's an issue when you have a local repository that has the a single branch locally, which is no longer the default remote branch.  In that case, normalize detects an issue (but not the fundamental issue of the missing branch), but cannot fix it.  I think in such cases we should fetch the new default branch and check it out.  WDYT?

raxod502:
I think the CI failure for master will be fixed if you rebase.

ahyatt:
I fixed `straight-vc-git--ensure-head` so that it will just checkout the default remote branch if a remote ref has not been passed in.  I think that should be good, however, there's a lot of complicated logic afterwards that we'll be skipping in this case.  I'm not so sure if that's OK or not, however, based on what we discussed before, it seems like what you want to see.

I also rebased, hence the extreme amount of commits you see.

straight.el
```diff
@@ -1954,6 +1951,11 @@ confirmation, so this function should only be run after
             local-repo branch)
            ("a" (format "Attach HEAD to branch %S" branch)
             (straight--get-call "git" "checkout" branch)))))
+      ((and (null ref)
+            ;; Are we not on the actual default remote branch? We should be
+            ;; switch to it. If it is not existing, "git checkout" will
+            ;; automatically create it.
+            (straight--get-call "git" "checkout" branch)))
```
raxod502, line 1958: Hm. It seems like the addition of this clause causes the code farther below to never run, as you mentioned. Why are we adding this again? Is it because the current code doesn't properly handle when the remote branch doesn't exist locally, or is it something else?

ahyatt: It should run if there's a ref, which can happen in some flows.  I guess what I don't understand is why the following code is there.  Is you are on a branch when the package is normalized, and it isn't the default branch, why not just check out the default branch?  I think that's the behavior we actually want, right?  At least for normalization - the same code is run for merging, and I think there the remote repository is passed in as a ref, so it should behave differently, and run the code below.

raxod502:
I am responding above. I just noticed, however, that this PR is based on the `master` branch, which explains why rebasing (actually merging would have been fine also) did not fix the build. Latest development takes place on the `develop` branch of `straight.el`.

ahyatt:
Thanks for the info.  I forced a rollback of the rebase, sorry for the messiness.  I'll re-rebase to `develop` soon.  In the meantime, please help me understand what you want from the code by responding to my response to your code review comment.

progfolio:
@ahyatt: Just to confirm what was mentioned above, the CI failure is:

> straight.el:4313:12: Warning: `update-directory-autoloads' is an obsolete
    function (as of 28.1); use `make-directory-autoloads' instead.

Which was addressed a couple of weeks ago on the develop branch:

https://github.com/raxod502/straight.el/commit/48fae3822c00f8c48aac957474566101c1c7b6d4

Rebasing will fix that issue.

Unfortunately, I can't comment on your other question as I'm not that familiar with this section of the codebase yet.

straight.el
raxod502: Sorry for the delay, I got interrupted between writing my comment below and writing my comment here :o 

So the reason we have the code below is that it allows the user to interactively decide in what manner they want to get the default branch checked out. When I wrote that code, I was thinking of a situation like: you're working on a feature branch, and you do a normalize operation. You might want to not just check out master, but rather merge your feature branch into master. The idea of the VC popup below is to let the user decide whether they want to do that or whether they want to do something else, like just checking out master directly.

Now, we might decide that the interactive deciding is overkill because there isn't a real use case for it, but that's why I added the code.

I guess what I would expect to see in this PR is that it would re-use the logic below, but in the case of a default remote branch, it would use that branch as the branch to ask the user how to check out. Basically, the added feature is that the name of the branch to be used is fetched from the remote instead of being known locally, but the actual VC logic is the same.

Does that make sense?

ahyatt: That makes sense, but also doesn't work.  Let's take the example of the package "org-roam".  My local branch is `master`, but the default branch is `develop`.  `develop` is an ancestor of `master`, so all it could ask me to do is to reset to head which does nothing.  I think the correct thing to do is to consider `ref-ahead-p` and `ref-behind-p` only to be true if we're on the right branch.  Otherwise, I don't think you can think of things as "ahead" or "behind".  Maybe it might make sense, but not with the current options at least.  So I modified the code so that is what happens - if the branch is different, neither of those variables will be true, so the user will get a reasonable set of options, which works.

This should be closer to what you'd like to see, I hope.

raxod502: Yes, you are right: if the branches have diverged, then neither of those variables should be true if the branches have diverged. (`ref-ahead-p` means "descendant" and `ref-behind-p` means "ancestor", in Git's terminology.) We would expect the user to get the prompt saying "your branch has diverged". Let me take another look.

straight.el
```diff
@@ -1977,23 +1981,26 @@ confirmation, so this function should only be run after
...
+                 (format "HEAD%s is behind default branch %S"
+                         on-branch branch))
+                (t (format "HEAD%s has diverged from default branch %S"
                            on-branch branch))))))
           ;; Here be dragons! Watch the quoting very carefully in
           ;; order to get the lexical scoping to work right, and don't
           ;; confuse this syntax with the syntax of the
           ;; `straight--popup' macro.
-          `(,@(when ref-ahead-p
+          `(,@(when (and ref-ahead-p)
```
raxod502, line 2003: I think this `and` was left over from something else :)

straight.el
raxod502: Okay, so looking at that code snippet -- it's a bit tricky for me to understand exactly what the original intent was, but -- I'm thinking the right fix is to replace `branch` with `cur-branch` in the original conditional. I'm not sure why we would care whether the recipe's branch is ahead or behind -- we are trying to fix up the local repository, so presumably we would care whether the _actual_ checked-out branch is ahead or behind.

I think making that change, instead of the one in the PR currently, would also restore reasonable behavior.

What do you think of that idea? Is there anything I've missed?

straight.el
ahyatt: Fixed, kind of.  Maybe I'm wrong, but I'm not sure this clause ever makes sense.  When would the user want to reset their state to a ref?  But in any case, it especially doesn't make sense without a ref, so I added that to the `and` clause.

straight.el
ahyatt: It does make sense, and it does work - although I noticed another issue where I would get two "f" options if I did things that way.  I fixed this by making one of them a "c" option to checkout, and changed the user visible error message, as well as restricting the other f option.  See the comment on the code below.

raxod502: > I would get two "f" options

Well, I can't say I'm surprised. It's certainly not a problem with your change, those popups are known to not always do a very good job at producing the correct set of options :'(

Thanks for fixing it!

straight.el
raxod502: So I believe this would be the case where you're on `master` and then you pull additional commits from `origin/master`. So then if you're executing `straight-pull-package`, you'd end up calling `straight-vc-git--ensure-head` with a ref of `origin/master`, and the current branch would be `master`. Since `origin/master` is ahead of `master`, but `master` isn't ahead of `origin/master`, we'd have `ref-ahead-p` non-nil and `ref-behind-p` nil. Presuming that `straight-vc-git-auto-fast-forward` is nil, we'd end up in this case. Now semantically you'd expect us to be running `git merge` or something in order to fast-forward `master` to `origin/master`, but implementation-wise it's simpler to do it with `git reset --hard`.

That said -- I think you've observed by now that a lot of the VC code is slightly suspect. If you want to go to town on replacing the existing logic with something that produces more reasonable-seeming popups/results when you call the interactive commands, be my guest! You could probably do a better job than the original version in many cases.

ahyatt:
Right now I think this should be good, or at least good as far as I've tested.  I haven't been able to successfully push a rebase of this branch on top of `develop` yet.  However, I'm not sure that's necessary - you have a pull request which you can apply to `develop`, which should lead to clean CIs, if not I'm happy to supply whatever fixes.  Let me know if you think something else is missing.  Based on the number of problems cropping up, I think it's more than time we get this in.

progfolio:
We still need to rebase this on top of develop. I'll see if I can help you out with that.

straight.el
```diff
@@ -2081,13 +2091,15 @@ Return non-nil. If no local repository, do nothing and return non-nil."
...
         (cl-return t))
-      (let ((push-error-message nil))
+      (let ((push-error-message nil)
+            (branch
+             (or branch
+                 (straight-vc-git--default-remote-branch remote local-repo))))
         (while t
           (while (not (straight-vc-git--ensure-local recipe)))
-          (let ((ref (format "%s/%s" remote branch)))
+          (let* ((ref (format "%s/%s" remote branch)))
```
progfolio, line 2100: Is there any reason for this to be a let*?

progfolio (changes requested):

straight.el
ahyatt: There may have been when I initially wrote it, but now it must be superfluous.  I'll revert it.

progfolio:
@ahyatt if everything looks good on #630, you can rebase this branch on that and make any other changes you need to.

progfolio:
Merged w #630

PR #546, opened 2020-07-16
straight--with-plist: use cl-destructuring-bind

progfolio:
`cl-destructuring-bind` can destructure a plist given a list of symbols.
Leveraging this simplifies the definition of `staright--with-plist`.


raxod502:
Thanks!

PR #547, opened 2020-07-17
Feature/block on native comp

blak3mill3r:
This is a useful option to have, I feel.

I want my `emacs --daemon` to start and block on native-compilation finishing, so that I can then CRIU-checkpoint the process with https://gitlab.com/blak3mill3r/emacs-ludicrous-speed

This change reintroduces the `straight--wait-for-async-jobs` function and disables it by default.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -4338,12 +4338,10 @@ individual package recipe."
                                  straight-disable-native-compilation))))
 
 (defun straight--native-compile-package (recipe)
-  "Queue native compilation for the symlinked package specified by RECIPE.
+  "Native-compile files for the symlinked package specified by RECIPE.
 RECIPE should be a straight.el-style plist. Note that this
 function only modifies the build folder, not the original
-repository. Also note that native compilation occurs
-asynchronously, and will continue in the background after
-`straight-use-package' returns."
+repository."
   (when (and (fboundp 'native-compile-async)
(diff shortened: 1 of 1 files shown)


blak3mill3r:
This code as it is does not appear to work. Even with `straight--wait-for-async-jobs` set to `t` it still seems to happen asynchronously.

straight.el
```diff
@@ -4338,12 +4338,10 @@ individual package recipe."
                                  straight-disable-native-compilation))))
 
 (defun straight--native-compile-package (recipe)
-  "Queue native compilation for the symlinked package specified by RECIPE.
+  "Native-compile files for the symlinked package specified by RECIPE.
 RECIPE should be a straight.el-style plist. Note that this
 function only modifies the build folder, not the original
-repository. Also note that native compilation occurs
-asynchronously, and will continue in the background after
-`straight-use-package' returns."
```
raxod502, line 4346: This docstring needs to explain what happens depending on the value of `straight--wait-for-async-jobs`, which should itself be a user option.

raxod502:
This thread is being closed automatically by [Tidier](https://github.com/raxod502/tidier) because it is labeled with "waiting on response" and has not seen any activity for 90 days. But don't worry—if you have any information that might advance the discussion, leave a comment and I will be happy to reopen the thread :)

Issue #115, opened 2017-07-25
Pull recipes from el-get


raxod502:
I noticed they've got some shiny things that I can totally steal, like instructions on [how to build Org](https://github.com/dimitri/el-get/blob/4a1dde57e8a4d4c22ed279bac8253f5f16da9197/recipes/org-mode.rcp#L7-L14).

agsdot:
:+1

agsdot:
Thumbs-up

raxod502:
Note that this probably doesn't make sense until after #71 and #72, and it also won't be making it into the 1.0 release unless somebody provides a pull request.

Issue #72, opened 2017-07-09
Support arbitrary build commands


raxod502:
Some packages (like Org) require you to `make` them before they can be run correctly. Others have nonstandard types of documentation, and so on.

vyp:
Sorry! It was in agreement with this issue that such packages are silly, than with disagreement or anything. Anyway removed to remove confusion, haha.

sooheon:
So to be clear one cannot currently have straight.el fully manage org?

raxod502:
@sooheon No, it works just fine. You'll just get some warnings about the Org version not being defined, unless you add [some hacks](https://github.com/raxod502/radian/blob/ee92ea6cb0473bf7d20c6d381753011312ef4a52/radian-emacs/radian-org.el#L46-L112).

dieggsy:
Can confirm it works just fine. I don't think I get any version warnings (?) but that hack is cool anyway.

rgrinberg:
One package that doesn't work at all without a build step is proof-general. It's quite an important package as well so it would be nice to have support.

dieggsy:
@rgrinberg the same may be true of pdf-tools

vyp:
also structured-haskell-mode

raxod502:
Indeed. You can probably get a pretty good idea of the list of such packages by just searching [all the el-get recipes](https://github.com/dimitri/el-get/tree/master/recipes), since those specify build commands.

Are those packages not in MELPA? Because as I understand it, MELPA does not support build commands.

In any case, I am quite busy at the moment, but would gladly merge any pull requests that came along.

dieggsy:
@raxod502 this is pretty late but I think MELPA does seem to support build commands, see here: 

https://github.com/melpa/melpa/blob/269acf093c3aa508479bf532a4e07c9c6d568c72/recipes/pdf-tools#L1

I believe pdf-tools gets compiled at `package-install` time.

raxod502:
@dieggsy MELPA doesn't build pdf-tools. What you see in the recipe is just a specification to create a `build/` directory in the built package and to link the named files into it:

```
drwxr-xr-x     - raxod502  2 Dec  7:29  .
drwxr-xr-x     - raxod502  2 Dec  7:29  ├── build
lrwxr-xr-x    58 raxod502  2 Dec  7:29  │  ├── Makefile -> /Users/raxod502/.emacs.d/straight/repos/pdf-tools/Makefile
[...]

Issue #161, opened 2017-10-10
straight-freeze-versions tries to clone MELPA


diamond-lizard:
With the only straight-related code in my emacs init file being the bootstrap code from [this comment](https://github.com/raxod502/straight.el/issues/159#issuecomment-334600171) and the only repo in `~/.emacs/straight/repos` being `straight.el` itself, when I run `M-x straight-freeze-versions` what I expect to happen is for `straight.el` to simply create one single lockfile for itself.  Instead what happens is it started to clone MELPA (which attempt I immediately aborted).

I don't even use MELPA, and I don't understand why `straight.el` would try to clone it without an explicit request from me (or at least permission to do so), especially as it seems completely unrelated to the simple act of creating a single lockfile for `straight.el` itself.

raxod502:
Ah! Good catch. The reason this is happening is that recipe repositories are treated, internally, just like packages. The default recipe repositories are registered in `bootstrap.el`:

```
(straight-use-recipes '(melpa :type git :host github
                              :repo "melpa/melpa"
                              :no-build t))

(straight-use-recipes '(emacsmirror :type git :host github
                                    :repo "emacsmirror/epkgs"
                                    :nonrecursive t
                                    :no-build t))
```

and as you can see, the definition of `straight-use-recipes` just delegates to a `straight-use-package` call:

```
(defun straight-use-recipes (melpa-style-recipe)
  "Register a recipe repository using MELPA-STYLE-RECIPE.
This registers the recipe and builds it if it is already cloned.
Note that you probably want the recipe for a recipe repository to
include a non-nil `:no-build' property, to unconditionally
inhibit the build phase.

This function also adds the recipe repository to
`straight-recipe-repositories', at the end of the list."
  (straight-use-package-lazy melpa-style-recipe)
  (add-to-list 'straight-recipe-repositories
               (if (listp melpa-style-recipe)
                   (car melpa-style-recipe)
                 melpa-style-recipe)
               'append))
```

[...]

Issue #149, opened 2017-08-30
Allow installing packages directly (not using Git)


raxod502:
See https://github.com/raxod502/straight.el/issues/148#issuecomment-325849696 and https://github.com/raxod502/straight.el/issues/148#issuecomment-325875032.

raxod502:
My plan:

1. Factor out the `:host` handling into an extensible system which can have hosts added dynamically.
2. For each host, add a function that generates a tarball download URL for any given revision, or returns nil if direct download is not supported. We could even support things like CGit, I think.
3. Controllable by a user option and/or recipe keyword, download a tarball instead of a Git repo the first time, and write a file `.straight-head` or something which contains the SHA of the tarball revision.
4. Repository management operations, instead of doing stuff with Git, just download a new tarball when you need to do a checkout. Alternatively you can ask for them to be converted into full Git repositories.
5. Using `find-file-hook` or similar, files inside tarball repositories are marked as read-only, and if you try to edit them then `straight.el` first asks to convert them to full Git repositories (in-place).

The main problem with this approach is that I'm not sure how to get the revision SHA when you're cloning master (i.e. initially). Any ideas?

raxod502:
cc @vyp @JAremko @et2010 @stonejiajia

raxod502:
Amazing! Thanks.

Requiring two network calls is of course not ideal but I think it would still be an improvement, especially when cloning repositories in parallel.

yantar92:
This feature might not only be useful to reduce disk space usage. I just encountered a situation when Emacswiki package (dired+) in https://github.com/emacsmirror/ is quite outdated in comparison with the wiki version. This problem is likely unavoidable because packages in emacsmirror are updated manually. I would prefer if straight.el provided an option to download Emacswiki packages directly from Emacswiki to ensure the latest version.

raxod502:
Sorry, but installing packages from Emacswiki will never be supported. See https://github.com/melpa/melpa/pull/5008 for the reasoning. This issue is specifically for installing snapshots of packages that are hosted on GitHub and similar registries.

yantar92:
> Sorry, but installing packages from Emacswiki will never be supported. 

To be frank, I thought that straight.el is possibly the safest way to install packages from such untrusted sources. Unlike direct download, I can easily view diff of the package when updating. Hence, I can directly review any suspicious changes in the package.

raxod502:
[...]

PR #549, opened 2020-07-27
Add el-get recipe support

progfolio:
Builds off #535 to add support for el-get recipes.
Closes: #115 
Related: #72

# Caveats:

- only lists/retrieves recipes of `:type` `git` and `github`

This accounts for roughly 76% of el-get's packages. (1373 out of 1793)

- does not implement el-get's `:prepare` or `:post-init` behavior.

We breifly discussed this in #535, but I have some concerns about implementing it.
I'm not sure how to do this in a way that doesn't stomp `use-package` declarations and I think that sort of behavior is best left to `use-package` or similar solutions.
One idea I had was to simple report whether or not a package has a `:prepare` and/or `:post-init` value when installing the package. This way the user can add those commands to their init via their preferred method if they so choose.

- dependency management

I see that `straight--compute-dependencies` handles pkg.el files and the `Package-Requires:` header keyword of .el files.
el-get also has the notion of an explicit `:depends` keyword that maps to an alleged list (I have seen some that do not match their stated dependencies in the `Packages-Requires` header) of the package's dependencies.
Not sure if it would be worth it or not to use this information.

Still a work in progress, but I figured we could get a conversation started on the implementation and try some ideas out.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -3081,6 +3081,126 @@ Emacsmirror, return a MELPA-style recipe; otherwise return nil."
   "Return the current version of the Emacsmirror retriever."
   2)
 
+;;;;;; el-get
+
+;;;;;; shims for el-get variables/functions
+;; If possible, we would like to avoid loading `el-get'
+;; to make use of its recipes. Very few recipes make use of the
+;; `:build` commands. A subset of those recipes use some
+;; variables or functions to compute part of :build commands.
+;; This is an incomplete list of shims, but it should handle most cases.
+
+(defun straight--el-get-emacs ()
+  "Return path to currently running Emacs.
+Shim for `el-get-emacs' variable."
+  (concat invocation-directory invocation-name))
+
+(defun straight--el-get-install-info ()
+  "Shim for `el-get-install-info' variable."
+  (or (executable-find "ginstall-info")
+      (executable-find "install-info")))
+
+(defun straight--el-get-package-directory (package)
+  "Return absolute path to PACKAGE repository.
+Shim for `el-get-package-directory'."
+  (straight--repos-dir (format "%S" package)))
+
+(defmacro straight--recipes-el-get-build-commands (plist)
+  "Convert `el-get' PLIST's :build/* commands to a programmatic elisp form.
+Return nil if no :build/* commands are available."
+  (let* ((names
+          (cl-sort
+           (delq nil
+                 (mapcar (lambda (el)
+                           (when (keywordp el)
+                             (let ((name (symbol-name el)))
+                               (and (string-prefix-p ":build" name) name))))
+                         plist))
+           ;; Sort so :build is the last case in the expanded pcase.
+           (lambda (a b) (> (length a) (length b)))))
+         (systems
+          (mapcar (lambda (name)
+                    (let ((system (replace-regexp-in-string "^:build/?"
+                                                            "" name))
+                          (value (plist-get plist (intern name))))
+                      `(,(if (string-empty-p system)
+                             '_
+                           (list '\` (intern system)))
+                        ,(if (symbolp (car value))
+                             value
+                           (list '\` (if (straight--quoted-form-p value)
+                                         (cadr value)
+                                       value))))))
+                  names)))
+    (list '\,
+          `(cl-letf (((symbol-function #'el-get-package-directory)
+                      (lambda (package)
+                        (straight--repos-dir (format "%S" package))))
+                     (el-get-install-info (straight--el-get-install-info))
+                     (el-get-emacs (straight--el-get-emacs))
+                     (el-get-dir (straight--repos-dir)))
+             (pcase system-type ,@systems)))))
+
+(defun straight--el-get-flatten (arg)
+  "Return a version of ARG as a one-level list."
+  (if (listp arg)
+      (apply 'append (mapcar 'el-get-flatten arg))
+    (list arg)))
+
+(defun straight-recipes-el-get-retrieve (package)
+  "Look up a an `el-get' PACKAGE's recipe.
+PACKAGE must be a symbol. If the package has an `el-get' recipe that
+uses one of the Git fetchers, return it; otherwise return nil."
+  (with-temp-buffer
+    (ignore-errors
+      (progn
+        (insert-file-contents-literally
+         (expand-file-name (concat (symbol-name package) ".rcp")
+                           "recipes/"))
+        (let ((recipe (read (current-buffer)))
+              plist)
+          (straight--with-plist recipe
+              (type name url branch load-path ((:pkgname repo)))
+            (when (stringp name) (setq name (intern name)))
+            (when (member type '(git github))
+              (straight--put plist :type 'git)
+              (if (eq type 'git)
+                  (straight--put plist :repo url)
+                (straight--put plist :host 'github)
+                (straight--put plist :repo repo))
+              (when branch (straight--put plist :branch branch))
+              ;; Differentiate between recipe explicitly declaring
+              ;; :autoloads nil and recipe not declaring :autoloads.
+              (when-let ((autoloads (plist-member recipe :autoloads)))
+                ;; ignore    :autoloads t
+                ;; translate :autoloads nil -> :no-autoloads t
+                ;; el-get also allows a file or list of files which are
+                ;; ignored for now.
+                (unless (cadr autoloads)
+                  (straight--put plist :no-autoloads t)))
+              (when (plist-get recipe :build)
+                (straight--put plist :build
+                               (macroexpand
+                                `(straight--recipes-el-get-build-commands
+                                  ,recipe))))
+              (straight--put
+               plist :files
+               (append '(:defaults)
+                       (cl-remove-if (lambda (file)
+                                       (string= "." file))
+                                     (straight--el-get-flatten load-path))))
+              (list '\` (cons name plist)))))))))
+
+
+(defun straight-recipes-el-get-list ()
+  "Return a list of recipes available in `el-get', as a list of strings."
+  (mapcar #'file-name-sans-extension
+          (straight--directory-files "recipes")))
+
+(defun straight-recipes-el-get-version ()
+  "Return the current version of the `el-get' retriever."
+  1)
+
 ;;;;; Recipe conversion
 
 (defcustom straight-built-in-pseudo-packages '(emacs nadvice python)
--- a/bootstrap.el
+++ b/bootstrap.el
@@ -99,6 +99,10 @@
                                    :local-repo "elpa"
                                    :no-build t)))
 
+(straight-use-recipes '(el-get :type git :host github
+                               :repo "dimitri/el-get"
+                               :no-build t))
+
 (if straight-recipes-emacsmirror-use-mirror
     (straight-use-recipes
      '(emacsmirror-mirror :type git :host github


straight.el
```diff
@@ -2874,6 +2892,78 @@ Emacsmirror, return a MELPA-style recipe; otherwise return nil."
...
+                                 (append '(:defaults)
+                                         (if (listp files)
+                                             files
+                                           `(,files)))))
+                (when-let ((commands
+                            (straight-recipes-el-get-build-commands
+                             recipe)))
+                  (straight--put plist :build commands))
+                (cons name plist)))))
+      (error nil))))
```
raxod502, line 2949: This could be replaced with `ignore-errors`.

raxod502 (dismissed):
> This accounts for roughly 76%

Quite sufficient in my opinion. `straight.el` will likely never support anything that's not on Git, because everyone should be using Git.

> :prepare or :post-init behavior

Can you clarify what you mean by "stomp use-package declarations"? But I suspect that we don't want to support these keywords, as (in the example recipe you gave for Helm) it seems like they sometimes contain actual user configuration which is wholly inappropriate to run automatically (e.g. `(helm-mode)`).

> dependency management

We can add it if we receive a deluge of bug reports about wrong dependencies, otherwise I suggest that we just blame it on the relevant upstream packages.

straight.el
```diff
@@ -2638,7 +2640,23 @@ return nil."
...
+       :files (:defaults "contrib/lisp/*.el")
+       ;;These :build instructions are lifted from el-get's recipe for
+       ;;org. I've modified it for now so there is no need to stub out
+       ;;el-get's functions/varaibles that are in the original recipe,
+       ;;and to test backquoting/elisp evaluation in the :build
+       ;;portion of the recipe.
+       :build
+       `((default .
+           ,(list
+             `(,(concat (when (eq system-name 'berkley-unix) "g")
```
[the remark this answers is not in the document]
progfolio: Not 100% sure, but I assume it's because of this:

> 'gmake' refers specifically to GNU make. 'make' refers to the system's default make implementation; on most Linux distros this is GNU make, but on other unixes, it could refer to some other implementation of make, such as BSD make, or the make implementations of various commercial unixes.
>
> The language accepted by GNU make is a superset of the one supported by the traditional make utility.
>
> By using 'gmake' specifically you can use GNU make extensions without worrying about them being misinterpreted by some other make implementation.


https://stackoverflow.com/questions/1194957/what-is-the-difference-between-gmake-and-make

straight.el
```diff
@@ -2874,6 +2892,78 @@ Emacsmirror, return a MELPA-style recipe; otherwise return nil."
...
+                                           `(,files)))))
+                (when-let ((commands
+                            (straight-recipes-el-get-build-commands
+                             recipe)))
+                  (straight--put plist :build commands))
+                (cons name plist)))))
+      (error nil))))
+
+;;@OPTIMIZATION: Is it faster to allocate a temp buffer or
+;;use a single temp buffer and call erase-buffer per recipe?
```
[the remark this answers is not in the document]
progfolio: Ah, I didn't consider deferring to the next source. Will fix.

straight.el
```diff
@@ -2874,6 +2892,78 @@ Emacsmirror, return a MELPA-style recipe; otherwise return nil."
...
+        backquoted)
+    (dolist (keyword keywords
+                     (when commands
+                       (if backquoted (list '\` commands) commands)))
+      (let ((name (symbol-name keyword)))
+        (when (string-prefix-p ":build" name)
+          (let* ((s (replace-regexp-in-string "^:build/?" "" name))
+                 (system (if (string-empty-p s) 'default (intern s)))
+                 (command (plist-get plist keyword)))
+            (when (eq (car command) '\`) (setq backquoted t)
```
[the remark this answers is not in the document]
progfolio: We're mapping each of el-get's `:build/*` keywords to straight's `:build` alist.
If any of them are backquoted, we want the whole alist backquoted so `straight--run-build-commands` evaluates it properly.

straight.el
```diff
@@ -2874,6 +2892,78 @@ Emacsmirror, return a MELPA-style recipe; otherwise return nil."
...
+                  (straight--put plist :host 'github)
+                  (straight--put plist :repo repo))
+                (when branch (straight--put plist :branch branch))
+                (when files
+                  ;;@QUESTION: Can :defaults be inherited?
+                  (straight--put plist :files
+                                 (append '(:defaults)
+                                         (if (listp files)
+                                             files
+                                           `(,files)))))
```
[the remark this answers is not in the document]
progfolio: > I think we're best off doing it for now, and fixing it when somebody comes to us with a concrete example of how it doesn't work.

That approach sounds  reasonable to me. I started looking through el-get's recipes to get a better idea of all the options/keywords/etc, but there are a lot of them.

You're  correct about the niave handling of el-get's `:autoloads`. I'll take another look.

progfolio:
> Can you clarify what you mean by "stomp use-package declarations"?

Apologies for the vague wording.

> it seems like they sometimes contain actual user configuration which is wholly inappropriate to run automatically (e.g. (helm-mode)).

We're on the same page. The recipe should not be responsible for running any configuration code.

I can see a case for logging that info if it is present in an el-get recipe.
Take mu4e, for example. Its el-get recipe builds the mu executable, but does not run `make install`.
The recipe sets `mu4e-mu-binary` in the `:post-init` section.
Without doing that the user will hit an error when invoking `mu4e` that will tell them to set the binary path,
but they then have to locate the executable in mu's repository.
If we log that configuration it should make it more obvious where the executable is.
I suspect most recipes are just using `:post-init` for less critical configuration, though.

Up to you whether you think implementing that is worth it.

> We can add [dependency management] if we receive a deluge of bug reports about wrong dependencies, otherwise I suggest that we just blame it on the relevant upstream packages

This sounds reasonable to me.

straight.el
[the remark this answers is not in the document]
progfolio: > Seems to me that :autoloads nil should translate into :no-autoloads t, and :autoloads t should be ignored

Changed it to that behavior. If el-get recipe specifies a file or list of files, those are ignored for now, too.

straight.el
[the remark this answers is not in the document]
raxod502: Okay... but I see recipes where it uses gmake if available, else regular make. If it works with regular make, what's the difference running with gmake?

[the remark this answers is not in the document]
progfolio: Not sure to be honest. If you feel it's unnecessary, I'm happy to remove it and hardcode `make`.

straight.el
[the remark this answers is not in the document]
progfolio: Made the variable name more specific and added a comment to document the intent.

README.md
```diff
@@ -1787,6 +1787,42 @@ meaning in a recipe (unknown keywords are ignored but preserved):
...
+  - A list of commands
+  - An alist in which each cons cell's car is a system-type symbol
+  (or the symbol `default`) and cdr is either of the values mentioned above.
+  If an alist is used one of the cons cells must provide a `default` key.
+  - A backquoted form which evaluates to any of the above.
+  - nil, in which case no commands are executed.
+    Note this is not the same as `:no-build` mentioned below.
+    `:no-build` takes precedence over `:build`.
+
+For example:
```
raxod502, line 1814: The indentation is wrong here.

README.md
```diff
@@ -1787,6 +1787,42 @@ meaning in a recipe (unknown keywords are ignored but preserved):
...
 
+* `:build`
+
+  This specifies system commands and/or elisp to be evaluated before
+  before symlinking, generating autoloads, and byte-compiling a package.
+
+  Each command is either an elisp form to be evaluated or a list of
+  strings to be executed in a shell context of the form:
+
+  `("executable" "arg"...)`
```
raxod502, line 1798: This should be a code block.

README.md
```diff
@@ -1787,6 +1787,42 @@ meaning in a recipe (unknown keywords are ignored but preserved):
...
+  Commands are executed in the package's repository directory.
+
+  The `:build` keyword's value may be:
+
+  - A single command
+  - A list of commands
+  - An alist in which each cons cell's car is a system-type symbol
+  (or the symbol `default`) and cdr is either of the values mentioned above.
+  If an alist is used one of the cons cells must provide a `default` key.
+  - A backquoted form which evaluates to any of the above.
```
raxod502, line 1809: Can you add a note here that the backquote is supposed to actually be quoted itself? This is nonstandard and therefore somewhat confusing.

straight.el
```diff
@@ -2626,6 +2628,10 @@ Return a list of package names as strings."
                                       'list source cause)))))))
 
 ;;;;;; Org
+(make-obsolete-variable
```
raxod502, line 2631:
```suggestion

(make-obsolete-variable
```

straight.el
```diff
@@ -2626,6 +2628,10 @@ Return a list of package names as strings."
...
+ 'straight-fix-org
+ "No longer necessary. straight supports build commands."
```
raxod502, line 2633:
```suggestion
 "No longer necessary, as straight.el supports external build commands."
```

straight.el
```diff
@@ -2634,12 +2640,29 @@ return nil."
...
-           :local-repo "org"))
+           :local-repo "org"
+           :build `,(let ((make (if (eq system-name 'berkeley-unix)
+                                    "gmake"
+                                  "make"))
+                          (emacs (concat "EMACS="
+                                         invocation-directory
+                                         invocation-name)))
+                      (list `(,make "oldorg" ,emacs)
+                            '(message "%s" "Done building Org")))))
```
raxod502, line 2651: I don't think we want this. If anything, we want such a message to be generated automatically, with all the arrow prefixing stuff, for every package.

straight.el
```diff
@@ -2874,6 +2897,72 @@ Emacsmirror, return a MELPA-style recipe; otherwise return nil."
...
+              (cons name plist))))))))
+
+(defun straight-recipes-el-get-list ()
+  "Return a list of recipes available in `el-get', as a list of strings."
+  (mapcar #'file-name-sans-extension
+          (straight--directory-files "recipes" "^[^.]" nil t)))
+
+(defun straight-recipes-el-get-version ()
+  "Return the current version of the `el-get' retriever."
+  1)
```
raxod502, line 2965:
```suggestion
  1)
  
```

straight.el
```diff
@@ -2874,6 +2897,72 @@ Emacsmirror, return a MELPA-style recipe; otherwise return nil."
...
+                          (straight--recipes-el-get-build-commands
+                           recipe)))
+                (straight--put plist :build commands))
+              (straight--put plist :files '(:defaults))
+              (cons name plist))))))))
+
+(defun straight-recipes-el-get-list ()
+  "Return a list of recipes available in `el-get', as a list of strings."
+  (mapcar #'file-name-sans-extension
+          (straight--directory-files "recipes" "^[^.]" nil t)))
```
raxod502, line 2961:
```suggestion
          (straight--directory-files "recipes")))
```

The defaults should all work here.

straight.el
```diff
@@ -4423,10 +4552,10 @@ recipe in `straight--build-cache' for the package are updated."
 
 (defun straight--build-package (recipe &optional cause)
   "Build the package specified by the RECIPE.
-This includes symlinking the package files into the build
-directory, building dependencies, generating the autoload file,
-byte-compiling, and updating the build cache. It is assumed that
-the package repository has already been cloned.
+This includes running RECIPE's :build commands, symlinking the package
```
raxod502, line 4555:
```suggestion
This includes running RECIPE's `:build` commands, symlinking the package
```

straight.el
```diff
@@ -4061,6 +4150,46 @@ other value, the behavior is not specified."
...
+  (straight--with-plist recipe (build package local-repo)
+    (when build
+      (let ((c (car build)))
+        (when (or (eq c '\`) (eq c 'quote)) (setq build (eval build))))
+      (when-let ((commands (or (alist-get system-type build)
+                               (alist-get 'default build)
+                               build))
+                 (repo (straight--repos-dir (or local-repo package))))
+        (let ((default-directory repo))
+          ;;allow a single command or a list of commands
```
raxod502, line 4185: Just for the sake of consistency, would you mind capitalizing / putting periods at the end of comments?

README.md
progfolio: I'm not sure what you mean here. The backquote doesn't need to be explicitly quoted. e.g.

```emacs-lisp
(:build `((static) ,(dynamic)))
```

straight.el
progfolio: Vestigial testing code. Will remove.

progfolio:
Premature CI failure due to smoke test. Addressed in #554

README.md
raxod502: Typically one would use a backquote to produce a recipe which does not itself contain any quotes, e.g. [in the example in the README](https://github.com/raxod502/straight.el/tree/5d046ad09413f3aba6198229e392cdd3f07e77ee#the-recipe-format). In your syntax, the evaluated recipe _itself_ contains a backquote. You need to do this:

```elisp
(straight-use-package '(package :build `(foo bar)))
```

In other words, there are two quotes: the outer quote is quoting the inner backquote. I haven't seen any other package in Emacs where this kind of syntax is used, hence why I thought some clarification was in order for the docs.

straight.el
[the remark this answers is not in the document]
raxod502: I received an answer at https://github.com/dimitri/el-get/issues/2781, and it sounds based on that like using `gmake` is indeed appropriate in order to support both BSD and GNU systems.

raxod502:
I've done my best to respond to all the comment threads, but GitHub makes them a bit difficult to find. Let me know if I missed anything, or if you need anything else.

README.md
progfolio: I see what you mean, but in this case the user would be able to backquote the whole recipe and evaluate the `:build` command list before it hits `straight--run-build-commands`. e.g.

```emacs-lisp
(straight-use-package `(example :build ,(list  'foo 'bar)))
```

progfolio: Though, experimenting, this does fail when using the use-package integrated `:straight` keyword. For example:

```emacs-lisp
(use-package doct
  :straight `(doct  :host github :repo "progfolio/doct" :branch "development"
                    :build ,(list "ls" "-lah"))
  :commands (doct))
```

expands to:

```emacs-lisp
(progn
  (straight-use-package '\`)
  (straight-use-package
   '(doct :host github :repo "progfolio/doct" :branch "development" :build ,(list "ls" "-lah")))
  (unless
      (fboundp 'doct)
    (autoload
      (function doct)
      "doct" nil t)))
```

Shouldn't the `:straight` keyword allow the same type of input as `straight-use-package`?

progfolio: Now that #558 is merged, do you still feel the note about backquoting is necessary?

raxod502: Yes, because the confusion doesn't really relate to `use-package` at all: passing a recipe with a dynamic build recipe to `straight-use-package` still entails two layers of quotes (as shown in my previous comment), which I think would be surprising to most users.

progfolio: I can see how the behavior of your example:

    (straight-use-package '(package :build `(foo bar)))

would be surprising.

The double quoting really only comes into play when reading a recipe from a static file (e.g. el-get's recipes, or any future format that follows that pattern).
Perhaps a better solution is to decompose the logic that read/evals recipe files into its own function. Then the el-get/recipe providers can use that if they need to and the generic case
will obey the normal quoting/backquoting evaluation rules. I'll take a look into that today.

progfolio:
--text follows this line-

Thought on this and implemented the best solution I could think 
of.
I've added generic support for evaluating backquoted recipes in 
`straight--convert-recipe`.

Now a recipe retrieval function (e.g. 
`straight-recipes-el-get-retrieve`) can return a backquoted 
recipe.
For example:

```emacs-lisp
(straight-recipes-retrieve 'org-mode '(el-get))
```

```emacs-lisp
`(org-mode :type git :repo 
"https://code.orgmode.org/bzg/org-mode.git"
           :build
           ((berkeley-unix \` ,(mapcar
                                (lambda
                                  (target)
                                  (list "gmake" target
                                        (concat "EMACS="
                                                (shell-quote-argument 
                                                el-get-emacs))))
                                '("oldorg")))
            (default \` ,(mapcar
                           (lambda
                            (target)
                            (list "make" target
                                  (concat "EMACS="
                                          (shell-quote-argument 
                                          el-get-emacs))))
                          '("oldorg"))))
           :files
           (:defaults))
```

Which `straight--convert-recipe` evaluates.
This handles the case of recipes stored in files without the 
implicit evaluation when a user provides a recipe.

```emacs-lisp
;; No longer evaluated; errors as expected.
(straight-use-package '(foo :build ,(list 'message "test")))

;; works as expected
(straight-use-package `(foo :build ,(list 'message "test")))
```

Let me know what you think.
Date: Mon, 31 Aug 2020 20:07:11 -0400
Message-ID: <87tuwiqs0g.fsf@gmail.com>

README.md
progfolio: Thought on this and implemented the best solution I could think
of.
I've added generic support for evaluating backquoted recipes in
`straight--convert-recipe`.

Now a recipe retrieval function (e.g.
`straight-recipes-el-get-retrieve`) can return a backquoted
recipe.
For example:

```emacs-lisp
(straight-recipes-retrieve 'org-mode '(el-get))
```

```emacs-lisp
`(org-mode :type git :repo
"https://code.orgmode.org/bzg/org-mode.git"
           :build
           ((berkeley-unix \` ,(mapcar
                                (lambda
                                  (target)
                                  (list "gmake" target
                                        (concat "EMACS="
                                                (shell-quote-argument
                                                el-get-emacs))))
                                '("oldorg")))
            (default \` ,(mapcar
                           (lambda
                            (target)
                            (list "make" target
                                  (concat "EMACS="
                                          (shell-quote-argument
                                          el-get-emacs))))
                          '("oldorg"))))
           :files
           (:defaults))
```

Which `straight--convert-recipe` evaluates.
This handles the case of recipes stored in files without the
implicit evaluation when a user provides a recipe.

```emacs-lisp
;; No longer evaluated; errors as expected.
(straight-use-package '(foo :build ,(list 'message "test")))

;; works as expected
(straight-use-package `(foo :build ,(list 'message "test")))
```

Let me know what you think.

raxod502: Now writing from California: I actually thought what you had before seemed reasonable, I just thought that it should be documented clearly. I do suppose you are right, though, that simply implementing this behavior at the top level will obviate the need to do the same thing again when some other part of the recipe system needs it. The only thing I am confused about is the presence of those `` \` `` characters in the output of the command you showed. Why not just one at the top level?

progfolio: I'll keep working on this. One thing I completely forgot about is bootstrapping. Went to use this branch on in a clean environment and it didn't work the way I expected it would. I'll see if I can clean up the recipe output while I'm at that, too.

progfolio: Think we may have some progress.
Now when retrieving a recipe from el-get, we return a backquoted list similar to above, but we strip the leading quotes in the various build commands. This should produce a top-level backquoted recipe that still evaluates properly. 

I've tried a few recipes so far and it seems to work.
Here's Org built from el-get:

```emacs-lisp
(straight-bug-report :preserve nil :user-dir "el-get-recipe-support.straight" :pre-bootstrap
                     (setq straight-repository-user "progfolio")
                     (setq straight-repository-branch "feat/el-get-recipe-support")
                     :post-bootstrap
                     ;;Should we register el-get recipes by default?
                     (straight-use-recipes 'el-get)
                     (message "straight-recipes-retrieve: %S"
                              (straight-recipes-retrieve 'org-mode
                                                         '(el-get)))
                     (message "org-version: %S"
                              (straight-use-package
                               '(org-mode :source el-get :post-build
                                          (message "%S"
                                                   (org-version nil 'full))))))

```
</details>

- Test run at: `2020-09-24 19:22:20`
- system-type: `gnu/linux`
- straight-version: `prerelease (HEAD -> feat/el-get-recipe-support, fork/feat/el-get-recipe-support) 11aa11f 2020-09-24`
- emacs-version: `GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, X toolkit, cairo version 1.17.3, Xaw3d scroll bars)
 of 2020-09-15`

```emacs-lisp
Bootstrapping straight.el...
Building straight...
Building straight...done

Test run with version: prerelease (HEAD -> feat/el-get-recipe-support, origin/feat/el-get-recipe-support) 11aa11f 2020-09-24
Looking for org-mode recipe → Cloning el-get...
Looking for org-mode recipe → Cloning el-get...done
Looking for org-mode recipe → Building el-get...
Looking for org-mode recipe → Building el-get...done

straight-recipes-retrieve: `(org-mode :type git :repo "https://code.orgmode.org/bzg/org-mode.git" :build ((berkeley-unix ,(mapcar (lambda (target) (list "gmake" target (concat "EMACS=" (shell-quote-argument el-get-emacs)))) '("oldorg"))) (default ,(mapcar (lambda (target) (list "make" target (concat "EMACS=" (shell-quote-argument el-get-emacs)))) '("oldorg")))) :files (:defaults))
Cloning org-mode...
Cloning org-mode...done
Building org-mode...

"Org mode version 9.4 (release_9.4-29-gbc9664 @ /tmp/el-get-recipe-support.straight/straight/build/org-mode/)"
Building org-mode...done

org-version: t

```
</details>

README.md
```diff
@@ -2769,34 +2823,6 @@ different package managers), `straight.el` will helpfully disable
 `:ensure` whenever you include `:straight` in a `use-package` form.
 See [#425].
 
-#### Integration with Org
```
raxod502, line 2772: The table of contents should be regenerated with `make toc`.

README.md
```diff
@@ -1890,6 +1890,56 @@ meaning in a recipe (unknown keywords are ignored but preserved):
...
+  Otherwise, it is identical to the `:build` keyword in terms of the values
+  it accepts and how it is executed.
+
+    For example:
+
+        (straight-use-package
+         '( example :type git :host github :repo "user/example.el"
+            :build (("./pre-build.sh") (message "hi"))
+            :post-build (("./post-build.sh") (message "bye"))))
+
```
raxod502, line 1942: Do we document the new general facility for passing a backquoted recipe to force re-evaluation?

straight.el
```diff
@@ -2745,18 +2747,38 @@ Return a list of package names as strings."
...
+        :type git :repo "https://code.orgmode.org/bzg/org-mode.git"
+        :local-repo "org"
+        :files (:defaults "contrib/lisp/*.el")
+        :build ,(let ((make (if (eq system-type 'berkeley-unix)
+                                "gmake"
+                              "make"))
+                      (emacs (concat "EMACS="
+                                     invocation-directory
+                                     invocation-name)))
+                  `(,make "oldorg" ,emacs))))
```
raxod502, line 2781: I do rather like this. Perhaps in el-get recipes, we could simply translate the multiple `:build` recipes into conditional code. That would be a more elegant recipe format than what el-get uses by default.

README.md
progfolio: Addressed in #535

straight.el
progfolio: I suppose we could do something like that.
Just to make sure I'm on the same page, let's take evil's el-get recipe as an example:

<details><summary>"evil.rcp"</summary>

    ( :name evil
      :website "https://github.com/emacs-evil/evil"
      :description "Evil is an extensible vi layer for Emacs. It
           emulates the main features of Vim, and provides facilities
           for writing custom extensions."
      :type github
      :pkgname "emacs-evil/evil"
      :features evil
      :depends (undo-tree goto-chg)
      :build (("make" "info" "all"))
      :build/berkeley-unix (("gmake" "info" "all"))
      :build/darwin `(("make" ,(format "EMACS=%s" el-get-emacs) "info" "all"))
      :info "doc")

would convert to:

    `( evil
       :type git
       :host github
       :repo "emacs-evil/evil"
       :build ,(pcase system-type
                 (`berkeley-unix `(("gmake" "info" "all")))
                 (`gnu/linux `(("make" ,(format "EMACS=%s" "emacs") "info" "all")))
                 (_ `(("make" "info" "all"))))
       :files (:defaults))

</details>

<details><summary>"org-mode.rcp"</summary>

    ( :name org-mode
      :website "http://orgmode.org/"
      :description "Org-mode is for keeping notes, maintaining ToDo lists, doing project planning, and authoring with a fast and effective plain-text system."
      :type git
      :url "https://code.orgmode.org/bzg/org-mode.git"
      :info "doc"
      :build/berkeley-unix `,(mapcar
                              (lambda (target)
                                (list "gmake" target (concat "EMACS=" (shell-quote-argument el-get-emacs))))
                              '("oldorg"))
      :build `,(mapcar
                (lambda (target)
                  (list "make" target (concat "EMACS=" (shell-quote-argument el-get-emacs))))
                '("oldorg"))
      :load-path ("." "contrib/lisp" "lisp")
      :load ("lisp/org-loaddefs.el"))

would convert to:

    `( org-mode
       :type git
       :repo "https://code.orgmode.org/bzg/org-mode.git"
       :build ,(pcase system-type
                 (`berkeley-unix
                  `,(mapcar (lambda (target)
                              (list "gmake" target
                                    (concat "EMACS=" (shell-quote-argument el-get-emacs))))
                            '("oldorg")))
                 (_ `,(mapcar (lambda (target)
                                (list "make" target
                                      (concat "EMACS=" (shell-quote-argument el-get-emacs))))
                              '("oldorg"))))
       :files (:defaults))

<details>

I have no problem with going in this direction.
I think (haven't tried it yet) it would only require a slight modification to `straight--recipes-el-get-build-commands`.

This also raises the question of whether or not straight's `:build` and `:post-build` keywords should accept the `system-type → commands` alist option.
Recipe authors could encode system specific behavior with elisp as above.
This has the benefit of keeping the API simple.
Also, the system-specific `:build` case only makes up a small portion of el-get's recipes (95 out of 1782 last I checked), so it's reasonable to assume most recipe authors will not utilize the alist option.

I agree it does seem simpler and more elegant, so I'll experiment with it and see how it turns out.

raxod502: I don't think we need a `system-type → commands` option. The way of `straight.el` has typically been to implement the most reasonable interface only, and translate from legacy formats to our version.

progfolio: I've removed the system-type alist option from straight's `:build`, `:post-build` keywords
in #535. To account for this in el-get recipe translation we translate from the system-specific `:build/*`
key val pairs to a pcase as described above.
Additionally I've shimmed out some of the common el-get functions/variables used in computing el-get recipe `:build/*` commands.
e.g. Running:

    (let ((default-directory (straight--repos-dir "el-get")))
      (straight-recipes-el-get-retrieve 'magit))

returns:

    `(magit :type git
            :host github
            :repo "magit/magit"
            :branch "master"
            :build ,(pcase system-type
                      (`berkeley-unix
                       `(("gmake" ,(format "EMACSBIN=%s"
                                           (straight--el-get-emacs))
                          "docs")
                         ("touch" "lisp/magit-autoloads.el")))
                      (`windows-nt
                       (with-temp-file "lisp/magit-autoloads.el" nil))
                      (_
                       `(("make" ,(format "EMACSBIN=%s"
                                          (straight--el-get-emacs))
                          "docs")
                         ("touch" "lisp/magit-autoloads.el"))))
            :files
            (:defaults))

Which on my (gnu/linux) system evaluates to:

    (magit :type git
           :host github
           :repo "magit/magit"
           :branch "master"
           :build (("make" "EMACSBIN=/usr/bin/emacs" "docs")
                   ("touch" "lisp/magit-autoloads.el"))
           :files
           (:defaults))

during `straight--convert-recipe`.
This should play nice with `straight--run-build-commands`.

    ;; -*- lexical-binding: t; -*-
    (straight-bug-report
      :user-dir "el-get-magit-test"
      :preserve nil
      :pre-bootstrap
      (setq straight-repository-user "progfolio")
      (setq straight-repository-branch "feat/el-get-recipe-support")
      :post-bootstrap
      (straight-use-recipes 'el-get)
      (straight-use-package
       '(magit :source el-get :post-build ((message "magit-version: %S" (magit-version))))))

The following works for me:


```emacs-lisp
(straight-bug-report
  :user-dir "el-get-magit-test"
  :pre-bootstrap
  (setq straight-repository-user "progfolio")
  (setq straight-repository-branch "feat/el-get-recipe-support")
  :post-bootstrap
  (straight-use-recipes 'el-get)
  (straight-use-package
   '(magit :source el-get :post-build
           ((message "magit-version: %S"
                     (magit-version))))))
```
</details>

- Test run at: `2020-09-29 14:53:54`
- system-type: `gnu/linux`
- straight-version: `prerelease (HEAD -> feat/el-get-recipe-support, fork/feat/el-get-recipe-support) 5d9abcc 2020-09-29`
- emacs-version: `GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, X toolkit, cairo version 1.17.3, Xaw3d scroll bars)
 of 2020-09-27`

```emacs-lisp
Bootstrapping straight.el...
Building straight...
Building straight...done

Test run with version: prerelease (HEAD -> feat/el-get-recipe-support, origin/feat/el-get-recipe-support) 5d9abcc 2020-09-29
Looking for magit recipe → Cloning el-get...
Looking for magit recipe → Cloning el-get...done
Looking for magit recipe → Building el-get...
Looking for magit recipe → Building el-get...done
Cloning magit...
Cloning magit...done
Building magit...
Building magit → Cloning emacs-async...
Building magit → Cloning emacs-async...done
Building magit → Building async...
Building magit → Building async...done
Building magit → Cloning dash.el...
Building magit → Cloning dash.el...done
Building magit → Building dash...
Building magit → Building dash...done
Building magit → Building git-commit...
Building magit → Building git-commit → Cloning transient...
Building magit → Building git-commit → Cloning transient...done
Building magit → Building git-commit → Building transient...
Building magit → Building git-commit → Building transient...done
Building magit → Building git-commit → Cloning with-editor...
Building magit → Building git-commit → Cloning with-editor...done
Building magit → Building git-commit → Building with-editor...
Building magit → Building git-commit → Building with-editor...done
Building magit → Building git-commit...
Building magit → Building git-commit...done
Building magit...

magit-version: "v2.90.1-1064-g6c1c92e5"
Building magit...done

```
</details>

README.md
progfolio: I'm not exactly sure what you mean here. Could you give an example?

straight.el
raxod502: That looks phenomenal!

README.md
raxod502: Sorry, the place where I added this comment was not helpful. I should have put it at this snippet from https://github.com/raxod502/straight.el/pull/535:

```diff
+        ;; Recipes retrieved from files may be backquoted.
+        (when (straight--quoted-form-p full-melpa-style-recipe)
+          (setq full-melpa-style-recipe (eval full-melpa-style-recipe)))
```

Namely there's a new feature where if `straight-use-package` gets a recipe (either directly from the user or retrieved from a recipe repository) that literally contains a backquote at its start, then it'll be evaluated an additional time ("re-evaluated"). This feature is not actually illustrated by the example in the documentation that I commented on here, where the backquote gets handled by the Elisp reader and doesn't make its way into `straight-use-package`.

Does that make sense?

progfolio: I see what you mean, though I can't think of a case where a user would intentionally opt for 
passing a quoted backquoted list over a backquoted list.

I can add a note that, due to the extra evaluation, they are equivalent.
Perhaps with an example similar to:

```emacs-lisp
(equal (eval '`(,(identity 'foo))) `(,(identity 'foo))) ;;t
```

But using `straight-use-package`.

raxod502: Yes, I agree. If I understand correctly how this is implemented, the evaluation happens before the recipe interfaces with the modification detection system, so a change in the backquoted logic (e.g. check against `system-type` returns a different value than it did before) should still trigger a package rebuild. In that case they should be entirely equivalent, and I agree with you. There should be no reason for the user to use the doubly-quoted version. But the feature should probably be documented in the section on writing custom recipe repositories, as that's where it is useful.

progfolio: Good point. This is important for recipe authors to know. I'll add to that section with some examples.

progfolio: Added in #535. I included it under the "Defining new recipe repositories" section. Tried to provide a simple example. Let me know what you think.

raxod502: see below

progfolio:
Do we want to enable el-get recipes by default, or require the user to:
```emacs-lisp
(straight-use-recipes 'el-get)
```
in their init?

Other than that, anything else you think this needs?

straight.el
```diff
@@ -3032,6 +3032,125 @@ Emacsmirror, return a MELPA-style recipe; otherwise return nil."
...
+                  ("\\(?:el-get-emacs\\)" . "(straight--\\&)")
+                  ;; el-get-install-info -> (striaght--el-get-install-info)
+                  ("\\(?:el-get-install-info\\)" . "(straight--\\&)")
+                  ;; (el-get-package-directory args) ->
+                  ;; (straight--el-get-package-directory args)
+                  ("\\(?:(\\(el-get-package-directory\\)\\([^z-a]*?\\))\\)"
+                   . "(straight--\\1\\2)")
+                  ;; if quoted (e.g. as arg to `apply' or `funcall')
+                  ("\\(?:'\\(el-get-package-directory\\)\\)"
+                   . "'\\straight--\\1")))
```
raxod502, line 3071: This is fine, but another approach which might be more robust is to evaluate the forms in a context where these shim functions are bound using `cl-letf`. Then unexpected whitespace and stuff like that wouldn't cause problems with the regexp replacements.

raxod502:
I think for now it is fine to enable the El-Get recipes globally. Right now, we don't have a great story for how to customize the default recipe repositories (by which I mean it's totally impossible); see #161 where I suggest forking `straight.el` as the easiest way to do this. Ideally, that could be fixed in future; I have just filed https://github.com/raxod502/straight.el/issues/619 to track this.

But for this pull request, yeah, I'd suggest just putting in a new form to `bootstrap.el`. I'm not sure what the appropriate ordering is. Basically, we want to order the recipe repositories so that the ones most likely to work come first. That's why MELPA is first (basically everything works, except for the default branch shenanigans) and Emacsmirror is last (many repos are broken because its recipes are not manually curated). So I guess you can play around and see how reliable the El-Get stuff seems to be, and make a judgment about whether it's more or less reliable than GNU ELPA, etc.

> Other than that, anything else you think this needs?

I think you're good to go! Thanks again.

straight.el
raxod502: I can produce an example of doing this in case you haven't seen `cl-letf` and `symbol-function` used together before.

progfolio: Agreed that the `cl-letf` approach is more robust and the correct way to go about this. I'll work something out for this soon.

progfolio:
Recipe `:build` lists are now wrapped in a `cl-letf` for the shims. Definitely the more robust approach. Thank you for catching that.

I've placed el-get after GNU ELPA and before Emacsmirror in \`straight-recipe-repositories'.
This seems like fair placement to me especially considering users can override the source on a per recipe basis.
Ran a clean build on my personal laptop and everything worked smoothly for the el-get recipes I am using.

Added support for el-get's `:load-path` keyword as it maps nicely to our `:files` keyword.

I also ran an automated test on all the el-get recipes we list:


```emacs-lisp
(straight-bug-report
  :user-dir "el-get-recipe-support.straight"
  :preserve t
  :interactive t
  :pre-bootstrap
  (setq straight-repository-user "progfolio")
  (setq straight-repository-branch "feat/el-get-recipe-support")
  :post-bootstrap
  (let* ((default-directory (straight--repos-dir "el-get"))
          (recipes (delq nil (mapcar (lambda (name)
                                      (straight-recipes-el-get-retrieve (intern name)))
                                    (straight-recipes-el-get-list)))))
    (message "total recipes retreived: %d" (length recipes))
    (setq debug-on-error nil)
    ;;adding MELPA after el-get to cover a couple dependencies which do not have el-get recipes...
    (setq straight-recipe-repositories '(el-get melpa))
    (dolist (recipe recipes)
      (with-demoted-errors "ELGETERR: %S"
        (when (straight--quoted-form-p recipe)
          (setq recipe (eval recipe)))
        (let ((r (cdr recipe)))
          (setcdr recipe (straight--put r :source '(el-get melpa)))
          (straight-use-package recipe))))))
```


The results of the test puts us at roughly 76% coverage of el-get's 1789 recipes.

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">


<colgroup>
<col  class="org-left" />

<col  class="org-right" />

<col  class="org-right" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">type</th>
<th scope="col" class="org-right">count</th>
<th scope="col" class="org-right">% of total</th>
</tr>
</thead>

<tbody>
<tr>
<td class="org-left">recipes listed</td>
<td class="org-right">1373</td>
<td class="org-right">1.00</td>
</tr>
</tbody>

<tbody>
<tr>
<td class="org-left">successfully cloned</td>
<td class="org-right">1303</td>
<td class="org-right">0.95</td>
</tr>


<tr>
<td class="org-left">inaccessible repository</td>
<td class="org-right">45</td>
<td class="org-right">0.03</td>
</tr>


<tr>
<td class="org-left">workarounds</td>
<td class="org-right">17</td>
<td class="org-right">0.01</td>
</tr>


<tr>
<td class="org-left">misc</td>
<td class="org-right">8</td>
<td class="org-right">0.01</td>
</tr>
</tbody>

<tbody>
<tr>
<td class="org-left">TOTAL</td>
<td class="org-right">1373</td>
<td class="org-right">1.00</td>
</tr>
</tbody>
</table>

## inaccesible repos

I believe most of these should be resolved or removed upstream and have opened:

<https://github.com/dimitri/el-get/issues/2789>


<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">


<colgroup>
<col  class="org-left" />

<col  class="org-left" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">Package</th>
<th scope="col" class="org-left">Reason</th>
</tr>
</thead>

<tbody>
<tr>
<td class="org-left">ampc</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">company-childframe</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">company-edbi</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">company-tern</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">darcula-theme</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">deft</td>
<td class="org-left">jblevins.org[0: 69.164.221.6]: errno=Connection refused</td>
</tr>


<tr>
<td class="org-left">discover-my-major</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">edbi-database-url</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">edbi-sqlite</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">emacs-sos</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">flycheck-protobuf</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">flyguess</td>
<td class="org-left">fatal: unable to look up git.naquadah.org (port 9418) (Name or service not known)</td>
</tr>


<tr>
<td class="org-left">gtranslate</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">header-button</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">helm-gist</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">helm-google</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">helm-ipython</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">hrb-mode</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">hsnippets</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">ide-skel</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">identica-mode</td>
<td class="org-left">repository <a href="http://git.savannah.gnu.org/cgit/identica-mode.git/">http://git.savannah.gnu.org/cgit/identica-mode.git/</a> not found</td>
</tr>


<tr>
<td class="org-left">js3-refactor</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">listing</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">mo-git-blame</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">nurumacs</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">opam-mode</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">org-doing</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">org-sync</td>
<td class="org-left">orgmode.org[0: 45.77.159.66]: errno=Connection refused</td>
</tr>


<tr>
<td class="org-left">pbcopy</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">pov-mode</td>
<td class="org-left">Could not resolve host: git.gitorious.org</td>
</tr>


<tr>
<td class="org-left">processing-mode (processing-emacs dependency)</td>
<td class="org-left">Could not resolve host: git.gitorious.org</td>
</tr>


<tr>
<td class="org-left">pythonbrew-mini</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">reflected-buffers</td>
<td class="org-left">submodule bad ref: Failed to recurse into submodule path 'util/ecukes'</td>
</tr>


<tr>
<td class="org-left">relative-line-numbers</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">rt-liberation</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">sourcegraph-mode</td>
<td class="org-left">repo 404</td>
</tr>


<tr>
<td class="org-left">tempbuf-mode</td>
<td class="org-left">Could not resolve host: git.gitorious.org</td>
</tr>


<tr>
<td class="org-left">term+mode</td>
<td class="org-left">depends on multi-mode, submodule repo 404 (available in emacsattic)</td>
</tr>


<tr>
<td class="org-left">twitter</td>
<td class="org-left">fatal: unable to connect to git.busydoingnothing.co.uk: errno=Connection timed out</td>
</tr>


<tr>
<td class="org-left">vimpulse</td>
<td class="org-left">Could not resolve host: git.gitorious.org</td>
</tr>


<tr>
<td class="org-left">wb-line-number</td>
<td class="org-left">repo 404</td>
</tr>
</tbody>
</table>

And <https://github.com/raxod502/straight.el/issues/149> might be able to take care of the following (depending on how it is implemented)

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">


<colgroup>
<col  class="org-left" />

<col  class="org-left" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">Package</th>
<th scope="col" class="org-left">Reason</th>
</tr>
</thead>

<tbody>
<tr>
<td class="org-left">mdfind-dired</td>
<td class="org-left">github gist</td>
</tr>


<tr>
<td class="org-left">ob-objc</td>
<td class="org-left">github gist</td>
</tr>


<tr>
<td class="org-left">spotlight-comment</td>
<td class="org-left">github gist</td>
</tr>


<tr>
<td class="org-left">why3-mode</td>
<td class="org-left">github gist</td>
</tr>
</tbody>
</table>


## workarounds

The following recipes are easily amended via inheritance.
I figure I'd make a list we can point people to should the need arise:

- **capture**

```emacs-lisp
(capture :source el-get :files (:defaults "src/*"))
```

- **color-theme-tomorrow**

```emacs-lisp
(color-theme-tomorrow :source el-get :files (:defaults "GNU Emacs"))
```

  - **company-emoji**

```emacs-lisp
(company-emoji :source el-get :branch "trunk")
```

- **ensime**

```emacs-lisp
(ensime :source el-get :branch "2.0")
```

- **google-calendar**

```emacs-lisp
(google-calendar :source el-get :build ("/usr/bin/emacs" "--batch" "--execute" "(progn (require 'ob-tangle) (org-babel-tangle-file \"./google-calendar.org\"))"))
```

- **magit**

```emacs-lisp
(magit :source el-get
       :build (let ((emacs (concat invocation-directory invocation-name)))
                (pcase system-type
                  (`berkeley-unix `("gmake" ,(format "EMACSBIN=%s" emacs) "docs"))
                  (_ `("make" ,(format "EMACSBIN=%s" emacs) "docs")))))
```

- **nim-emacs-module**

```emacs-lisp
(nim-emacs-module :source el-get :files (:defaults "extra"))
```

- **nterm**

```emacs-lisp
(nterm :source el-get :branch "web-import")
```

- **proof-general**

```emacs-lisp
(proof-general :source el-get :files (:defaults "generic"))
```

- **react-snippets**

```emacs-lisp
(react-snippets :source el-get :files (:defaults "snippets"))
```

- **sclang**

```emacs-lisp
(sclang :source el-get :files (:defaults "editors/scel/el/*"))
```

- **sclang-snippets**

```emacs-lisp
(sclang-snippets :source el-get :files (:defaults "snippets"))
```

- **tdtd**

```emacs-lisp
`(tdtd :build ,(list (concat invocation-directory invocation-name)
                     "-batch" "-Q" "-nw" "-l" "tdtd-make-regexp.el"
                     "-l" "tdtd-font-maker.el"
                     "-f" "dtd-make-tdtd-font"))
```

- **wanderlust**
  incompatible :build format
  either override the :build command list, or use MELPA recipe repository.

- **yasnippet-snippets**

```emacs-lisp
(yasnippet-snippets :source el-get :files (:defaults "snippets"))
```

- **yasnippets-orgmode**

```emacs-lisp
`(yasnippets-orgmode :source el-get :build ("./install.sh" ,(car yas-snippet-dirs)))
```

- **rust-racer**

Manually set racer-cmd in init file


<a id="org2f68891"></a>

## miscellaneous

The following package recipes are miscellany which provide no elisp packages to build or have been deprecated:

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">


<colgroup>
<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">Package</th>
<th scope="col" class="org-left">Reason</th>
<th scope="col" class="org-left">Solution</th>
</tr>
</thead>

<tbody>
<tr>
<td class="org-left">paredit-cheatsheet</td>
<td class="org-left">not a library (just a pdf/tex doc)</td>
<td class="org-left">&#xa0;</td>
</tr>


<tr>
<td class="org-left">pier-mode</td>
<td class="org-left">deprecated</td>
<td class="org-left"><a href="https://github.com/DamienCassou/pillar-mode">https://github.com/DamienCassou/pillar-mode</a></td>
</tr>


<tr>
<td class="org-left">swank-clojure</td>
<td class="org-left">deprecated</td>
<td class="org-left">author suggests: <a href="https://github.com/clojure-emacs/cider">https://github.com/clojure-emacs/cider</a></td>
</tr>


<tr>
<td class="org-left">tern-lint</td>
<td class="org-left">not elisp library (tern plugin)</td>
<td class="org-left">&#xa0;</td>
</tr>


<tr>
<td class="org-left">realgud</td>
<td class="org-left">dependency on loc-changes during :build step</td>
<td class="org-left">(possible to move this to :post-build)?</td>
</tr>


<tr>
<td class="org-left">rope</td>
<td class="org-left">python library</td>
<td class="org-left">&#xa0;</td>
</tr>


<tr>
<td class="org-left">ropemacs</td>
<td class="org-left">python library</td>
<td class="org-left">&#xa0;</td>
</tr>


<tr>
<td class="org-left">ropemode</td>
<td class="org-left">python library</td>
<td class="org-left">&#xa0;</td>
</tr>
</tbody>
</table>


Let me know what you think.

raxod502:
Wow, this is great! I'm reminded of the epic tables Jonas Bernoulli makes in his quest to normalize Emacs packaging! e.g. https://github.com/melpa/melpa/issues/2342#issuecomment-291906557

> #149 might be able to take care of the following (depending on how it is implemented)

You might be surprised to hear we can clone Gists as Git repositories!

```elisp
(straight-use-package '(mdfind-dired :repo "https://gist.github.com/900452.git" :local-repo "mdfind-dired"))
```

> capture, color-theme-tomorrow, nim-emacs-module, proof-general, react-snippets, sclang, sclang-snippets, yasnippet-snippets

These seem like they suffer from similar problems most likely. Do you know what incompatibility with the El-Get recipe format leads these recipes to work under El-Get out of the box but not under `straight.el`? Maybe some edge case with the `:files` semantics, or `:load-path`?

Does the weird leading slash on the sclang `:load-path` directive cause problems?

> company-emoji, ensime, nterm

These are default branch problems which will be solved by #543, right?

> google-calendar

Is this because El-Get automatically tangles Org files if it finds them?

> magit

I assume this is because the weird hack in the upstream recipe with the autoloads file breaks things for `straight.el`? Seems ok to me, as the MELPA recipe will be preferred by default so people won't see this one.

> tdtd

Difficult for me to tell exactly what inspired the original recipe for this one, but is the issue `el-get-emacs` or some backtick thingie?

> wanderlust

What feature of the format is incompatible, out of curiosity? I agree it's no big deal because there's a MELPA recipe.

> rust-racer

It's not actually clear to me why this is needed. Like it seems gratuitous to build Racer from source, when one could instead just install it through the system package manager and then install the `racer` package from MELPA which just does a standard `$PATH` lookup for the binary.

aspiers:
@progfolio This is absolutely epic work.  Many, *many* thanks!  That is all I have to say :grin:

progfolio:
> You might be surprised to hear we can clone Gists as Git repositories!

Hey! That works, too. I wasn't aware of that. Thanks. I'll add those as workarounds.

> These seem like they suffer from similar problems most likely. Do you know what incompatibility with the El-Get recipe format leads these recipes to work under El-Get out of the box but not under straight.el? Maybe some edge case with the :files semantics, or :load-path?

I believe the way el-get works is just by cloning the entire repository into `user-emacs-directory/el-get/repo`. Files are then compiled in place in the repo, so there's no linking to a separate build folder. Some recipes just rely on this structure and don't specify what we would need to in our `:files` directive (e.g. react-snippets.el assumes `snippets` is in the same directory as itself). Others make use of el-get's `:prepare` (e.g. color-theme-tomorrow) or `:autoloads` (e.g. proof-general) keywords to load files. Some, as far as I can tell, don't work out of the box. If I install `capture` via el-get and then try to `M-x load-library capture`, I have to manually drill down to the `src` folder to select an elisp file. I may be missing some other set-up step, but it's not as simple as `M-x el-get-install capture`.


> These are default branch problems which will be solved by #543, right?

Yes, #543 should cover those.

> google-calendar
> Is this because El-Get automatically tangles Org files if it finds them?

`goole-calendar.rcp` uses el-get's `:post-init` keyword to tangle the source file if it hasn't already been tangled. 

> magit
> I assume this is because the weird hack in the upstream recipe with the autoloads file breaks things for straight.el? Seems ok to me, as the MELPA recipe will be preferred by default so people won't see this one.

Exactly right. The el-get magit recipe creates an empty `magit-autoloads.el` which is never populated in our case.
Agreed that the MELPA recipe is the way to go.

> tdtd
> Difficult for me to tell exactly what inspired the original recipe for this one, but is the issue el-get-emacs or some backtick thingie?

el-get allows nested lists for shell command arguments:

https://github.com/dimitri/el-get/blob/6138ce5b0a9df779e33ef42eadc4c15a5cf65f39/el-get.texi#L1006-L1012


For example, tdtd's `:build` instructions evaluate to:

``emacs-lisp
(("/usr/bin/emacs"
  ("-batch" "-Q" "-nw" "-l" "tdtd-make-regexp.el" "-l" "tdtd-font-maker.el" "-f" "dtd-make-tdtd-font")))
```

Out of all the recipes, this and wanderlust were the only two recipes I noticed that utilized the nested lists for shell-command args. Seems unneccessarily complex to me.


> rust-racer

I agree that most people would want to delegate the binary management to their system package manager. That said, mu/mu4e seems to me easier to manage through straight.el alone: building mu through straight keeps the mu4e/mu versions in sync, whereas you can end up out of sync if you're relying on two package managers. I haven't used racer, so I can't say whether or not the elisp is as tightly coupled to the binary. In either case, it's easy enough to work around if someone prefers to install the binary via the recipe.

**Comment by raxod502** [comment]
Yeah, that all sounds correct!

> Gists as Git repositories

One thing that could be fancy is a `gist` host, e.g.

```elisp
(straight-use-package '(mdfind-dired :host gist :repo "900452" :local-repo "mdfind-dired"))
```

Maybe that's a little too cute, though.

> building mu through straight keeps the mu4e/mu versions in sync

That does make sense. In some cases the binary receives active development in coordination with the Elisp package, whereas in other cases the binary is totally independent. It's nice that we have both ways to handle it now.

progfolio:
> One thing that could be fancy is a gist host, e.g.

That's a clever idea to keep in mind if/when the need arises.


I've updated the workarounds list to account for #543 landing and the manual `:repo` overrides for the gist repos.

progfolio:
More good news: https://github.com/dimitri/el-get/issues/2789 has been addressed. I haven't done an exhaustive test, but the recipes with inaccessible repos should be good now, too.

Issue #556, opened 2020-08-11
if-let backport behavior limited


progfolio:
## What's wrong
The backport for `if-let` does not match the behavior of `if-let`.
It only allows and checks a single variable binding and does not evaluate multiple forms for its `else` clause.

PR #557, opened 2020-08-11
Fix if-let, when-let macro backports

progfolio:
Allow multiple bindings in variable spec.
Declare indentation.
if-let: allow multiple else forms.

Define macros with proper namespace and alias when needed.
This will make it easier to debug the macros without overriding
if/when-let when developing on Emacs versions that already have those macros
bound.

fixes: https://github.com/raxod502/straight.el/issues/556


progfolio:
#535 and #549 pass all CI test with this fix rolled in.

Issue #235, opened 2018-02-15
Don't install packages when :if condition fails in use-package


Luis-Henriquez-Perez:
I noticed when using the `use-package` `:if` keyword to conditionally load a package and making sure the condition results as `nil` (ie. `(use-package 'ivy :if  nil)`) the package was still installed. I could tell because the autoloaded commands were available. 
With `straight-use-package-by-default` set to `t` I run `(macroexpand '(use-package ivy))` and get:

    (progn
      (straight-use-package 'ivy)
      (defvar use-package--warning138
        (function
         (lambda (keyword err)
           (let ((msg (format "%s/%s: %s"
                              'ivy
                              keyword
                              (error-message-string err))))
             (display-warning 'use-package msg :error)))))
      (condition-case-unless-debug err
          nil
        (error (funcall use-package--warning138 :catch err))))

With this sexp calling `use-package` will always install `ivy` regardless of whether you use the `:if` or `:disable` keyword. It also seems that this may slow down startup because all packages are installed with straight immediately upon startup. Only the configuration and settings added by use-package are deferred but not the package itself. Ideally, `straight-use-package` should be called just before `use-package` tries to configure the package (maybe in between `:init` and `:config`).

I've been thinking about these three ways of fixing it:
1)  create a wrapper around use-package - would involve duplicating a lot of the work use-package does
2) create a wrapper around a function that use-package uses - possibly best solution?
3) change the order of straight in use-package-keywords - risky

xuchunyang:
> I noticed when using the use-package :if keyword to conditionally load a package and making sure the condition results as nil (ie. (use-package 'ivy :if nil)) the package was still installed.

Maybe it is desired?

`:straight` servers the same purpose as use-package's built-in `:ensure` and it does the same thing, i.e., `(use-package foo :if nil :ensure t)` will install `foo`. I also find it is strange and reported last year in https://github.com/jwiegley/use-package/issues/415.

Luis-Henriquez-Perez:
Oh ok. I can see how for some crucial packages like helm or ivy that would be desirable.  However, is it possible to defer the package installation until needed? For instance, I wouldn't want `(straight-use-package 'clojure-mode)` to be called on startup because I might never use clojure in my emacs session.

Perhaps manually specifying the load-path would do it?

`(use-package clojure-mode :load-path "~/.emacs.d/straight/build/clojure-mode")`

raxod502:
[...]

PR #558, opened 2020-08-14
use-package integration: allow backquoted/quoted recipe

progfolio:
The `:straight` keyword now accepts the same arguments
`straight--use-package` does. This allows one to programmatically
specify recipes.

For example:

```emacs-lisp
    (defvar my/github-username "me")

    (use-package example
      :straight `(:repo ,(concat my/github-username "/repo")))
```

Expands to:

```emacs-lisp
    (progn
      (straight-use-package
       '(example :repo "me/repo"))
      (require 'example nil nil))
```


progfolio:
Not familiar with the older syntax mentioned here:

> Previously, straight.el used a different syntax for its use-package integration. For backwards compatibility, you can use this syntax instead by customizing straight-use-package-version.

But I don't think I've broken anything.

progfolio:
```emacs-lisp
(use-package test
  :straight t)
```
  
Expands to:

```emacs-lisp
(progn
  (straight-use-package 'test)
  (require 'test nil nil))
```

Is this incorrect? What should it expand to?

progfolio:
> Would it not make more semantic sense if the syntax were to use , instead of `? The former means to evaluate, the latter means don't evaluate.

I am familiar with backquotes, but I don't understand where/how you think they should be used. An example would help.

progfolio:
> breaks several cases

I'm not sure which cases it breaks, but you are obviously more familiar with how this integration is supposed to work. AFAICT, The expansion matches for all the cases here:

```
(use-package el-patch
  :straight t)

macroexpands (essentially) to:

(straight-use-package 'el-patch)

And this:

(use-package el-patch
  :straight (:host github :repo "raxod502/el-patch"
             :branch "develop"))

becomes:

(straight-use-package
 '(el-patch :host github :repo "raxod502/el-patch"
            :branch "develop"))

If the feature you are requiring with use-package is different from the package name, you can provide a full recipe:

(use-package tex-site
  :straight (auctex :host github
                    :repo "emacsmirror/auctex"
                    :files (:defaults (:exclude "*.el.in"))))

And you may also provide just the package name:

(use-package tex-site
  :straight auctex)

If you don't provide :straight, then by default nothing happens. You may customize straight-use-package-by-default to make it so that :straight t is assumed unless you explicitly override it with :straight nil.
```

progfolio:
> Why not just check at the top level whether it's quasiquoted, and eval the form before running the existing code?

That was my original implementation, but I think this is a better approach. For one it is more transparent to the end user when the form is expanded. Secondly, I believe evaling when handling the `:straight` keyword may complicate things down the line for features like this: https://github.com/raxod502/straight.el/issues/235

progfolio:
The current implementation works for the undocumented, multiple argument case as well.
Previously, in this case the forms would appear in reverse order in the expanded body.
Though I think this is a rare corner case, I've made it so they expand in declared order in.

Seems to work for me for all of the following cases:

    (let ((tests '(;; nil
                   ((use-package test :straight nil) . (require 'test nil nil))
                   ;; t
                   ((use-package test :straight t) . (progn
                                                       (straight-use-package 'test)
                                                       (require 'test nil nil)))
                   ;; implicit quote
                   ((use-package test :straight (:repo "user/test"))
                    . (progn
                        (straight-use-package '(test :repo "user/test"))
                        (require 'test nil nil)))
                   ;; explicit quote
                   ((use-package test :straight '(:repo "user/test"))
                    . (progn
                        (straight-use-package '(test :repo "user/test"))
                        (require 'test nil nil)))
                   ;; backquote
                   ((use-package test :straight `(:repo ,(concat "user" "/test")))
                    . (progn
                        (straight-use-package `(test :repo ,(concat "user" "/test")))
                        (require 'test nil nil)))
                   ;; package name different from repo name
                   ((use-package provided :straight (package :repo "user/package"))
                    . (progn
                        (straight-use-package '(package  :repo "user/package"))
                        (require 'provided nil nil)))
                   ;; (undocumented) multiple args to :straight keyword
                   ((use-package test
                      :straight
                      `(foo :repo ,(identity "user/package"))
                      bar
                      'baz)
                    . (progn
                        (straight-use-package `(foo :repo ,(identity "user/package")))
                        (straight-use-package 'bar)
                        (straight-use-package 'baz)
                        (require 'test nil nil))))))
      (mapcar (lambda (test)
                (equal (macroexpand (car test)) (cdr test)))
              tests))

One thing to note about the undocumented, multiple argument case is that you can pass contradictory arguments.
e.g.:

    (use-package test
    :straight nil t)

expands to:

    (progn
      (straight-use-package 'test)
      (require 'test nil nil))

And:

    (use-package test
    :straight t nil)

expands to:

    (require 'test nil nil)

The last argument "wins" in these cases. Not sure if that's desirable or not, but I just wanted to note it.

straight.el
```diff
@@ -5803,19 +5803,25 @@ NAME, KEYWORD, ARGS, REST, and STATE are explained by the
...
-                    ;; `use-package-defaults' currently operates on
-                    ;; the post-normalization values, rather than the
-                    ;; pre-normalization ones.
-                    ',(if (eq arg t)
-                          name arg))
-                  body))
-          args)
-    body))
+    (straight--remq rest '(:ensure))
+    (setq args (delq 'quote (nreverse args))))
```
raxod502, line 5807:
```suggestion
    (setq args (delq 'quote (reverse args))))
```

Probably not needed, but just in case `use-package` doesn't make a copy of the list, this could introduce self-modifying code which is never fun.

straight.el
```diff
@@ -5803,19 +5803,25 @@ NAME, KEYWORD, ARGS, REST, and STATE are explained by the
...
+  (let ((body (use-package-process-keywords name rest state))
+        backquoted
+        forms)
```
raxod502, line 5810:
```suggestion
        (backquoted nil)
        (forms nil))
```

Just for consistency.

progfolio:
> a bug in the normalizer

We'll tackle that in a separate pull.

> Just add a comment in the handler

Addressed in latest commit.

> Thank you for your work here; it's great and I really appreciate your patience.

Thank you for straight.el! It's an indispensable tool for me.

jimeh:
Just in case someone else runs into the same issue, the changes in this PR caused my config to break due to how I was passing in custom package names via the `:straight` property.

Details are [here](https://github.com/jimeh/.emacs.d/commit/1c81570ef7cf6b932b888e08467f41308a23039a), TL;DR below:

```elisp
;; bad (no longer works)
(use-package helm-global-bindings
  :straight (helm))
```

```elisp
;; good (worked before, still works)
(use-package helm-global-bindings
  :straight helm)
```

progfolio:
@jimeh Apologies for the breakage. I've pushed a fix in #566 that should allow either case if you don't mind testing it out.

PR #564, opened 2020-08-26
Add straight-vc-git-post-clone-hook for post-clone git config

aspiers:
Add a hook so that users can automate custom configuration of elisp git repositories after they have been cloned.  For example, the `user.email` git config variable could be set on clone, to make upstream contributions more convenient for developers who use different email addresses for different repositories.

See https://github.com/aspiers/emacs/blob/3b2ec1b027065d360880984186ccd1ae419ecd4e/.emacs.d/init.d/as-package-loading.el#L133-L143 for an example.

----

#

### DIFF
--- a/README.md
+++ b/README.md
@@ -1781,7 +1781,32 @@ since otherwise this cache file may grow quite large over time.
 
 #### Hooks run by `straight-use-package`
 
-Currently, `straight-use-package` supports two hooks:
+Currently, `straight-use-package` supports three hooks:
+
+* `straight-vc-git-post-clone-hook`: The functions in this hook are
+  run just after cloning a git repository.  This allows users to
+  automate custom configuration of Elisp Git repositories after they
+  have been cloned.  For example, the `user.email` `git-config`
+  variable could be set on clone, to make upstream contributions more
+  convenient for developers who use different email addresses for
+  different repositories.
+
+  Each hook function is passed the following [keyword arguments]:
+
+    - `:repo-dir` - the local directory to which the repository was
+      cloned
+    - `:remote` - the name of the remote from which the repository was
+      cloned
+    - `:url` - the URL from which the repository was cloned
+    - `:branch` - the branch as specified by the recipe, if any,
+      otherwise `nil`
+    - `:depth` - the clone depth as specified by the recipe or
+      `straight-vc-git-default-clone-depth`
+    - `:commit` - the specific commit which was requested via the
+      lockfile, if any, otherwise `nil`
+
+  Since keyword arguments are used, each function should be defined
+  via `cl-defun`, and `&key` used at the front of the argument list.
 
 * `straight-use-package-prepare-functions`: The functions in this hook
   are run just before a package would be built, even if the package
@@ -3437,6 +3462,7 @@ savings on network bandwidth and disk space.
 [hydra-wiki-straight-entry]: https://github.com/abo-abo/hydra/wiki/straight.el
 [hydra]: https://github.com/abo-abo/hydra
 [issues]: https://github.com/raxod502/straight.el/issues
+[keyword arguments]: https://www.emacswiki.org/emacs/KeywordArguments
 [magit]: https://magit.vc/
 [markdown-toc]: https://github.com/jonschlinkert/markdown-toc
 [melpa-recipe-format]: https://github.com/melpa/melpa#recipe-format
--- a/straight.el
+++ b/straight.el
@@ -258,6 +258,27 @@ computes the fork as \"githubUser/fork\"."
                            (const :tag "bitbucket" bitbucket))
                 :value-type (string :tag "username")))
 
+(defcustom straight-vc-git-post-clone-hook nil
+  "Functions called after straight.el clones a git repository.
+
+Each hook function is passed the following keyword arguments:
+
+  - `:repo-dir' - the local directory to which the repository was cloned
+  - `:remote' - the name of the remote from which the repository was cloned
+  - `:url' - the URL from which the repository was cloned
+  - `:branch' - the branch as specified by the recipe, if any,
+    otherwise nil
+  - `:depth' - the clone depth as specified by the recipe or
+    `straight-vc-git-default-clone-depth'
+  - `:commit' - the specific commit which was requested via the
+    lockfile, if any, otherwise nil
+
+Since keyword arguments are used, each function should be defined
+via `cl-defun', with `&key' used at the front of the argument
+list, and `&allow-other-keys' at the end to ensure forwards
+compatibility."
+  :type 'hook)
+
 ;;;; Utility functions
 ;;;;; Association lists
 
@@ -2357,10 +2378,13 @@ specified in RECIPE instead. If that fails, signal a warning."
                 (straight--get-call
                  "git" "submodule" "update" "--init" "--recursive")))
             (setq success t))
-        ;; Make cloning an atomic operation.
-        (unless success
-          (when (file-exists-p repo-dir)
-            (delete-directory repo-dir 'recursive)))))))
+        (if (not success)
+            ;; Make cloning an atomic operation.
+            (when (file-exists-p repo-dir)
+              (delete-directory repo-dir 'recursive))
+          (run-hook-with-args 'straight-vc-git-post-clone-hook
+                              :repo-dir repo-dir :remote remote :url url
+                              :branch branch :depth depth :commit commit))))))
 
 (cl-defun straight-vc-git-normalize (recipe)
   "Using straight.el-style RECIPE, make the repository locally sane.


README.md
```diff
@@ -1715,7 +1715,24 @@ since otherwise this cache file may grow quite large over time.
...
+  following arguments:
+
+    - the local directory to which the repository was cloned
+    - the name of the remote from which the repository was cloned
+    - the URL from which the repository was cloned
+    - the branch from which the repository was cloned
+    - the clone depth
+    - the specific commit which was checked out (or `nil` for `HEAD`)
+
+  This allows users to automate custom configuration of elisp git
```
raxod502, line 1731:
```suggestion
  This allows users to automate custom configuration of Elisp Git
```

straight.el
```diff
@@ -2256,10 +2256,12 @@ specified in RECIPE instead. If that fails, signal a warning."
...
-        ;; Make cloning an atomic operation.
-        (unless success
-          (when (file-exists-p repo-dir)
-            (delete-directory repo-dir 'recursive)))))))
+        (if (not success)
+            ;; Make cloning an atomic operation.
+            (when (file-exists-p repo-dir)
+              (delete-directory repo-dir 'recursive))
+          (run-hook-with-args 'straight-vc-git-post-clone-hook
+                              repo-dir remote url branch depth commit))))))
```
raxod502, line 2264: What would you think of making these into keyword arguments? Six positional arguments are a bit difficult to keep straight, I think (although I'm sure I'm guilty of doing things like that elsewhere).

aspiers: I love that idea!  I didn't really know that was possible in Elisp, but a quick Google led me to `cl-defun` and `&key` which I see you are already using, so I will switch it to that.

aspiers:
> This looks like a great improvement! By the way, I saw in your linked configuration [that you pull in Try](https://github.com/aspiers/emacs/blob/3b2ec1b027065d360880984186ccd1ae419ecd4e/.emacs.d/init.d/as-package-loading.el#L131). By default if you `M-x straight-use-package` then the package is only "installed" for the session, and will be gone from `load-path` once you restart (although it's true that the repository will still be cloned unless you delete it); is there something inconvenient about this workflow which makes you prefer Try?

Thanks for the info!  Nope, nothing at all - simply that I discovered `try` before I switched to `straight.el` and it didn't occur to me that it renders `try` unnecessary.  I'll ditch it.

straight.el
aspiers: Done and force-pushed.

aspiers:
BTW I thought about including a link to my usage to serve as an example usage of the hook, as this could save people time if they want to write their own to do something similar.  However I didn't know whether you would prefer to minimise external references, so I left it out.  But if you want to include the link, perhaps this would be a good way of getting a stable permalink:

https://github.com/aspiers/emacs/search?q=straight-vc-git-post-clone-hook&type=Code

aspiers:
Hrmm, CI is failing with:

```
[checkdoc] straight.el
Warning (emacs): 
straight.el:267: First line is not a complete sentence
Warning (emacs): 
straight.el:277: Symbols t and nil should not appear in single quotes
Makefile:39: recipe for target 'checkdoc' failed
```

I don't know why the former is an issue since many of the other `defcustom` first lines don't form a complete sentence either.  But I'll fix both anyway.

raxod502:
> I didn't know whether you would prefer to minimise external references

External references are fine, but I think even better would be pasting that snippet into a page on the newly created `straight.el` wiki and linking to that. I think wikis are a good way of aggregating handy resources, examples, configuration etc. if they don't seem quite general enough to put directly in the documentation. That said, you can certainly do what you suggested if you feel strongly about it.

> perhaps this would be a good way of getting a stable permalink

There's also just linking to a particular commit hash for the file that contains it. This is what I typically do when I want to link to example code on GitHub.

> I don't know why the former is an issue since many of the other defcustom first lines don't form a complete sentence either

The message is misleading. It's not that the first line is supposed to be a complete sentence, it's that the first line is supposed to stand on its own and end in a period. This is a convention because there are various tools that display only the first line as a summary (e.g. ElDoc) which become much more useful if the first line stands on its own. Sometimes the length restriction (70 characters; actually, the official style guide says even shorter, I believe 67 characters) produces unnatural phrasing, but you can always clarify in subsequent lines.

See https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html (third bullet point).

straight.el
```diff
@@ -263,6 +263,24 @@ Safe mode is not guaranteed to be as performant as normal
...
+  - `:repo-dir' - the local directory to which the repository was cloned
+  - `:remote' - the name of the remote from which the repository was cloned
+  - `:url' - the URL from which the repository was cloned
+  - `:branch' - the branch from which the repository was cloned
+  - `:clone' - the clone depth
+  - `:commit' - the specific commit which was checked out (or nil for HEAD)
+
+Since keyword arguments are used, each function should be defined
+via `cl-defun', and `&key' used at the front of the argument
+list."
```
raxod502, line 281: If you're explaining `cl-defun`, you might consider mentioned also `&allow-other-keys` which allows forward-compatibility. In fact I might suggest saying `&allow-other-keys` is mandatory, so that we can add more keys in future without breaking backwards compatibility.

straight.el
```diff
@@ -263,6 +263,24 @@ Safe mode is not guaranteed to be as performant as normal
...
+successfully clones a package's git repository.
+
+Each hook function is passed the following keyword arguments:
+
+  - `:repo-dir' - the local directory to which the repository was cloned
+  - `:remote' - the name of the remote from which the repository was cloned
+  - `:url' - the URL from which the repository was cloned
+  - `:branch' - the branch from which the repository was cloned
+  - `:clone' - the clone depth
+  - `:commit' - the specific commit which was checked out (or nil for HEAD)
```
raxod502, line 277: I suppose this is only set when a lockfile is used, even though necessarily a specific commit was still checked out in the end? Might be useful to clarify.

straight.el
```diff
@@ -263,6 +263,24 @@ Safe mode is not guaranteed to be as performant as normal
...
+  "This list of functions will be called after straight.el
+successfully clones a package's git repository.
+
+Each hook function is passed the following keyword arguments:
+
+  - `:repo-dir' - the local directory to which the repository was cloned
+  - `:remote' - the name of the remote from which the repository was cloned
+  - `:url' - the URL from which the repository was cloned
+  - `:branch' - the branch from which the repository was cloned
+  - `:clone' - the clone depth
```
raxod502, line 276: Does this reflect the actual clone depth, or the value of `:depth` which may also be a symbol? And shouldn't it be `:depth` rather than `:clone`?

aspiers:
[**@raxod502**](https://github.com/raxod502) commented on [September 1, 2020 1:46 PM](https://github.com/raxod502/straight.el/pull/564#issuecomment-684826608 "2020-09-01T12:46:53Z - Replied by Github Reply Comments"):
> > I didn't know whether you would prefer to minimise external references
> 
> External references are fine, but I think even better would be pasting that snippet into a page on the newly created `straight.el` wiki and linking to that. I think wikis are a good way of aggregating handy resources, examples, configuration etc. if they don't seem quite general enough to put directly in the documentation. That said, you can certainly do what you suggested if you feel strongly about it.

I did think about that; the problem is that then I have to try to maintain my code in two separate places, and I generally hate DRY violations ;-)

> > perhaps this would be a good way of getting a stable permalink
> 
> There's also just linking to a particular commit hash for the file that contains it. This is what I typically do when I want to link to example code on GitHub.

Sure, but again that leads to out-of-date code.

> > I don't know why the former is an issue since many of the other defcustom first lines don't form a complete sentence either
> 
> The message is misleading. It's not that the first line is supposed to be a complete sentence, it's that the first line is supposed to stand on its own and end in a period. This is a convention because there are various tools that display only the first line as a summary (e.g. ElDoc) which become much more useful if the first line stands on its own. Sometimes the length restriction (70 characters; actually, the official style guide says even shorter, I believe 67 characters) produces unnatural phrasing, but you can always clarify in subsequent lines.
> 
> See [gnu.org/software/emacs/manual/html\_node/elisp/Documentation-Tips.html](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html) (third bullet point).

Ah, gotcha!  Thanks for the clarification!

straight.el
aspiers: Nice point!  I'll fix that.

raxod502:
This thread is being closed automatically by [Tidier](https://github.com/raxod502/tidier) because it is labeled with "waiting on response" and has not seen any activity for 90 days. But don't worry—if you have any information that might advance the discussion, leave a comment and I will be happy to reopen the thread :)

aspiers:
Please reopen as I am finally updating my branch to address your feedback :-)

straight.el
aspiers: > Does this reflect the actual clone depth, or the value of :depth which may also be a symbol?

The latter.  Updated to clarify.

> And shouldn't it be :depth rather than :clone?

Yes it should, good catch!  Fixed.

straight.el
aspiers: Correct (I think) - fixed.

aspiers:
I've now addressed your feedback, rebased, and force-pushed to my branch.  However it's not showing here because the PR is closed.  Hopefully it will update if you reopen.

progfolio:
@aspiers: I can reopen this for you, but github blocks that since there was a foce push after it was closed.
In order to reopen, can you:
- force push back to last commit before close

Then I'll reopen and you should be able to force push to the current commit.

aspiers:
Done.

aspiers:
BTW I suspect two force pushes after reopening would also work.

progfolio:
> BTW I suspect two force pushes after reopening would also work.

I'm not too familiar with github's UI in this scenario, but it looks like I'm still blocked:

blocked

I was going off of what I read here: **https://github.com/isaacs/github/issues/361**
but it doesn't look like it's done us any favors

progfolio:
I think I see now:

> We're blocking the pull request reopen if the current head isn't a descendant of the stored head sha (which is what the head was when the pull request was closed). We are not allowing the reopen in that case, because there is no good way to tell what changes have happened while a pull request was closed and the head branch has changed.

So we need to get your pull to the sha just before close, then I should be able to reopen.

aspiers:
> So we need to get your pull to the sha just before close, then I should be able to reopen.

I'm pretty sure that's what I already did :-(

aspiers:
I'm now trawling through the GitHub API to find the exact sha prior to close ...

From https://api.github.com/repos/raxod502/straight.el/issues/564/events, it appears to be this event:

https://api.github.com/repos/raxod502/straight.el/issues/events/3707368296

Next I think I have to look that node up via GraphQL ...

aspiers:
Hmm, based on my reflog I initially thought it was 23809c0055b9 but maybe it was actually ee29a1e535fa.

progfolio:
Sorry this is such a hassle. We can always open a new pull and just link to this one, too, if that works better for you.

aspiers:
Well, I'm using it as an excuse to practise my GraphQL ;-)  As apparently GitHub switched their API from REST to GraphQL between v3 and v4 ...

aspiers:
OK, I got it!  Check this out:

```graphql
{
  repository(name: "straight.el", owner: "raxod502") {
    pullRequest(number: 564) {
      timelineItems(last: 10, itemTypes: HEAD_REF_FORCE_PUSHED_EVENT) {
        nodes {
          ... on HeadRefForcePushedEvent {
            beforeCommit {
              url
            }
            afterCommit {
              url
            }
            createdAt
          }
        }
      }
    }
  }
}
```

If you feed that into https://docs.github.com/en/free-pro-team@latest/graphql/overview/explorer, you get:

```json
{
  "data": {
    "repository": {
      "pullRequest": {
        "timelineItems": {
          "nodes": [
            {
              "beforeCommit": {
                "url": "https://github.com/raxod502/straight.el/commit/92a7cb035660c404fd6e704ce6f330c09c4c2a08"
              },
              "afterCommit": {
                "url": "https://github.com/raxod502/straight.el/commit/fae0871e878d3cbc34a93c5f3c7faa300093df1f"
              },
              "createdAt": "2020-08-30T14:12:49Z"
            },
            {
              "beforeCommit": {
                "url": "https://github.com/raxod502/straight.el/commit/fae0871e878d3cbc34a93c5f3c7faa300093df1f"
              },
              "afterCommit": {
                "url": "https://github.com/raxod502/straight.el/commit/ee29a1e535fa78002b08d82f98c0369330339945"
              },
              "createdAt": "2020-08-30T14:50:25Z"
            }
          ]
        }
      }
    }
  }
}
```

This new GraphQL API is pretty damn cool when you figure it out ;-)

So it was indeed ee29a1e, which corresponds with the entry I saw in my reflog for the remote branch.  I've now force-pushed that - does it work?!

progfolio:
Good to go! Should be able to push back your other commits now.

aspiers:
OK, rebased against `develop` and force-pushed.

straight.el
```diff
@@ -258,6 +258,27 @@ computes the fork as \"githubUser/fork\"."
                            (const :tag "bitbucket" bitbucket))
                 :value-type (string :tag "username")))
 
+(defcustom straight-vc-git-post-clone-hook nil
+  "This list of functions will be called after straight.el
```
progfolio, line 262: checkdoc:
> straight.el:262: First line is not a complete sentence

straight.el
```diff
@@ -258,6 +258,27 @@ computes the fork as \"githubUser/fork\"."
...
+Each hook function is passed the following keyword arguments:
+
+  - `:repo-dir' - the local directory to which the repository was cloned
+  - `:remote' - the name of the remote from which the repository was cloned
+  - `:url' - the URL from which the repository was cloned
+  - `:branch' - the branch from which the repository was cloned
+  - `:depth' - the clone depth as specified by the recipe or
+    `straight-vc-git-default-clone-depth'
+  - `:commit' - the specific commit which was requested via the lockfile,
+    if any, otherwise `nil'
```
progfolio, line 274: checkdoc:
> straight.el:274: Symbols t and nil should not appear in single quotes

progfolio (changes requested):

straight.el
aspiers: Fixed.

straight.el
aspiers: Fixed.

progfolio (approved):

progfolio:
With the following test case I get `nil` for both the `:branch` and `:commit` in the hook function:


```emacs-lisp
(straight-bug-report
  :user-dir "straight-vc-git-post-clone-hook.straight"
  :pre-bootstrap 
  (setq straight-repository-user "aspiers" straight-repository-branch "post-git-clone-hook")
  :post-bootstrap 
  (cl-defun +straight-post-clone-test
      (&key repo-dir remote url branch depth commit &allow-other-keys)
    (message "repo-dir: %s\nremote: %s\nurl: %s\nbranch: %s\ndepth: %s\ncommit: %s" repo-dir remote url branch depth commit))
  (add-hook 'straight-vc-git-post-clone-hook #'+straight-post-clone-test)
  (straight-use-package 'marginalia))
```
</details>

- Test run at: `2020-12-16 15:16:25`
- system-type: `gnu/linux`
- straight-version: `prerelease (HEAD -> develop, origin/develop) 742bc66 2020-12-05`
- emacs-version: `GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.23, cairo version 1.17.4)
 of 2020-12-05`

```emacs-lisp
Bootstrapping straight.el...
Looking for emacsmirror-mirror recipe → Cloning el-get...
Looking for emacsmirror-mirror recipe → Cloning el-get...done
Building straight...
Building straight...done

Test run with version: prerelease (HEAD -> post-git-clone-hook, origin/post-git-clone-hook) 3628d76 2020-12-16
Cloning marginalia...

repo-dir: /tmp/straight-vc-git-post-clone-hook.straight/straight/repos/marginalia/
remote: origin
url: https://github.com/minad/marginalia.git
branch: nil
depth: full
commit: nil
Cloning marginalia...done
Building marginalia...
Building marginalia...done

```
</details>

Though, manually checking the repo shows we are on the correct branch and commit ("main", 56ad689)
Possibly related?: https://github.com/raxod502/straight.el/pull/630#issuecomment-729009892

aspiers:
[**@progfolio**](https://github.com/progfolio) commented on [December 16, 2020 8:23 PM](https://github.com/raxod502/straight.el/pull/564#issuecomment-746956624 "2020-12-16T20:23:44Z - Replied by Github Reply Comments"):
> With the following test case I get `nil` for both the `:branch` and `:commit` in the hook function:

Yes, that's expected because `:branch` isn't in the recipe and presumably you haven't got a lockfile locking the repo to a particular commit.

progfolio:
I see. Your documentation on the `:commit` keyword indeed mentions that. One small nit is that the docstring only mentions:

> the branch from which the repository was cloned

Though this is only true if it was specified in the recipe. We do compute the default branch just before the hook is run:
https://github.com/raxod502/straight.el/blob/949a0a70faebb289d430af709aa69afd684bd1de/straight.el#L2329-L2331

Do you think that information is worth holding onto in cases where no `:branch` was explicitly set?
If not, maybe the docstring could mention that it's only looking for the `:branch` declared in the recipe?

aspiers:
Good catch, I've updated the docstring and README to clarify this.  I also rebased on latest `develop`.

progfolio (approved):

progfolio:
Looks good to me. Thank you.

PR #566, opened 2020-08-28
use-package integration: normalize recipe list with only name

progfolio:
Normalize recipe lists which only have a name. e.g.
(use-package :straight (package))

see: https://github.com/raxod502/straight.el/pull/558#issuecomment-682262415

This change also provides a small optimization to avoid repeatedly checking the car of the arg.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -5782,17 +5782,22 @@ documentation."
        ((symbolp arg) (push arg parsed-args))
        ((not (listp arg))
         (use-package-error ":straight wants a symbol or list"))
-       ((keywordp (car arg))
-        ;; recipe without package name
-        (push (cons name-symbol arg) parsed-args))
-       ((cl-some #'keywordp arg)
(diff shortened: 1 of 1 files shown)


jimeh:
@progfolio Thanks, this does fix the issue I ran into after #558 was merged :)

I actually just assumed I was doing it wrong in my config though, as the readme doesn't actually contain an example with the exact syntax I was using.

progfolio:
Thank you for testing it out. We'll see what @raxod502 says about that particular syntax.

raxod502:
Arguably, it would be more consistent for that syntax to translate into `(straight-use-package '(foo))`, would it not?

progfolio:
> it would be more consistent for that syntax to translate into `(straight-use-package '(foo))`

Addressed in latest commit.

raxod502 (approved):
Thank you, and I'm sorry for the delay.

PR #569, opened 2020-09-01
Add no-native-compile packages to deferred-compilation-black-list

flatwhatson:
Deferred compilation automatically compiles any `.elc` that is loaded, and is now enabled by default on native-comp builds of Emacs.  This patch makes the `:no-native-compile` keyword behave as expected by blacklisting the package from deferred compilation.


raxod502 (approved):
Great, thank you!

Issue #93, opened 2017-07-16
Allow specifying :source in recipe


raxod502:
This would allow you to select a particular recipe repository, without copying the recipe out.

PR #570, opened 2020-09-01
Specify :source in recipe

progfolio:
Add keyword for overriding recipe repository sources on a per-recipe
basis.

see: #93


raxod502 (approved):
Awesome! Thank you!

Issue #541, opened 2020-07-02
Add a post-build hook


dertuxmalwieder:
For certain packages (like `vterm`), a few manual commands are required after each update. I couldn’t find it in the documentation: I would like to have an obvious way to perform commands after the build process.

Example:

    (use-package vterm
      :straight t
      :ensure t
      ;; would be new:
      :post-build-hook (vterm-module-compile))

Something like that.

(If I accidentally failed to see an already existing option for that: Sorry!)

raxod502:
Well, there is `straight-use-package-pre-build-functions` but indeed no `straight-use-package-post-build-functions`. And you are right that the most reasonable thing is not to have a hook but rather a new keyword in the recipe. Although it will be a `straight.el` keyword, not a `use-package` one like in your example code.

I think that you should take a look at https://github.com/raxod502/straight.el/pull/535, which will be about the right place for this functionality. (Not necessarily in that very pull request, but in the same part of the code.)

progfolio:
#535 makes it possible to specify a recipe like so now:

```emacs-lisp
(use-package vterm
  :straight t
  :ensure t
  :post-build (vterm-module-compile))
```

dertuxmalwieder:
Awesome! I hope that it will be merged. :-)

progfolio:
@dertuxmalwieder support for the `:build` and `:post-build` keywords is now available on the develop branch.
The following works for me to set up vterm. Let me know if you have any trouble, questions or suggestions.

<details><summary>Test Case</summary>

```emacs-lisp
(straight-bug-report
  :post-bootstrap 
  (straight-use-package
   '(vterm :post-build
           ((let
                ((vterm-always-compile-module t))
              (require 'vterm))))))
```
</details>

- Test run at: `2020-10-23 21:03:17`
- system-type: `gnu/linux`
- straight-version: `prerelease (HEAD -> develop, origin/develop) cd42db6 2020-10-23`
- emacs-version: `GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, X toolkit, cairo version 1.17.3, Xaw3d scroll bars)
 of 2020-10-14`

<details><summary>Output</summary>

```emacs-lisp
Bootstrapping straight.el...
Bootstrapping straight.el...done
Rebuilding all packages due to build cache schema change
Looking for gnu-elpa-mirror recipe → Cloning melpa...
Looking for gnu-elpa-mirror recipe → Cloning melpa...done
[...]

PR #572, opened 2020-09-04
Add straight-use-package-post-build-functions

progfolio:
Compliment to straight-use-package-pre-build-functions.

see: #541


progfolio:
As mentioned in #541, I agree that adding another keyword to the straight recipe format would be better for many use cases. However, I see no harm in having a complimentary hook. It costs nothing if it is not being used and may enable other use cases people may think of in the future. Off the top of my head I think this could be useful for collecting statistics on package builds. Better to provide a hook than force a user to advise `straight--build-package`.

raxod502 (approved):
Perfect! :sparkles:

Issue #509, opened 2020-04-19
"Process failed" error is very poor


raxod502:
Currently when a subprocess invocation fails unexpectedly, you get

```
(error "Failed to run \"git\"; see buffer *straight-process*")
```

which is pretty bad. Firstly we should show part of the output from the subprocess in the minibuffer. Secondly we should just pop to the `*straight-process*` buffer automatically, instead of telling the user to do it themselves.

progfolio:
Safe to close now that #573 has landed on develop?

raxod502:
I think so. Thanks!

PR #573, opened 2020-09-04
Fix: Better subprocess error handling

progfolio:
If subprocess error produces output, show it in minibuffer.
Switch to straight-process-buffer automatically.

See: #509

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -975,8 +975,18 @@ code of zero, and nil otherwise."
     (if (car result)
         t
       (prog1 nil
+        (when-let ((output (cdr result)))
+          ;; We're only interested in first line of output.
+          (setq output (replace-regexp-in-string "\n.*" "" output))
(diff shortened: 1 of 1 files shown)


straight.el
```diff
@@ -939,8 +939,11 @@ code of zero, and nil otherwise."
     (if (car result)
         t
       (prog1 nil
+        (when-let ((output (cdr result)))
+          (message "%s" output))
```
raxod502, line 943: Can we show only the first line, and limit it to at most like a hundred characters or perhaps something based on the frame width? It looks pretty ugly when too much stuff gets dumped into the minibuffer.

We should probably also prefix this with `Failed to run %S: `.

progfolio: Latest commit should grab first line and limit it to 100 characters.

raxod502:
Looks good to me!

Issue #397, opened 2019-07-09
What to do with non-existing dependencies


Fuco1:
```elisp
(straight-use-package 'emr)
```

fails to install because 

```
$ cd /home/matus/.emacs.d/straight/repos/
$ git clone --origin origin --no-checkout https\://github.com/emacsorphanage/clang-format.git /home/matus/.emacs.d/straight/repos/clang-format/

Cloning into '/home/matus/.emacs.d/straight/repos/clang-format'...
fatal: could not read Username for 'https://github.com': No such device or address

[Return code: 128]
```

Is there some way I could point straight to somewhere where this package can be found?  E.g. I could grab it from MELPA (it still exists there for some reason) and put it under my github account or something.

raxod502:
Just update your MELPA; it is three months out of date: https://github.com/melpa/melpa/commit/be27e728327016b819535ef8cae10020e5a07c2e

But if you wanted to provide a custom recipe, you could use `straight-recipe-overrides`.

Fuco1:
So I think this can result in some nice UX improvement, although I'm not sure how straight might predict this, maybe it can print somewhere "Blabla package not found here, maybe try to update your archives".

Because the part about updating the repositories is pretty obscure, even though it is well documented and described in the readme, it's a bit contrary to how most package managers work (i.e. against some central "index in the cloud" which is always valid).

raxod502:
Adding a new package with `straight.el` is rather like running `apt install`. You have to run `apt update` first if you want to get the updated listings.

That being said, I agree that the user experience isn't as transparent as it ought to be. I would be happy to get a pull request for this.

PR #575, opened 2020-09-05
Add straight-pull-recipe-repositories

progfolio:
Convenience command to update registered recipe repositories.
Adds better error message for missing recipe case in
straight-get-recipe.

See: #397


progfolio:
Not sure what happened here...sorry.

PR #576, opened 2020-09-05
Add straight-pull-recipe-repositories

progfolio:
Convenience command to update registered recipe repositories.
Adds better error message for missing recipe case in
straight-get-recipe.

See: #397

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -4807,12 +4807,36 @@ action, just return it)."
            ;; any messages.
            (recipe (straight-recipes-retrieve package sources)))
       (unless recipe
-        (user-error "Recipe for %S is malformed" package))
+        (user-error
+         (concat "Recipe for \"%S\" malformed or missing. "
+                 "Updating recipe repositories: %s "
+                 "with straight-pull-recipe-repositories may fix this.")
+         package sources))
       (pcase action
         (`insert (insert (format "%S" recipe)))
         (`copy (kill-new (format "%S" recipe))
                (straight--output "Copied \"%S\" to kill ring" recipe))
         (_ recipe)))))
+
+;;;;; Update recipe repositories
+(defun straight-pull-recipe-repositories (&optional sources)
+  "Update recipe repository SOURCES.
+When called with `\\[universal-argument]', prompt for SOURCES.
+If SOURCES is nil, update sources in `straight-recipe-repositories'."
+  (interactive (list (if (equal current-prefix-arg '(4))
+                         (completing-read-multiple
+                          "Recipe Repositories (empty to select all): "
+                          straight-recipe-repositories nil 'require-match)
+                       straight-recipe-repositories)))
+  (dolist (source (delete-dups
+                   (mapcar (lambda (src) (if (stringp src) (intern src) src))
+                           (or sources straight-recipe-repositories))))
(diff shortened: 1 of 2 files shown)


straight.el
```diff
@@ -4791,13 +4791,41 @@ action, just return it)."
...
+                     sources
+                     straight-recipe-repositories))))
+    (dolist (source (delete-dups sources))
+      (unless (member source straight-recipe-repositories)
+        (user-error
+         (concat "Unregistered recipe repository: %s.\n"
+                 "Register recipe source with straight-use-recipes")
+         source))
+      (straight-pull-package-and-deps (symbol-name source) 'upstream))))
+
```
raxod502, line 4828: What about making this use the same interface as `straight-get-recipe`, namely default to all recipe repositories and make a prefix argument let you select an individual one? Most users will not be offended by updating more recipe repositories than they intended, I think.

straight.el
```diff
@@ -4791,13 +4791,41 @@ action, just return it)."
            ;; any messages.
            (recipe (straight-recipes-retrieve package sources)))
       (unless recipe
-        (user-error "Recipe for %S is malformed" package))
+        (user-error
+         (concat "Recipe for \"%S\" malformed or missing\n"
+                 "Updating recipe repositories: %s\n"
+                 "with straight-pull-recipe-repositories may fix this.")
```
raxod502, line 4797: Can we keep messages to a single line if possible? The multiple line ones do not render as well in all configurations because the minibuffer window changes size, so I'd like to reserve them for when it really cannot be avoided.

straight.el
```diff
@@ -4791,13 +4791,41 @@ action, just return it)."
...
 
+;;;;; Update recipe repositories
+(defun straight-pull-recipe-repositories (&optional sources)
+  "Update recipe SOURCES.
+When called interactively, prompt for SOURCES.
+If SOURCES is nil, use `straight-recipe-repositories'."
+  (interactive)
+  (let ((sources
+         (mapcar (lambda (src) (if (stringp src) (intern src) src))
+                 (or (and (called-interactively-p 'interactive)
```
raxod502, line 4813: A more idiomatic way of doing this would be to follow the lead of `straight-get-recipe` and put the `completing-read` stuff into the `interactive` spec.

progfolio:
That should take care of everything. Let me know what you think.

raxod502 (approved):

raxod502:
After merging this, I had second thoughts about `Recipe for "xyz" malformed or missing`. Does this message really appear when a recipe is missing? I am pretty sure it is not possible to hit that case, because the user interactively selects from a list of packages whose recipes are already available in the cloned recipe repositories. This should only happen when `straight-recipes-retrieve` returns nil for a recipe that was listed by `straight-recipes-list`, which should only happen for cases of e.g. MELPA recipes that use Mercurial.

Did you mean to instead update the message that occurs when `straight--convert-recipe` fails to look up a package's recipe?

https://github.com/raxod502/straight.el/blob/f2aa672aa7f14149018ac1fb957bf003ddf4eda8/straight.el#L3075-L3078

progfolio:
Yeah, that makes a lot more sense. Good catch!

Issue #308, opened 2018-09-18
straight.el assigns :local-repo for :type built-in recipes


raxod502:
Current behavior:

```elisp
(straight--convert-recipe '(foo :type built-in))
;; => (:type built-in :package "foo" :local-repo "foo")
```

Desired behavior:

```elisp
(straight--convert-recipe '(foo :type built-in))
;; => (:type built-in :package "foo")
```

PR #577, opened 2020-09-05
Fix: Don't assign :local-repo for :type built-in recipes

progfolio:
Fixes: #308


raxod502 (approved):
This seems eminently reasonable!

PR #583, opened 2020-09-20
emasc-devel -> emacs-devel

agsdot:


raxod502 (approved):
Tragic. Thanks for the fix!

agsdot:
For sure, no problem! :D

PR #571, opened 2020-09-03
Add straight-host-usernames; Allow partial :fork repo spec.

progfolio:
Add support for default usernames per vc host.
Compute :fork repo name when given partial string or t.

See: #113

PR #585, opened 2020-09-21
Fix: compute :fork repo name

progfolio:
Call `straight-vc-git--fork-repo` (introduced in #571) with enough
information to compute fork repo in `straight--convert-recipe`.

I tried to simplify the logic for recipe inheritance in `straight--convert-recipe`, too.
Let me know what you think.


raxod502 (approved):
Ok!

PR #586, opened 2020-09-22
Fix: straight-bug-report: add missing format spec to message

progfolio:
Call to `message` was missing a formatting spec, so output of `straight-version` in testing environment was not being printed.


progfolio:
Sorry about the merge commit! I thought the github interface would've handled that properly. I'll be more careful.

raxod502:
It can be somewhat confusing. Don't worry about it; I've long since given up on a pristine commit log. It's inherently a hard problem since the log is write-only :)

PR #588, opened 2020-09-23
Ignore text properties in straight--process-get-output

mrBliss:
`straight--process-get-output` uses `buffer-substring` to copy the
output of a process, which means that the text properties are copied
too. Most of the time, this is fine, as there are no text properties.

However, when using the [`ws-butler`] package, the `ws-butler-chg`
text property is present and copied. This means that, for example,
`straight-freeze-versions` produces a `default.el` file looking like
this:

```emacs-lisp
(("alert" . #("7046393272686c7a1a9b3e7f7b1d825d2e5250a6" 0 40 (ws-butler-chg chg)))
 ("anchored-transpose" . #("43b37bbc5dc0fe4b7b1b46c5e1d98553e9b527ae" 0 40 (ws-butler-chg chg)))
 ..)
```

while it should be:

```emacs-lisp
(("alert" . "7046393272686c7a1a9b3e7f7b1d825d2e5250a6")
 ("anchored-transpose" . "43b37bbc5dc0fe4b7b1b46c5e1d98553e9b527ae")
 ..)
```

Fix this by using `buffer-substring-no-properties` instead of
`buffer-substring`.

[`ws-butler`]: https://github.com/lewang/ws-butler

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -952,8 +952,9 @@ This is nil if the process could not be run, or if
   (with-current-buffer (straight--process-get-buffer)
     (when (and straight--process-output-beginning
                straight--process-output-end)
-      (buffer-substring straight--process-output-beginning
-                        straight--process-output-end))))
+      (buffer-substring-no-properties
+       straight--process-output-beginning
+       straight--process-output-end))))
 
 (defun straight--call (program &rest args)
   "Run executable PROGRAM with given ARGS.


progfolio:
Thank you for the patch!

ws-butler provides:

    (defcustom ws-butler-global-exempt-modes
      '(markdown-mode)
      "Don't enable ws-butler in modes that inherit from these.")

This allows ignoring `special` mode buffers (desirable in most cases as these are usually read-only buffers):

    (add-to-list 'ws-butler-global-exempt-modes 'special-mode)

That said, unless @raxod502 has any objections, this seems like a reasonable change to me.

mrBliss:
> ws-butler provides:
> 
> ```
> (defcustom ws-butler-global-exempt-modes
>   '(markdown-mode)
>   "Don't enable ws-butler in modes that inherit from these.")
> ```
> 
> This allows ignoring `special` mode buffers (desirable in most cases as these are usually read-only buffers):
> 
> ```
> (add-to-list 'ws-butler-global-exempt-modes 'special-mode)
> ```

Aha! I also thought of doing the exact same thing (at some point I mentioned it even in the commit message but removed it again!) :slightly_smiling_face:. However, I think this PR offers a more robust approach: what about all the other Emacs packages that we don't know about that also add some text properties?

progfolio:
> I think this PR offers a more robust approach: what about all the other Emacs packages that we don't know about that also add some text properties?

Agreed. Just to be clear, I'm for the patch- just wanted to point to the ws-butler solution so it can be used before the patch is applied.

raxod502:
Yep, sounds like a good idea to me! Thanks for the fix.

raxod502 (approved):

mrBliss:
> > I think this PR offers a more robust approach: what about all the other Emacs packages that we don't know about that also add some text properties?
> 
> Agreed. Just to be clear, I'm for your patch- I just wanted to point to an alternative, configuration workaround that will work in the mean time.

Yeah, I realised that, although that might not have been clear from my last comment :slightly_smiling_face:.

> Yep, sounds like a good idea to me! Thanks for the fix.

You're welcome! Thanks for `straight.el`!

PR #589, opened 2020-09-24
Allow specifying protocol in package recipes

### LINKED ISSUE #92: Allow specifying :protocol in recipe
See also https://github.com/raxod502/straight.el/issues/81.

VojtechStep:
Fixes #92 

This PR allows specifying git protocol as part of a package recipe.  
I included the note about only using it for recipe overrides, because I couldn't think of another use case, correct me if I'm wrong (the only reason why I took this on is because I have all packages cloned over SSH, except for `org-drill`, which is on GitLab where I don't have an account, so I had to clone it over HTTPS).

Also, does it make sense to specify a different protocol for a fork? I haven't included this functionality because I don't use forks, but I don't think it would be hard to add it.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -1974,19 +1974,19 @@ The URLs do not necessarily need to match exactly; they just have
 to satisfy `straight-vc-git--urls-compatible-p'."
   (straight-vc-git--destructure recipe
       (local-repo upstream-repo upstream-host upstream-remote
-                  fork-repo fork-host fork-remote)
+                  fork-repo fork-host fork-remote protocol)
     (and (or (null upstream-repo)
              (straight-vc-git--ensure-remote
               local-repo
               upstream-remote
               (straight-vc-git--encode-url
-               upstream-repo upstream-host)))
+               upstream-repo upstream-host protocol)))
          (or (null fork-repo)
              (straight-vc-git--ensure-remote
               local-repo
               fork-remote
               (straight-vc-git--encode-url
-               fork-repo fork-host))))))
+               fork-repo fork-host protocol))))))
 
 ;; The following handles only merges, not rebases. See
 ;; https://github.com/raxod502/straight.el/issues/271.
@@ -2321,12 +2321,12 @@ out, signal a warning. If COMMIT is nil, check out the branch
 specified in RECIPE instead. If that fails, signal a warning."
   (straight-vc-git--destructure recipe
       (package local-repo branch remote upstream-repo upstream-host
-               upstream-remote fork-repo repo host nonrecursive depth)
+               upstream-remote fork-repo repo host protocol nonrecursive depth)
     (unless upstream-repo
       (error "No `:repo' specified for package `%s'" package))
     (let ((success nil)
           (repo-dir (straight--repos-dir local-repo))
-          (url (straight-vc-git--encode-url repo host))
+          (url (straight-vc-git--encode-url repo host protocol))
           (depth (or depth straight-vc-git-default-clone-depth)))
       (unwind-protect
           (progn
@@ -2340,7 +2340,7 @@ specified in RECIPE instead. If that fails, signal a warning."
                   (default-directory repo-dir))
               (when fork-repo
                 (let ((url (straight-vc-git--encode-url
-                            upstream-repo upstream-host)))
+                            upstream-repo upstream-host protocol)))
                   (straight--get-call "git" "remote" "add" upstream-remote url)
                   (straight--get-call "git" "fetch" upstream-remote)))
               (when commit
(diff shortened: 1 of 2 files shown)


raxod502:
> I included the note about only using it for recipe overrides, because I couldn't think of another use case, correct me if I'm wrong

Well, it would be easy to tweak your personal use case: what if org-drill was a package that was not prominent enough to be listed on MELPA? Then you'd be writing a custom recipe from scratch that would use `:protocol`.

> Also, does it make sense to specify a different protocol for a fork?

I think it does. I think you'd be able to support this just by using different variable names in your existing code, without needing to add anything new (use `upstream-protocol` and `fork-protocol` instead of `protocol` directly, and the macro will take care of the rest).

Thank you, this is a great improvement!

VojtechStep:
> Well, it would be easy to tweak your personal use case: what if org-drill was a package that was not prominent enough to be listed on MELPA? Then you'd be writing a custom recipe from scratch that would use :protocol.

My thinking was that in this case, I just wouldn't specify the protocol and allow other potential users (for example if I was running a custom repository) to keep using whatever they specify as their `straight-vs-git-default-protocol`.

I will look into the fork feature and try integrating it. I looked through some of the code and it seems easy enough.

VojtechStep:
Ok, so I guess it makes sense to specify the protocol even in non-override style recipes, for example if the remote only supports that specific protocol, so I reworded the documentation.

straight.el
```diff
@@ -2975,12 +2981,12 @@ Emacsmirror, return a MELPA-style recipe; otherwise return nil."
...
          ;; github.com/emacsmirror).
-         (cl-destructuring-bind (repo host _protocol)
+         (cl-destructuring-bind (repo host protocol)
              (straight-vc-git--decode-url url)
            (if host
                `(,package :type git :host ,host
-                          :repo ,repo)
-             `(,package :type git :repo ,repo))))))
+                          :repo ,repo :protocol ,protocol)
+             `(,package :type git :repo ,repo :protocol ,protocol))))))
```
raxod502, line 2989: Do we really want to do this? Even if the user has set a default protocol, this will override it.

straight.el
```diff
@@ -1708,7 +1708,8 @@ appropriately."
                                              "fork-" (symbol-name prop))))
                                 (setq require-fork t))
                               (let ((kw (intern (format ":%S" prop))))
-                                (if (memq prop '(repo host branch remote))
+                                (if (memq prop
+                                          '(repo host branch remote protocol))
```
raxod502, line 1712: I think this got changed by accident?

VojtechStep: If that's not included, the `straight-vc-git--destructure` macro is not able to properly destructure into `fork-protocol`, which I thought is what's expected.

Compare the following expansions:
Without:
```elisp
(let
      ((protocol
        (plist-get --recipe-- ':protocol))
       (fork-protocol
        (plist-get --recipe-- ':protocol))
       (upstream-protocol
        (plist-get --recipe-- ':protocol)))
    (message "%s %s %s" protocol fork-protocol upstream-protocol))
```
With:
```elisp
(let
      ((protocol
        (if-let
            ((fork
              (plist-get --recipe-- :fork)))
            (plist-get fork ':protocol)
          (plist-get --recipe-- ':protocol)))
       (fork-protocol
        (if-let
            ((fork
              (plist-get --recipe-- :fork)))
            (plist-get fork ':protocol)
          nil))
       (upstream-protocol
        (plist-get --recipe-- ':protocol)))
    (message "%s %s %s" protocol fork-protocol upstream-protocol))
```

Maybe this variable should be extracted into a separate constant, because it basically specifies which properties can be overridden by the `:fork` property (AFAICT).

VojtechStep: What should happen if the fork does not specify a protocol? It could be either:

1. `nil` (current implementation)
2. `straight-vc-git-default-protocol`
3. value of the upstream protocol

straight.el
VojtechStep: Fair point, reverted

straight.el
raxod502: > If that's not included

I was simply bad at reading, sorry about that! I didn't see that you added `protocol` to the list, and thought it was just a reformatting operation.

> What should happen if the fork does not specify a protocol?

I think (3) will surprise people. What's the difference between (1) and (2)? Don't both result in `straight-vc-git-default-protocol` being used? If that's the case, then between (1) and (2) I'd suggest doing the same as what happens when the upstream does not specify a protocol.

VojtechStep: As it's used now, there should be no difference between (1) and (2). (1) simply defers the choice to `straight-vc-git--encode-url`, which is the only place where the protocol is used.

If the upstream does not specify a protocol, it leaves it as `nil`, so same behavior as (1), which is currently implemented in this PR.

raxod502: Sounds like you're doing the right thing then :)

PR #590, opened 2020-09-25
straight-bug-report: Improve report form formatting

progfolio:
`pp` and `pp-to-string` do a fair job of pretty printing the report
form, but they do not make optimal use of vertical space and tend to
produce wide forms. Multiple KEY VAL pairs are printed on the same line.
`straight-bug-report--report-form` returns a string which makes better
use of vertical space.

Before (with `pp` or `pp-to-string`):

```emacs-lisp
(straight-bug-report :preserve t :raw nil :user-dir "pretty-print-test" :pre-bootstrap
                     (message "hi")
                     (message "two")
                     :post-bootstrap
                     (message "bye"))

```

After (with `straight-bug-report--report-form`):

```emacs-lisp
(straight-bug-report
  :preserve t
  :raw nil
  :user-dir "pretty-print-test"
  :pre-bootstrap 
  (message "hi")
  (message "two")
  :post-bootstrap 
  (message "bye"))
```

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -6263,16 +6263,18 @@ If PREAMBLE is non-nil, it is inserted after the instructions."
 
 (defun straight-bug-report--report-form (form)
   "Convert elisp FORM into formatted string."
-  (let ((string (mapconcat
-                 (lambda (el)
-                   (format (concat (when (and el (listp el)) "\n")
-                                   "%S"
-                                   (unless (keywordp el) "\n"))
-                           el))
-                 form " ")))
+  (let* ((last (car (last form)))
+         (string (mapconcat
+                  (lambda (el)
+                    (format (concat (when (and el (listp el)) "\n")
+                                    "%S"
+                                    (unless (or (keywordp el)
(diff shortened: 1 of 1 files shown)


straight.el
```diff
@@ -6261,6 +6261,27 @@ If PREAMBLE is non-nil, it is inserted after the instructions."
...
+                                    (unless (or (keywordp el)
+                                                (eq last el)) "\n"))
+                            el))
+                  form " ")))
+    (with-temp-buffer
+      ;; Trim last newline to prevent dangling paren
+      (insert "(" string ")")
+      (goto-char (point-min))
+      ;; Remove empty lines
+      (flush-lines "\\(?:^[[:space:]]*$\\)")
```
raxod502, line 6280: I learned a new function today! Thanks.
By the way, your approach here is similar to what we do for writing lockfiles: https://github.com/raxod502/straight.el/blob/5a3034767b61ced853d0b92159fd506ceb5f6882/straight.el#L5687-L5706

Issue #592, opened 2020-09-26
Non-github/gitlab repo recipe with fork:  error("No straight-host-usernames entry for host `%S'" nil)


yantar92:
## What's wrong

After recent update (I am on dev branch) I am getting error when trying to load my custom fork of org-mode repo:

```
  signal(error ("No straight-host-usernames entry for host ‘nil’"))
  error("No straight-host-usernames entry for host `%S'" nil)
  straight-vc--host-username(nil)
  straight-vc-git--fork-repo((:host nil :repo "https://code.orgmode.org/bzg/org-mode.git" :fork "/home/yantar92/Git/org-mode/"))
  straight-vc-git--fork-repo((:type git :repo "https://code.orgmode.org/bzg/org-mode.git" :local-repo "org" :fork (:repo "/home/yantar92/Git/org-mode/")))
  straight--convert-recipe((org :type git :repo "https://code.orgmode.org/bzg/org-mode.git" :local-repo "org" :files (:defaults "contrib/lisp/*.el") :fork (:repo "/home/yantar92/Git/org-mode/")) nil)
  straight-use-package((org :type git :repo "https://code.orgmode.org/bzg/org-mode.git" :local-repo "org" :files (:defaults "contrib/lisp/*.el") :fork (:repo "/home/yantar92/Git/org-mode/")))
```

## Directions to reproduce

Try to load the following code (non-existing file paths should not matter):

```elisp
(straight-use-package '(org
			:type git :repo "https://code.orgmode.org/bzg/org-mode.git"
                        :local-repo "org"
			:files (:defaults "contrib/lisp/*.el")
			:fork (:repo "/home/yantar92/Git/org-mode/")))
```

### Version information

* Emacs version: GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.20, cairo version 1.16.0) of 2020-08-15
* Operating system: Linux yantar92-laptop 5.8.1-gentoo

raxod502:
Perhaps you have thoughts @progfolio, I think this is caused by https://github.com/raxod502/straight.el/blob/6cc9fcfd4b009c38bb6753ebd78fed5d377dced2/straight.el#L1616-L1625 transforming the `(:repo "/home/yantar92/Git/org-mode/")` into just `"/home/yantar92/Git/org-mode/"`.

progfolio:
@yantar92: Apologies for the breakage and thank you for the detailed report.
I've pushed a fix at:

[...]

PR #595, opened 2020-09-26
straight-bug-report: Allow introspection

progfolio:
Provide a local variable binding, `straight-bug-report-args`, in the subprocess program.
This allows code in the :pre/:post-bootstrap program to make use of
other arguments passed to `straight-bug-report`. e.g.

```emacs-lisp
(straight-bug-report
  :user-dir "test"
  :post-bootstrap
  (message "user-dir: %S" (plist-get straight-bug-report-args :user-dir)))
```


raxod502:
Interesting! What use cases do you have in mind, out of curiosity?

progfolio:
For the test case in #592 I used it to set the default dir to the testing dir so I could pull in a local clone of Org. Another case that comes to mind is  conditioning parts of the test based on whether or not it's an interactive test.

Issue #599, opened 2020-10-04
Emacs freeze when using `watchexec` to improve startup


babygau:
## What's wrong

`emacs` freeze when scrolling down after use this snippet

## Directions to reproduce

Add the following snippet in `init.el` before `straight.el` bootstrap

```elisp
 (if (and (executable-find "watchexec")
          (executable-find "python3"))
     (setq straight-check-for-modifications '(watch-files find-when-checking))
   (setq straight-check-for-modifications
         '(find-at-startup find-when-checking)))
```

Open Emacs, scrolling down until screen refresh, Emacs will freeze and have to kill it.

### Version information

* Emacs version: 28.0.5
* Operating system: macOS 10.15.6

progfolio:
Which case is the freeze occurring for? It would be helpful to know what the value of `straight-check-for-modifications` is when you experience the freeze.

babygau:
@progfolio It's my bad. I couldn't reproduce the problem with new updated `gccemacs`. So I think that I used a "bad-git-commit" compiled gccemacs at the time.

babygau:
@progfolio, now it happened again. Same snippet:
```
 (if (and (executable-find "watchexec")
          (executable-find "python3"))
     (setq straight-check-for-modifications '(watch-files find-when-checking))
   (setq straight-check-for-modifications
         '(find-at-startup find-when-checking)))
```

FYI, I'm using `evil-mode`, so moving around, make some random edit then `Emacs` freeze. I opened new terminal, accessed freezing emacs with `kill -USR2 pid`, and it showed this bug

<img width="552" alt="Screen Shot 2020-10-05 at 3 48 40 pm" src="https://user-images.githubusercontent.com/1514823/95041373-3f775600-0722-11eb-88f4-716561625fb9.png">

If I comment out the above snippet, the issue's gone.

Please let me know if you need any further information

progfolio:
Could you try evaluating this in your `*scratch*` buffer?:

```emacs-lisp
(straight-bug-report
  :pre-bootstrap
  (if (and (executable-find "watchexec")
           (executable-find "python3"))
      (setq straight-check-for-modifications '(watch-files find-when-checking))
    (setq straight-check-for-modifications
          '(find-at-startup find-when-checking)))
  :post-bootstrap
  (message "straight-check-for-modifications: %S" straight-check-for-modifications))
```

It will bootstrap straight in a clean environment and give us some more information.
[...]

PR #600, opened 2020-10-05
straight-bug-report: programmaticly set executable default

progfolio:
Rather than assume a default of "emacs", we can use the path of the
currently running Emacs session. This should prevent the user from
having to manually specify an :executable in most cases.

See: #599 user is on MacOS and "emacs" is was not the appropriate executable.


progfolio:
Blocked by #601

Issue #478, opened 2020-02-12
Add unit and integration tests


raxod502:
Recently we have had several bad regressions, see e.g. https://github.com/raxod502/straight.el/commit/370059b2769a8100a03339a7ef245025ba55f6ea and https://github.com/raxod502/straight.el/commit/f54802514b6d3ba66269c4e09640d2de7a7dade2. I previously discarded the idea of tests as not worth it (see #69 and #70), but it's been a few years and `straight.el` is now used much more widely, including in particular by [Doom Emacs](https://github.com/hlissner/doom-emacs), which means that regressions affect many more users. Therefore, the current situation needs to change.

Unit tests will be helpful for sure, to test complicated pure functions like recipe conversion, but we also need a way of running a suite of integration tests without it taking too long, perhaps by setting up a fast sandbox or stubbing out network requests. Finally, some true end-to-end integration tests are needed, to ensure that changes to innocuous-seeming parts of the code don't break one or another popular package.

PR #603, opened 2020-10-07
Add unit tests

progfolio:
Added a few unit tests using buttercup. They can be run with `make test` from the straight repo directory. They do assume buttercup is installed, so we'd have to arrange for that if we wanted to run these on CI, but that should be easy to do with straight itself.

image

Related: #478 #69 

Let me know what you think about this setup in general. If you like it, I'll try and get us covered on all the unit tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,9 @@ EMACS ?= emacs
 
 SHELL := bash
 
+#Point to path of your local emacs-buttercup install
+BUTTERCUP = ../emacs-buttercup/
+
 # The order is important for compilation.
 for_compile := straight.el bootstrap.el install.el straight-x.el	\
 	benchmark/straight-bench.el
@@ -48,7 +51,6 @@ checkdoc: ## Check docstring style
 .PHONY: longlines
 longlines: ## Check for long lines
 	@scripts/check-line-length.bash
-
 .PHONY: checkindent
 checkindent: ## Ensure that indentation is correct
 	@tmpdir="$$(mktemp -d)"; for file in $(for_checkindent); do \
@@ -94,3 +96,8 @@ bench: ## Run benchmarking script against package.el
 .PHONY: docker
 docker: ## Start a Docker shell; e.g. make docker VERSION=25.3
 	@scripts/docker.bash "$(VERSION)" "$(CMD)"
+
+.PHONY: test
+test: straight.elc
+	$(EMACS) -Q --batch -L . -L $(BUTTERCUP) \
+		-l buttercup -f buttercup-run-discover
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -5,6 +5,9 @@ EMACS ?= emacs
 
 SHELL := bash
 
+#Point to path of your local emacs-buttercup install
+BUTTERCUP = ../emacs-buttercup/
```
raxod502, line 9:
```suggestion
BUTTERCUP ?= ../emacs-buttercup/
```
this allows setting it on the command line or by environment variable.

tests/test-straight.el
```diff
@@ -0,0 +1,120 @@
...
+(provide 'straight-test)
+
+;; Local Variables:
+;; End:
+
+;;; straight-test.el ends here
+
+
+
+;;; test-straight.el ends here
```
raxod502, line 120: We can probably ditch most of these headers, like in `bootstrap.el` and `install.el`, as the file isn't part of the package. Would be easier than filling them out with information that would probably be mostly redundant.

Makefile
```diff
@@ -48,7 +51,6 @@ checkdoc: ## Check docstring style
 .PHONY: longlines
 longlines: ## Check for long lines
 	@scripts/check-line-length.bash
-
```
raxod502, line 51: I think this line was deleted by accident?

tests/test-straight.el
```diff
@@ -0,0 +1,120 @@
...
+        (expect (straight--alist-set "a" 'b  '(("c" . d)))
+                :to-equal '(("a" . b) ("c" . d))))
+      (it "sets first KEY match cdr"
+        (expect (straight--alist-set "a" 'd  '(("a" . b) ("a" . c)))
+                :to-equal '(("a" . d) ("a" . c)))))
+    (describe "(key val alist symbol)"
+      (it "tests with eq when SYMBOL non-nil"
+        (expect (straight--alist-set "a" 'a  '((a . b) (a . c)) 'symbol)
+                :to-equal '(("a" . a) (a . b) (a . c))))))
+(describe "straight-vc-git--fork-repo"
```
raxod502, line 45: Should this be nested under `(describe "straight.el" ...)` as well? I'm not actually familiar enough with Buttercup to say.

raxod502 (commented):
Yeah, this looks great!

tests/test-straight.el
progfolio: I usually leave the bare minimum to satisfy checkdoc/package-lint. Up to you, though.

raxod502: Personally, I prefer to either write a proper header, or not have one at all. In this case I think having no header is quite reasonable.

progfolio: Removed

raxod502 (approved):

PR #609, opened 2020-10-15
Translate :branch from melpa recipes

isker:
A select few melpa recipes define a `:branch` property.  Stop dropping it on the floor when creating straight recipes from melpa recipes.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -2803,6 +2803,8 @@ return nil."
                 (straight--put
                  plist :files
                  (append files (list (format "%S-pkg.el" package)))))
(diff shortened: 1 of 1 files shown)


raxod502 (approved):
Huh. What do you know. That's rather embarrassing -- thank you for investigating and fixing the problem.

tummychow:
i think this also has to be backported to master? otherwise #608 has some issues with initial clones

raxod502:
The only reason I didn't backport this was because it was so recent and because it addressed a problem that did not affect very many people, so the risk-to-benefit ratio was not favorable.

Issue #612, opened 2020-10-17
straight-vc--host-username lookup fails when default recipe is nil


progfolio:
<details><summary>Test Case</summary>

```emacs-lisp
(straight-bug-report
  :user-dir "straight-host-username-nil-when-no-default-recipe.straight"
  :post-bootstrap 
  (straight-use-package
   '(svg-tag-mode :type git :host github :repo "rougier/svg-tag-mode" :branch "main" :fork t)))
```
</details>

- Test run at: `2020-10-17 15:25:28`
- system-type: `gnu/linux`
- straight-version: `prerelease (HEAD -> fix/straight--convert-recipe-no-default, origin/develop, develop) c710c1e 2020-10-17`
- emacs-version: `GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, X toolkit, cairo version 1.17.3, Xaw3d scroll bars)
 of 2020-10-14`

<details><summary>Output</summary>

```emacs-lisp
Bootstrapping straight.el...
Bootstrapping straight.el...done
Rebuilding all packages due to build cache schema change
Looking for gnu-elpa-mirror recipe → Cloning melpa...
Looking for gnu-elpa-mirror recipe → Cloning melpa...done
Looking for emacsmirror-mirror recipe → Cloning gnu-elpa-mirror...
Looking for emacsmirror-mirror recipe → Cloning gnu-elpa-mirror...done
Looking for straight recipe → Cloning emacsmirror-mirror...
Looking for straight recipe → Cloning emacsmirror-mirror...done
Building straight...
Building straight...done

Test run with version: prerelease (HEAD -> develop, origin/develop) c710c1e 2020-10-17
Debugger entered--Lisp error: (error "No straight-host-usernames entry for host ‘nil’")
  signal(error ("No straight-host-usernames entry for host ‘nil’"))
  error("No straight-host-usernames entry for host `%S'" nil)
  straight-vc--host-username(nil)
  straight-vc-git--fork-repo((:fork t))
  straight--convert-recipe((svg-tag-mode :type git :host github :repo "rougier/svg-tag-mode" :branch "main" :fork t) nil)
  straight-use-package((svg-tag-mode :type git :host github :repo "rougier/svg-tag-mode" :branch "main" :fork t))
[...]

PR #613, opened 2020-10-17
recipe conversion: use provided recipe for username lookup if  no default

progfolio:
Otherwise striaght--vc--host-username does not recieve enough
information to compute the :repo string.

See: #612


raxod502 (approved):
:sparkles:

Issue #617, opened 2020-10-23
straight-vc: VC backend ‘built-in’ does not implement method ‘keywords’


omidmnz:
## What's wrong
The following snippet (and all my other `:type built-in` recipes) raises the error in the title:
```emacs-lisp
(use-package recentf-mode
  :straight (recentf-mode :type built-in)
  :hook (after-init . recentf-mode))
```

## Directions to reproduce
I have already added the erroneous snippet. This might be specific to my EmacsGCC version, but I have bisected the error to this particular commit: https://github.com/raxod502/straight.el/commit/48b0f1f

### Version information

* Emacs version: GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.21, cairo version 1.16.0)
* Operating system: NixOS

progfolio:
Confirmed with:


<details><summary>Test Case</summary>

```emacs-lisp
(straight-bug-report
  :post-bootstrap 
  (straight-use-package
   '(recentf-mode :type built-in)))
```
</details>

- Test run at: `2020-10-23 11:48:36`
- system-type: `gnu/linux`
- straight-version: `prerelease (HEAD -> feat/straight-use-package-only-show-installed, origin/develop, develop) a856388 2020-10-17`
- emacs-version: `GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, X toolkit, cairo version 1.17.3, Xaw3d scroll bars)
 of 2020-10-14`

<details><summary>Output</summary>

```emacs-lisp
Bootstrapping straight.el...
Bootstrapping straight.el...done
Rebuilding all packages due to build cache schema change
Looking for gnu-elpa-mirror recipe → Cloning melpa...
Looking for gnu-elpa-mirror recipe → Cloning melpa...done
Looking for emacsmirror-mirror recipe → Cloning gnu-elpa-mirror...
Looking for emacsmirror-mirror recipe → Cloning gnu-elpa-mirror...done
Looking for straight recipe → Cloning emacsmirror-mirror...
Looking for straight recipe → Cloning emacsmirror-mirror...done
Building straight...
Building straight...done

Test run with version: prerelease (HEAD -> develop, origin/develop) a856388 2020-10-17
Debugger entered--Lisp error: (error "VC backend ‘built-in’ does not implement method ‘k...")
  signal(error ("VC backend ‘built-in’ does not implement method ‘k..."))
[...]

PR #618, opened 2020-10-23
"straight--convert-recipe: ignore straight-vc-keywords for built-in p…

progfolio:
…ackages"

There are no keywords implemented for the built-in package type.
Calling straight-vc-keywords will error with this type.

See: #617

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -3138,11 +3138,12 @@ for dependency resolution."
                                                              sources
                                                            (list sources))))
                          plist))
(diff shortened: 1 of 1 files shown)


raxod502:
Oopsies. I guess perhaps I should have just made `built-in` into an actual VC backend, instead of hacking it in as a special case in a bunch of places. I was just too lazy at the time to write all those stub functions, I think...

PR #626, opened 2020-11-03
straight-bug-report: switch buffer on idle timer

progfolio:
Prevents results buffer from stealing focus if user is doing something
else, e.g. typing while a longer running test is running.


raxod502:
Nice solution! I guess the alternatively would be to just display the buffer and not give it focus when it's displayed.

PR #627, opened 2020-11-04
straight-bug-report: nil-bind print level/length limits

progfolio:
Prevent truncation of printed lisp forms.


raxod502:
Yep, that one always bites me eventually too.

progfolio:
Same, luckily I hit it quickly when testing out the el-get integration.

PR #628, opened 2020-11-08
Fix defcustom for straight-check-for-modifications

aspiers:
A `:type` of `'(list ...)` fixes the number of elements, so anything other than a singleton list resulted in `(mismatch)` in the Custom buffer.  So change to `'(repeat ...)` to allow a variable number of symbols in the list.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -3456,7 +3456,7 @@ follows:
 
 This usage is deprecated and will be removed."
   :type
-  '(list
+  '(repeat
     (choice
      (const :tag "Use find(1) at startup" find-at-startup)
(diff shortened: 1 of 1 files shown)


raxod502:
I wonder if there is a tool that checks all your defcustoms to see if the default values meet the `:type` specifications. It would not be perfect, but it would catch some of these bugs.

progfolio:
That's an interesting idea. I do some simple type checking for doct's functions with a kind of light fuzzing:

https://github.com/progfolio/doct/blob/dabb30ebea866ef225b81561c8265d740b1e81c3/tests/doct-test.el#L148-L165

PR #629, opened 2020-11-09
Fix installation issue on Windows

clockoon:
Due to the way handling Unicode on Windows, the evaluation of `straight.el` could fail in some cases. Adding comment `;; -*- coding: utf-8 -*-` will force the encoding, therefore could solve this problem.


raxod502:
Got it. Yes, `print` already adds the necessary line break.

raxod502 (approved):

raxod502:
Thanks for the help!

PR #633, opened 2020-11-17
straight--run-build-commands: error on failed :pre/post-build command

progfolio:
Prefer errors over warnings for command failure.
Previous behavior would allow package :pre/post-build to fail and be silently
ignored on restarts.

Now if a `:pre/post-build` error occurs the user is presented with the `*Warnings*` buffer (which details the error and the recipe in which it occurred) and, in the case of a shell command error, the `*straight-process-buffer*` on start.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -4549,14 +4549,28 @@ The keyword's value is expected to be one of the following:
   (straight--with-plist recipe (pre-build post-build package local-repo)
     (when-let ((commands (if post post-build pre-build))
                (repo (straight--repos-dir (or local-repo package))))
-      (let ((default-directory repo))
+      (let ((default-directory repo)
+            (commanderror nil))
         ;; Allow a single command or a list of commands.
         (dolist (command (if (cl-every #'listp commands)
                              commands
                            (list commands)))
-          (if (cl-every #'stringp command)
-              (apply #'straight--warn-call command)
-            (eval command)))))))
+          (condition-case err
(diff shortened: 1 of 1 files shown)


progfolio:
e.g.

raxod502 (commented):
This seems like it should work, but perhaps we would be able to facilitate greater code reuse if we were to use `straight--get-call` instead of `straight--check-call`, because `straight--get-call` will already throw an error and display the process buffer when the command fails. Then all we would have to do here is (optionally) re-throw the error with some more explanation in the description. What do you think?

progfolio:
`straight--get-call` throws the error, but it does not display the `straight-process-buffer` when an error is thrown.  Would prefer a patch similar to #573 for `straight--get-call` as well?

raxod502:
Oh, I didn't realize that. I guess it is up to you what you think would be more elegant. I don't have a strong opinion. It would be nice as a concept for any command errors to pop open the buffer in the same way, but on the other hand it's possible that in some circumstance maybe we do not want to pop open the buffer on an error.

progfolio:
> It would be nice as a concept for any command errors to pop open the buffer in the same way, but on the other hand it's possible that in some circumstance maybe we do not want to pop open the buffer on an error.

Agreed. I haven't thought of a clean interface to allow this just yet, but I have it on my TODO list.
I'm going to merge this as is so we get the benefit of catching the errors. I have it in my notes to clean this up once the above is addressed.

Issue #635, opened 2020-11-22
Drop support for 24.5?


progfolio:
According to https://www.gnu.org/software/emacs/history.html

> 2015-04-10 - Emacs 24.5 released

Coming up on six years for that release and most of our compatibility code (shims for if-let, when-let, backquote requirements in pcase statements) could be phased out if we bump our minimum version to 25.1 (released 2016-09-17).

Not scientific by any means, but a recent poll shows a majority of users well beyond that version:

https://www.reddit.com/poll/imwewj?utm_source=reddit&utm_medium=usertext&utm_name=emacs&utm_content=t3_imwewj

Another survey is open until 2020-11-30:

https://emacssurvey.org/

I think it'd be useful to see the results of this survey, as I suspect it was more widely promoted and will have more participants.

I also realized that 24.5 does not have support for `make-process`, so unfortunately we lose `straight-bug-report` on that version and it may complicate the implementation of #76 (depending on our approach).

Thoughts?

raxod502:
Yep, I'd say it's totally safe to drop 24.5. It is, indeed, quite ancient :D 
It is actually pretty rare to find _any_ package that supports 24.5 these days, and I think more than once the smoke test in CI for `straight.el` broke because the package being installed as a test stopped supported that version.

progfolio:
Taken care of w #636

PR #636, opened 2020-11-22
Drop 24.5 support 

progfolio:
See: #635

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -6,7 +6,7 @@
 ;; Created: 1 Jan 2017
 ;; Homepage: https://github.com/raxod502/straight.el
 ;; Keywords: extensions
-;; Package-Requires: ((emacs "24.5"))
+;; Package-Requires: ((emacs "25.1"))
 ;; SPDX-License-Identifier: MIT
 ;; Version: prerelease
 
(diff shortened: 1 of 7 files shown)


straight.el
```diff
@@ -5084,14 +5026,7 @@ The default value is \"Processing\"."
...
                                         (setq next-repos (cdr next-repos))
                                         (cl-return-from loop))
-                                    (error e)
-                                    ;; Emacs 24.5 has a bug where the
-                                    ;; byte-compiler signals an unused
-                                    ;; argument warning for the target
-                                    ;; of a `condition-case' unless
-                                    ;; it's used on every error
-                                    ;; handler.
-                                    (quit (ignore e)))))
```
progfolio, line 5094: Not entirely sure about this change

straight.el
```diff
@@ -1230,8 +1185,8 @@ appropriately."
     (setq straight-find-flavor (straight--determine-find-flavor)))
   (memq symbol
         (pcase straight-find-flavor
-          (`gnu/bsd '(newermt))
-          (`busybox nil)
+          ('gnu/bsd '(newermt))
+          ('busybox nil)
```
raxod502, line 1189: Nice, I'll have to remember to stop using the backticks now that we don't have to support that ancient version!

straight.el
raxod502: I think we want to replace `(quit (ignore e))` with just `(quit)`.

straight.el
progfolio: Yeah, that's one that I always forget about myself. Good to have to only remember one way now.

straight.el
progfolio: Addressed in latest commit.

PR #637, opened 2020-11-23
Fixes for the latest native-compile-async API

flatwhatson:
A recent change on the `native-comp` branch has meant that passing load flag `late` is equivalent to passing load flag `t`, which causes packages to be unconditionally loaded after they're compiled.  This leads to all sorts of problems as some Elisp packages will error instantly on load!

~~This patch simply removes the `load` flag which should give the desired behaviour.~~

Edit: After discussion with upstream, we've agreed on an API, this PR implements the necessary changes.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -114,7 +114,7 @@ They are still logged to the *Messages* buffer.")))
 ;;;; Functions from other packages
 
 ;; `comp'
-(defvar comp-deferred-compilation-black-list)
+(defvar comp-deferred-compilation-deny-list)
 
 ;; `finder-inf'
 (defvar package--builtins)
@@ -4723,6 +4723,12 @@ individual package recipe."
        (not (straight--plist-get recipe :no-native-compile
                                  straight-disable-native-compilation))))
 
+(defun straight--native-compile-file-p (file)
+  "Predicate to check whether FILE should be native-compiled."
+  (not (cl-some (lambda (re)
+                  (string-match-p re file))
+                comp-deferred-compilation-deny-list)))
+
 (defun straight--native-compile-package (recipe)
   "Queue native compilation for the symlinked package specified by RECIPE.
 RECIPE should be a straight.el-style plist. Note that this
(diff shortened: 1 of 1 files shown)


progfolio:
Thank you for the pull. Just to make sure I understand correctly, this is due to the changes here?:

https://github.com/emacs-mirror/emacs/commit/6781cd670d1487bbf0364d80de68ca9733342769#diff-33e02a4d29a78e1ab21694eaba92b1b59839106ffb711326a7b7e344a4706b07

flatwhatson:
Yes that's correct.

flatwhatson:
FYI this change also breaks our usage of `comp-deferred-compilation-black-list` to prevent compilation of `no-native-compile` packages, because that list is only checked when compilation is requested with the `late` flag.

flatwhatson:
I've opened a ticket upstream to discuss this issue: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=44813

flatwhatson:
Ok, Andrea has kindly implemented the necessary changes to support straight's use-case upstream.  I've updated this patch with the necessary changes on our side.

The only possible problem I can see is that this is a hard API break; older builds of native-comp will no longer be supported.  Personally I think this is fine, but it should at least be mentioned.  I've added a warning note in the commit message.

progfolio:
CI failure due to:
```
Warning (emacs): 
straight.el:4748: White space found at end of line
```

progfolio (approved):
I don't use the native-comp branch just yet, but I read through your previous contributions and the latest patch to the API and everything looks good to me. We could `define-obsolete-variable-alias` for the symbol name change, but I don't think it's necessarily a blocker. As you've mentioned before this is a change that will affect a limited number of users (on our develop branch and on the native-comp branch of Emacs)

@raxod502: thoughts?

progfolio (approved):

flatwhatson:
Sorry for the fresh push, the previous version had a silly bug in `straight--native-compile-file-p`.

progfolio:
Not a problem. Just inverted logic for the predicate, correct?

flatwhatson:
Yeah.  Not sure how I missed that the first time through testing.  Whoops!

raxod502 (approved):
If it's a change in the variable name upstream, I don't think it's sensible to try adding an alias here in `straight.el`. Might be worth adding one upstream, depending on what the policy is there. I have no strong thoughts. I like the name change though, always good to be more inclusive.

progfolio:
Thank you, @flatwhatson.

PR #660, opened 2020-12-21
Makefile: Add 'all' target

progfolio:
A small convenience. Defaults to running `clean lint`.


raxod502:
Seems good to me. I often put the `make help` target first so that running `make` with no arguments will tell you the available targets, but the most common convention is for `make` with no arguments to run a "normal build", so doing it the way you suggest here is fine also.

Issue #659, opened 2020-12-21
Revamp build keywords


progfolio:
While working on #658, I noticed there is no `:no-info` build keyword to prevent straight from compiling the texinfo for a package.
Implementing this keyword is simple enough, but thinking about it gave me an idea: Rather than having individual `:no-` keywords to control different aspects of a package build, we could  have a `:build` keyword which operates like the `:files` directive.
Allowed values could be:

- `:build nil`  
  which replaces `:no-build`
- a list of symbols specifying which build steps to execute, e.g. `:build (autoloads compile native-compile info)`. 
Steps are executed in the order they are declared. In the future, we could make these steps asynchronous and allow a syntax to specify callbacks. e.g. given `:build ((one two) (three four))` each pair is run asynchronously with `two` executing after `one` and `four` after `three`. 
- a list of symbols prefixed by `:not` to prevent a build step, e.g. `:build (:not compile native-compile)`. These would inherit from a default list of build steps.

Admittedly, this costs a few more characters than the `:no-` prefixed keywords, but it simplifies the list of build keywords and (when combined with #654) clarifies straight's notion of building a package.

It also opens the door for extensibility. We could enforce a naming scheme for these build step functions similar to the naming scheme for recipe-repository functions. Adding a new step would be a matter of implementing that function and inserting it in the `:defaults` list.

Hope that was clear. Here's a final example which illustrates what  `org-plus-contrib`'s recipe could look like when combined with #654:

```emacs-lisp
(org-plus-contrib :type git
                  :repo "https://code.orgmode.org/bzg/org-mode.git"
                  :local-repo "org"
                  :files (:defaults "contrib/lisp/*.el" "lisp/*.el")
                  :pre-build ("make" "autoloads" "EMACS=/usr/bin/emacs")
                  :build (:not autoloads)
                  :post-build (message "example"))
```

Thoughts?

raxod502:
Excellent! This is a great improvement.

Issue #654, opened 2020-12-21
Rename :build keyword to :pre-build?


progfolio:
As mentioned here: https://github.com/raxod502/straight.el/pull/535#issuecomment-657144652

The term "build" is overloaded at the moment. We have the `:build` and `:post-build` recipe keywords and the notion of building an elisp package. Can't believe I didn't think of this at the time, but we could rename `:build` to `:pre-build`. It would clearly complement `:post-build` and distinguish the keyword from the symlink/byte-compile/autoload notion of "build".

progfolio:
See #655

PR #662, opened 2020-12-28
Refactor/straight  build functions

progfolio:
An implementation of what I was suggesting with #659 and #655.
The README will need to be reworded, but I figured we could discuss the viability of such an approach before making substantial edits.

A recipe may specify build steps via the `:build` keyword.
(Note the original `:build` keyword has been renamed to `:pre-build`)

The `:build` keyword accepts the following values:

- `nil`
No build is executed. `:pre-build` and `:post-build` commands are not run.

```emacs-lisp
     (example :build nil)
```

- `t`
Equivalent to running `straight--default-build-steps` (see below).

- `(step...)`
Each step in the list is a symbol which represents a function named `straight--build-SYMBOL`.
The function is passed the recipe as its sole argument.
Steps are executed in the order they are listed.

```emacs-lisp
(example :build (autoloads compile native-compile info))
```

- `(:not step...)`
  If the car of the list is the `:not` keyword, the build steps are the
  set difference of `straight--build-default-steps` and the cdr of the
  recipe's `:build` list.
  e.g. with `straight--build-default-steps` set to:

```emacs-lisp
(autoloads compile native-compile info)
```

   and the following recipe:

```emacs-lisp
(example :build (:not compile info))
```

  the package is not byte compiled and does not have it's texinfo generated.

  Steps may be disabled globally via the customization variable named
  `straight-disable-SYMBOL`. e.g.:

```emacs-lisp
(setq straight-disable-info t)
;;info generation disabled when :build is not explicitly declared
(straight-use-package '(example))
;;or starts with :not
;;Info disabled in addition to compilation.
(straight-use-package '(example :build (:not compile)))
```

In the absence of a `:build` keyword, `straight--build-default-steps` are run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ chat][gitter-badge]][gitter]
   * [How do I pin package versions or use only tagged releases?](#how-do-i-pin-package-versions-or-use-only-tagged-releases)
   * [How can I use the built-in version of a package?](#how-can-i-use-the-built-in-version-of-a-package)
 - [News](#news)
+  * [Jan 1, 2021](#jan-1-2021)
   * [April 19, 2020](#april-19-2020)
 
 <!-- tocstop -->
@@ -502,7 +503,7 @@ repository, but this is not necessary.
 
 The build recipe is not a literal data structure. It is a concept that
 represents a certain subset of the package's recipe. Specifically, the
-`:files`, `:local-repo`, and `:no-build` keywords.
+`:files`, `:local-repo`, and `:build` keywords.
 
 To transform this *information* into an actual package that Emacs can
 load, `straight.el` *builds* the package. This means that some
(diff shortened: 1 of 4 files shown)


raxod502:
Yeah, this all sounds perfect! It seems like a much better interface all around. Same with your comments in https://github.com/raxod502/straight.el/pull/655, https://github.com/raxod502/straight.el/issues/654, https://github.com/raxod502/straight.el/issues/659: this all seems great.

progfolio:
Sounds good. I'll add some tests, `define-obsolet-variable-alias`es for the option name changes, and edit the documentation. How do you prefer to deal with the news of the `:build` -> `:pre-build` rename? Is a notice in the README sufficient?

progfolio:
Added a unit test for `straight--build-setps`, properly obsoleted changed defcustoms, updated README and made note of the breaking change in the NEWS section.

raxod502:
> Is a News entry in the README sufficient?

I think that would be fine, given that `:build` itself was only added recently.

PR #678, opened 2021-01-16
fix typo in documentation of straight-bug-report

mjoerg:


progfolio:
Good catch. Thank you.

Issue #665, opened 2020-12-31
Windows: Have to manually clone in to the straight/repos/straight.el folder to make it work


pRot0ta1p:
## What's wrong

On a fresh install of Emacs 27.1, have to clone straight.el in to .emas.d/straight/repos/straight.el to install straight.el, 
otherwise, emacs will hang on start, this also happens with the msys version of emacs 27.1, and my self-compiled native-comp branch emacs.

## Directions to reproduce
In a fresh install emacs, with empty .emacs.d folder, and a init.el with following content:

```
(defvar bootstrap-version)
(let ((bootstrap-file
       (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
      (bootstrap-version 5))
  (unless (file-exists-p bootstrap-file)
    (with-current-buffer
        (url-retrieve-synchronously
         "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
         'silent 'inhibit-cookies)
      (goto-char (point-max))
      (eval-print-last-sexp)))
  (load bootstrap-file nil 'nomessage))
```

Then start emacs.exe or runemacs.exe. emacs will hang but with abnormally high cpu usage.

### Version information

* Emacs version: 27.1
* Operating system: windows 10 version 2004

I dont recall such a problem in windows 10 version 1909, so maybe its a windows problem?...

raxod502:
Not sure what this could be. Here are some things that might help identify the problem:

Does it print anything before hanging? If you type `C-g` or send `SIGUSR2` to Emacs (I don't know if the latter is even possible in Windows), does that produce any useful behavior? If you add `(setq debug-on-error t)` and/or `(setq debug-on-quit t)` before the bootstrap snippet, does that manage to produce a stack trace when you poke Emacs while hung?

pRot0ta1p:
> Does it print anything before hanging? If you type `C-g` or send `SIGUSR2` to Emacs (I don't know if the latter is even possible in Windows), does that produce any useful behavior?

`C-g` can un-hung up emacs, but will wont throw up a backtrace buffer or anything, in *Message* buffer there is only a single line: 
```
Quit
```

> If you add `(setq debug-on-error t)` and/or `(setq debug-on-quit t)` before the bootstrap snippet, does that manage to produce a stack trace when you poke Emacs while hung?

Also tried this, but nothing showed up, I tried like 10 time, 2 of these I get a error said :
```
GnuTLS error: raw.githubusercontent.com, -53
[...]

PR #691, opened 2021-02-02
Refuse to load if git executable not found.

progfolio:
If the user does not have git installed or it is not on the `exec-path` we can error sooner.
This should help us pinpoint those types of bugs sooner.
See #665

### DIFF
--- a/install.el
+++ b/install.el
@@ -85,6 +85,9 @@
       (error "The `straight-repository-branch' must be a string (was: %S)"
              straight-repository-branch)))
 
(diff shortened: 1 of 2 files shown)


raxod502:
This is a good idea; we should probably also go even farther and fail the installation process in `install.el`, [like we do when the Emacs version is too old](https://github.com/raxod502/straight.el/blob/0f9b828d8a41cf3d312678e82573066aebf2ab6e/install.el#L74-L81). After all, the installation process also requires Git.

PR #693, opened 2021-02-03
straight--directory-files: remove ".", "..", ".git" dirs when FULL paths

progfolio:
Previous logic assumed the FULL argument was nil when filtering directories.


raxod502:
Oof. Thanks for catching that!

Issue #692, opened 2021-02-03
Reimplement file watcher in elisp


progfolio:
`filenotify.el` provides an abstraction layer over OS file system watchers.
I was curious if you had explored this as an option for straight's file system watcher.

Pros:

-   Reduces dependencies on external tooling: python, watchexec
-   Should sidestep #447
-   Should be easy to implement #291
-   straight.el code reuse in watcher script.

Cons (I'm aware of):

-   Emacs must be compiled with file notification support (Though it seems like this is a default option as of Emacs 24.4: https://www.masteringemacs.org/article/whats-new-in-emacs-24-4)

The idea just occurred to me, and I'm thinking we could do something similar to what we already do using a detached child Emacs process.

Issue #519, opened 2020-05-05
Org is rebuilt even though no source files have changed (just the mtime of the lisp subdir)


andersjohansson:
## What's wrong
I often visit the source of org files, and even if I don’t change anything, the mtime of the `lisp` subdir in the org repo where all elisp files reside is often changed because flycheck creates temporary files that are later removed. The command for detecting changes which (as I understand it) should be something like `find org -name .git -prune -o -newermt CACHEDMTIME -print` will then output `lisp` and org will be rebuilt, even though no "real" changes to the source have occurred.

This got a lot more noticeable now that I’m testing out the emacs `native-comp` branch and all fans go off at max when recompiling all the org elisp to `eln` files :slightly_smiling_face:.

I suppose this only happens for repos like org where the elisp files are in a subdir. Temporary and removed files in the main repo directory resulting in changed mtime for that directory is not registered as a change. Would it be feasible with more precision for the change detection? But than we would risk to miss what should be detected as a change in other repos? For example only looking at elisp files won’t work, as it’s that’s clearly not enough for for many packages that contain for example C code like pdf-tools)?

## Directions to reproduce
I created a `.emacs.d` with only this `init.el`
```emacs-lisp
(setq straight-use-package-by-default t
      straight-vc-git-default-clone-depth 1
      straight-cache-autoloads t
      straight-repository-branch "develop")

(defvar bootstrap-version)
(let ((bootstrap-file
       (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
      (bootstrap-version 5))
  (unless (file-exists-p bootstrap-file)
    (with-current-buffer
        (url-retrieve-synchronously
         "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
         'silent 'inhibit-cookies)
      (goto-char (point-max))
      (eval-print-last-sexp)))
  (load bootstrap-file nil 'nomessage))

(straight-use-package 'use-package)

(use-package org
  :commands org-mode
  :demand t)
```
Then ran it:
`emacs --quick --eval "(setq user-emacs-directory \"/home/aj/.emacs.d-test-straight\")" --eval "(setq user-init-file \"/home/aj/.emacs.d-test-straight/init.el\")" --load "/home/aj/.emacs.d-test-straight/init.el"`

Straight, use-package and org is installed.
Quitting and restarting, everything works as expected.
```sh
touch ~/.emacs.d-test-straight/straight/repos/org/lisp/testfile
[...]

Issue #508, opened 2020-04-18
Flycheck temporary files cause spurious rebuilds


doolio:
I've noticed an odd observation where, when I visit a file in a `straight`  repo without making any changes the package is rebuilt on the next startup of Emacs. I understood from the [README](https://github.com/raxod502/straight.el#how-does-straightel-know-when-to-rebuild-packages) that packages are only rebuilt when changes are made to the local repository. Is this a bug or a misconfiguration on my part?

The only customisations I've made to `straight` are as follows:
```
 '(straight-process-buffer " *straight-process*")
 '(straight-profiles (quote ((nil . "../../pkg-versions.el"))))
 '(straight-use-package-by-default t)
```

Thanks for your time.

raxod502:
What is the value of `straight-check-for-modifications` in your configuration? Do you have anything in your configuration that might cause files to be created, deleted, or modified when you visit a file (such as auto-save, lockfiles, etc.)? If so then that would of course cause `straight.el` to think there was a modification because the directory mtime will be updated.

doolio:
Sorry for the late reply.

`straight-check-for-modifications` is at the default value of `(find-at-startup find-when-checking)`. Well, if I'm not mistaken doesn't Emacs by default create auto-save files. I haven't changed this behaviour but I assumed such files were only created when the file is edited and not simply visited. Similarly, I've not changed the default setting for `create-lockfiles` which is set to `t`. Should I?

I do configure several settings related to backup files (e.g. `backup-by-copy`, `delete-old-versions`, `version-control` all set to `t` - though `vc-make-backup-files` is set to the default value of `nil`) but again assumed such files are only created when a file is edited but perhaps I'm wrong in my assumption. Would this explain what I've observed?

Thanks for your time.

raxod502:
Your assumptions seem reasonable. Why don't you check `*straight-process*` to see what files `straight.el` thinks are modified?

doolio:
>  Why don't you check *straight-process* to see what files straight.el thinks are modified?

Ok, so after the `find` command I see the last repo directory that I visit. For this test case, it is:

`/home/doolio/.emacs.d/straight/repos/org-gtd/`

but not the file I visit namely `org-gtd.el`. If I visit more than one straight repo before restarting Emacs it will list the different repos as so but again not the file itself:
````
/home/doolio/.emacs.d/straight/repos/org-gtd/
/home/doolio/.emacs.d/straight/repos/no-littering/
````

[...]

Issue #488, opened 2020-03-01
An option to use different find tools


seagle0128:
Generally `find` is common but slow on the most of platform. Is there any way to use `fd` or `rg` instead of `find`(1)? They have much better performance and I believe it's benefit for the startup speed. I couldn't find an option to do that. Thanks!

raxod502:
No. Such a feature could be useful, though.

manuel-uberti:
Chiming in just to say "yes, please!". :)

PR #694, opened 2021-02-03
Add straight-watcher.el

progfolio:
Reimplementation of the watcher script in elisp.

See: #692

Not quite complete, but I figured I'd share it to show the general direction so far.


progfolio:
In consideration of #390, #447, #488, #508, #519, and #712:

I think it would be best to move away from using file mtimes as our modification heuristic.
The mtime approach works well in that `find` is fast, but it has no way of telling us whether or not the current
content of the package warrants a rebuild. What I'm thinking is that we could store a hash and compare against that.

The question is, what do we hash? We could have the watcher process expand the repo recipe's `:files`, append those files in a temp buffer and hash the contents.
The watcher can save these to disk. We can then compare the package's last built hash (constructed as above and stored in `straight--build-cache` instead of the mtime during the build process) with the latest value the watcher reported.
We only rebuild if the hashes do not match.

  Pros:
  
  - replaces `find` dependency with elisp
  - Should have similar performance to the `find` method on start-up. Of course I'd have to measure this, but we'd only be paying the one-time cost of reading the wathcer's data in from disk and one hash lookup + string comparison per package.
  - Simplifies the codebase. We would have less points of failure if/when packages are not being built at the appropriate time.
  
  Cons:
  
  - Memory/CPU usage of having the watcher process run. I haven't done any formal measurements, but the cost seems negligible IMO. Especially if we were to measure it against the cost of spurious rebuilds.
    In any case I'd aim to keep that cost minimal.
  - Less redundancy to detect changes. I'm not sure how much this will matter in practice, though. I personally only use the watcher now and spare #447 (which would be a non-issue with an elisp implementation) have rarely had any problems.
    
  Thoughts?

raxod502:
Yeah! I think that makes a lot of sense. In fact, if we implement this, we might be able to get rid of the `find` approach entirely, and force everyone to use the filesystem watcher, because it wouldn't have problematic system dependencies. That would certainly be an improvement in maintainability.

I think hashing the `:files` recipe makes a lot of sense, and it would have no problem with performance as long as it's done in the watcher subprocess which won't block init. I suppose there could be a race condition where you make a modification and then restart Emacs before the watcher has the chance to do its analysis, but that's no worse than what we have with the current filesystem watcher, and I've never actually run into that case in my few years of usage.

Here's a fun trick I've used in the past to have an external process write hashes like this in a format that's super fast to read: when package `xyz` gets processed by the watcher process and it computes hash `abcdef`, write a file `xyz-abcdef` into the results directory. That way, you can get a listing of all the detected modifications as well as their hashes, from a single `directory-files` call, and you can copy the relevant data into the build cache before clearing out the directory.

Thanks for taking the lead on this improvement!

progfolio:
This functionality (and arguably many of the other components of straight.el) would be easier to reason about if it were modularized into its own file/sub-feature. However, loading that file is complicated by the bootstrap/installation process. Haven't found a clean way to do this yet.

raxod502:
I'll certainly agree with you that `straight.el` would be more maintainable if it were split into multiple files. In my past work I had an alarming tendency to never split code into multiple files, which I've since learned from.

I think I see what you mean about the bootstrap process. Here's one useful note, though: anything that isn't actually needed during initial bootstrap can be in a separate file no problem, since the single-file download used during bootstrap is only used to install `straight.el` itself, and then is discarded.

You just need to make sure that any of these separate files used after init/bootstrap are loaded only when needed, and not eagerly. (Which is good for performance anyway.)

Since a filesystem watcher doesn't need to be launched during initial installation, it's fine to put this code in a separate file, as long as we `require` it only when needed.

progfolio:
I won't be working on this feature any time soon, but anyone who is interested is welcome to any of the code in this patch.

One potential problem with this approach is the large number of open file descriptors when simulating recursive directory watching. There was a proposed patch to Emacs's library to allow recursive directory watching, but it stalled.

PR #699, opened 2021-02-22
Include ox-odt styles when building org

andersjohansson:
ox-odt looks for template-files in the etc/styles subdirectory. Including this and the styles files ensures correct behaviour. Previously, ox-odt found the styles-directory included with the org-version packaged with emacs (see the definition of org-odt-styles-dir-list).


progfolio (commented):
Thank you for catching that. Your change to the recipe looks good to me. Can you bump `straight-recipes-org-elpa-version` to `7`, please?

andersjohansson:
Done!

progfolio (approved):
Thanks again.

Issue #354, opened 2019-02-19
Log of changes of updated packages


blaenk:
Something I really miss from [Paradox](https://github.com/Malabarba/paradox) is the ability to very easily see a git log of the changes to a package since it was last updated. This was amazing for discovering new features added to a package, or otherwise keep up with its development, which in my case encouraged involvement (contribution, feedback, etc.) in package development. One could argue this benefits the emacs package ecosystem.

I feel like it should be even more possible/appropriate with straight given that with Paradox it would have to find the GitHub git repo, assuming it was even on GitHub, and then query the GitHub API to obtain the commit list, since it can't assume an existing local repo. Straight could instead invoke `git log old..new` or whatever the syntax for that is, and present it in a neat format.

It would be amazing if a magit-like interface were presented. In fact, maybe magit features themselves can be leveraged, given they already are when intervention of certain problems is required. Just did a quick skim of the manual and trying `(magit-log-other '("oldrev..newreg"))` accomplishes this, though I have no clue if that would be the best way. If magit were delegated to, the only remaining link would be an easy list/dash view of all updated packages which upon pressing Enter (for example) would open this buffer.

raxod502:
Yep, agreed and sounds good. We should use Magit for this. The main technical problem to be solved is figuring out what the old revision was. I think the best paradigm is to copy Git and keep a file on disk somewhere that records these revisions (à la `FETCH_HEAD`)—or maybe we can even leverage Git itself to keep track of it (would using `FETCH_HEAD` directly work)?

This feature has been proposed before, but has never had a dedicated issue. Thanks for making one :)

SkySkimmer:
`(setq straight-vc-git-auto-fast-forward nil)`

When I want to update my packages I do
- `M-x straight-fetch-all` (go watch cat videos or someting while it downloads)
- `M-x straight-merge-all`

For every package with possible updates straight will say something like "remote is ahead of local" (not verbatim) and offer to open in magit with `g` (without the `setq` above it silently updates).
Then for magit I have `'([unpulled status] . show)` in `magit-section-initial-visibility-alist` so it shows the list of unpulled commits, or I can magit-log. Looks like this: 
![magit](https://user-images.githubusercontent.com/2461932/53046224-df219b00-348f-11e9-9184-45900f29c64c.png)

To go back to the straight prompt `C-M-c`. If you updated in magit it will go to the next package, otherwise you can tell straight to update with `f`.

[...]

PR #711, opened 2021-03-07
Feature/overhaul interactive merge

pdbrown:
Hi there, thanks for this package manager! It's great and has brought some sanity back to my Emacs config. I started using it a little while ago, and was recently been doing a bunch of package upgrades. I thought it might be cool if there were a way to pop up a log of exactly the commits an update would include, without firing up all of magit and having to see for myself. I started digging through the code to see what it would take to add this and I noticed a few things that seemed like they could be improved on, and after making a few notes, see here
https://github.com/pdbrown/straight.el/compare/scratch/interactive-merge%5E...scratch/interactive-merge
ended up overhauling a larger portion than I'd set out to.
In addition to a bit of rewriting, this change adds a new option during interactive updates:
```l --- Git log $COMMIT_RANGE and open recursive edit```
to pop up a git log buffer.

It also fixes a few related issues:

;; Rebase in progress
https://github.com/raxod502/straight.el/issues/271

;; Does not appear to handle detached head correctly
https://github.com/raxod502/straight.el/issues/194

;; Duplicate fast forward popup
https://github.com/raxod502/straight.el/issues/205

;; Ambiguous ref names
https://github.com/raxod502/straight.el/issues/306

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -79,9 +79,6 @@ They are still logged to the *Messages* buffer.")))
 (declare-function flycheck-get-next-checker-for-buffer "flycheck")
 (declare-function flycheck-start-current-syntax-check "flycheck")
 
-;; `magit'
-(declare-function magit-status-setup-buffer "magit-status")
-
 ;; `package'
 (defvar package-selected-packages)
 (declare-function package--ensure-init-file "package")
@@ -1754,29 +1751,28 @@ appropriately."
                        props))
          ,@body))))
 
-;; We don't define `straight--profile-cache' until later. I don't
-;; think it's possible to avoid the circular dependency in any sane
-;; way, since the VC layer is used to clone packages and we need to
-;; clone Magit here.
-(defvar straight--profile-cache)
-
-(cl-defun straight--magit-status (directory)
-  "Like `magit-status', but install Magit if necessary.
-Magit is only installed if the user responds to a `y-or-n-p'
-prompt. Return non-nil if Magit was installed. DIRECTORY is as in
-`magit-status'."
-  (unless (or (require 'magit nil 'noerror)
-              (gethash "magit" straight--profile-cache))
-    (if (y-or-n-p "Install Magit? ")
+(defun straight--ensure-magit-p (feature)
+  "Load Magit FEATURE and return non-nil if it's avaliable.
+If the feature isn't available, offer to install Magit with a
+`y-or-n-p' prompt before loading it."
+  (or (require feature nil 'noerror)
+      (when (y-or-n-p "Install Magit? ")
         (straight-use-package 'magit)
-      (cl-return-from straight--magit-status)))
-  (prog1 t
-    (magit-status-setup-buffer directory)))
+        (require feature nil 'noerror))))
+
+(defun straight--magit-status (directory)
+  "Like `magit-status', but offer to install Magit if necessary.
+Return non-nil if `magit-status-setup-buffer' is invoked successfully.
+DIRECTORY is as in `magit-status'."
+  (when (and (straight--ensure-magit-p 'magit-status)
+             (fboundp 'magit-status-setup-buffer))
+    (magit-status-setup-buffer directory)
+    t))
 
 (defun straight--recursive-edit ()
   "Start a new recursive edit session.
 Make sure that other packages such as `server.el' don't cause us
-to loose our session."
+to lose our session."
   ;; Don't mess up recursive straight.el operations. The wonderful
   ;; thing about using our own variable is that since it's not
   ;; buffer-local, a recursive binding to nil is actually able to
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: straight.el [hunk 1]
```diff
@@ -2037,139 +2063,274 @@ LOCAL-REPO is a string."
...
+            ("c" (if head-detached
+                     (format "Attach HEAD to %S" left-ref)
+                   (format "Checkout %S" left-ref))
+             (straight--get-call "git" "checkout" left-ref)))))
+       ;; Else done; default branch is checked out:
+       (t t)))))
+
+(defun straight-vc-git--ensure-default-branch-current
+    (local-repo default-branch remote-branch)
+  "Ensure that DEFAULT-BRANCH is ahead of REMOTE-BRANCH in LOCAL-REPO.
```
progfolio, line 2307: CI is failing due to checkdoc complaining about the order of the arguments in the docstring.

```suggestion
  "Ensure LOCAL-REPO's DEFAULT-BRANCH is ahead of REMOTE-BRANCH.
```

straight.el
```diff
@@ -5177,6 +5338,30 @@ local repository is already on disk."
               (or (null predicate) (funcall predicate package))))))
    action))
 
+;;;;; Changelog popup
+
+(defvar straight-changelog-buffer "*straight-changelog*")
+
+(defun straight-vc-git--changelog (local-repo range &rest extra-git-args)
+  "Show changelog buffer for RANGE of commits in LOCAL-REPO.
```
progfolio, line 5346: Same as above:

```suggestion
  "Show LOCAL-REPO's changelog buffer for RANGE of commits.
```

straight.el
```diff
@@ -2037,139 +2063,274 @@ LOCAL-REPO is a string."
...
-(cl-defun straight-vc-git--ensure-head (local-repo branch &optional ref)
-  "Ensure that LOCAL-REPO has BRANCH checked out.
-If REF is non-nil, instead ensure that BRANCH is ahead of REF.
-Any untracked files created by checkout will be deleted without
-confirmation, so this function should only be run after
-`straight-vc-git--ensure-worktree' has passed."
+(defun straight-vc-git--abbrev-ref (ref)
+  "Return the unambiguous, abbreviated name of REF."
+  ;; Force locale for git rev-parse.
+  (let ((process-environment (append (list "LC_ALL=C") process-environment)))
```
progfolio, line 2069: If this pattern of overriding the locale is common enough, we could write a macro for it:
Something like the following should work:

```emacs-lisp
(defmacro straight-vc-git-with-locale (&rest body)
  "Execute BODY with user locale overridden."
  (declare (indent defun))
  `(let ((process-environment (append (list "LC_ALL=C") process-environment)))
     ,@body))
```

straight.el
```diff
@@ -2037,139 +2063,274 @@ LOCAL-REPO is a string."
...
+    ;; We run git rev-parse twice here to more reliably detect
+    ;; ambiguous refs. When ref is ambiguous, rev-parse exits
+    ;; successfully and prints a warning to stderr. Ideally, we'd look
+    ;; for exit status 0 and non-empty stderr to detect this case, but
+    ;; the current `straight--call' implementation lumps stdout and
+    ;; stderr together, so we must parse the combined output instead.
```
progfolio, line 2075: Thank you for bringing this to my attention. I agree it would be useful to distinguish stderr from stdout. Working on a patch now.

straight.el
pdbrown: Ah yeah, thanks for the heads up: I'd originally been parsing the warning message, but opted to look for any extra rev-parse output instead. I don't think we need to override the locale just to read ref names, so I got rid of these below.

straight.el
pdbrown: Oh nice! I'd considered it briefly but this change had become big enough, so I went with the workaround for now.

raxod502: Unfortunately, this might be harder than you think. I don't believe there's any non-hacky way to get stderr separately without migrating to the asynchronous process invocation API (`make-process`), and it turns out to be extremely tricky to just say "wait synchronously for the process to finish".

I've tried a number of times in https://github.com/raxod502/straight.el/issues/334.

raxod502:
This is greatly appreciated and much-needed; a lot of that code is a bit of a mess :flushed: 
Thank you!

pdbrown:
Sure thing -- I'm happy to pitch in where helpful, and thanks for the reviews!

progfolio:
Should take care of #354,too

straight.el
progfolio: The method I ended up implementing uses `call-process` synchronously. The `destination` argument may be a list, `(REAL-BUFFER STDERR-FILE)`. See: #714

straight.el
```diff
@@ -2037,139 +2063,271 @@ LOCAL-REPO is a string."
...
-If REF is non-nil, instead ensure that BRANCH is ahead of REF.
-Any untracked files created by checkout will be deleted without
-confirmation, so this function should only be run after
-`straight-vc-git--ensure-worktree' has passed."
+(defun straight-vc-git--abbrev-ref (ref)
+  "Return the unambiguous, abbreviated name of REF."
+  ;; We run git rev-parse twice here to more reliably detect
+  ;; ambiguous refs. When ref is ambiguous, rev-parse exits
+  ;; successfully and prints a warning to stderr. Ideally, we'd look
+  ;; for exit status 0 and non-empty stderr to detect this case, but
```
progfolio, line 2071: psuedo-code illustration of what this would look like once #714 is merged:

```emacs-lisp
(straight--process-with-result
    (straight--process-run "git" "rev-parse" "--verify" ref)
  (if (and invoked stderr)
      (error "Failed to parse-rev %S\n%s" ref stderr)
    ;;happy case
    ))
```

straight.el
```diff
@@ -1991,25 +1993,49 @@ to satisfy `straight-vc-git--urls-compatible-p'."
...
+handle cases we don't try to here. For example, magit's version does
+something more complicated to locate the git directory from Emacs
+running under cygwin."
+  (let ((default-directory straight--default-directory))
+    (or (when (fboundp 'magit-rebase-in-progress-p)
+          (magit-rebase-in-progress-p))
+        (let ((git-dir (file-name-as-directory
+                        (string-remove-suffix
+                         "\n"
+                         (straight--get-call "git" "rev-parse" "--git-dir")))))
```
progfolio, line 2008: #714 Has been merged into develop. You'll want to replace calls to `straight--get-call` with `straight--process-output`. The output will have the leading/trailing whitespace trimmed unless you bind `straight--process-trim` to nil. e.g. to get the old behavior of `straight--get-call-raw`:

```emacs-lisp
(let ((straight--process-trim nil))
  (straight--process-output "program" "arg"))
```

straight.el
progfolio: #714 merged. Should be able to distinguish stderr from stdout as noted in the previous comment.

progfolio (changes requested):
You'll want to rebase your changes on top of the develop branch and replace any of the old process function calls with the new ones. If you any help or have any questions, let me know.
Thank you.

straight.el
```diff
@@ -1753,29 +1750,28 @@ appropriately."
...
-;; clone Magit here.
-(defvar straight--profile-cache)
-
-(cl-defun straight--magit-status (directory)
-  "Like `magit-status', but install Magit if necessary.
-Magit is only installed if the user responds to a `y-or-n-p'
-prompt. Return non-nil if Magit was installed. DIRECTORY is as in
-`magit-status'."
-  (unless (or (require 'magit nil 'noerror)
-              (gethash "magit" straight--profile-cache))
```
pdbrown, line 1768: One thing to note here is that this `straight--profile-cache` lookup was in pretty old code and I couldn't tell what it was for: if the `require` fails, how would this help? I got rid of it in the new version, but can adjust it if I missed something.

straight.el
```diff
@@ -2042,131 +2060,265 @@ LOCAL-REPO is a string."
...
+           ,(lambda ()
+              (when (and (straight--ensure-magit-p 'magit-log)
+                         (fboundp 'magit-log-setup-buffer)
+                         (fboundp 'magit-log-arguments))
+                (let ((args (car (magit-log-arguments)))
+                      (default-directory straight--default-directory))
+                  (when (and diverged
+                             (not (member "--left-right" args)))
+                    (setq args (cons "--left-right" args)))
+                  (magit-log-setup-buffer (list range) args nil)
```
pdbrown, line 2256: This is new since the last update, `magit-log` is pretty great!

pdbrown (commented):
Awesome! Thanks for the heads up and sorry about the delay, but here's an updated, squashed change.

straight.el
raxod502: Great question. The intent was to see if the user had already pulled in Magit via `straight-use-package`, and so requesting it again would have no effect. Of course, as you observe, checking for this condition won't actually help anything, so I don't think there any reason to keep the check around. I couldn't tell you exactly what was going through my mind at the time.

pdbrown:
I rebased again to run with the new tests, and to fix the small merge conflict. I'd changed `(require 'cl-lib)` to `(require 'cl-macs)`, but reverted back to `cl-lib`, since that was/is working fine.

progfolio (approved):
Looks good to me. Thanks again for this!

pdbrown:
No problem and thanks again for all the reviews!

PR #714, opened 2021-03-09
Refactor external process API

progfolio:
`straight--process-call`:
 wrapper for call-process which returns a list with exitcode, stdout, stderr

`straight--process-with-result`:
 anaphoric macro which provides common bindings for results returned
 by `straight--process-call`

`straight--process-run`:
 wrapper for `straight--process-call` which logs and warns. Logging and
 warnings are controlled by the variables `straight-process-log` and
 `straight-process-warn`. The default behavior is to log and not warn.

`straight--process-run-p`:
  Predicate which runs process and returns t if it exited with 0.

`straight--process-output`:
  Runs process and returns its output.
  The output is trimmed if the `straight--process-trim` variable is non-nil (defaults to t).

Add command and error faces to `straight-process-buffer` logs.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -105,6 +105,21 @@ They are still logged to the *Messages* buffer.")))
   :group 'applications
   :prefix "straight-")
 
+(defgroup straight-faces nil
+  "Faces used in straight.el."
+  :group 'straight
+  :group 'faces)
+
+(defface straight-process-error
+  '((t (:weight bold :foreground "red")))
+  "Face for process errors in `straight-process-buffer'."
+  :group 'straight-faces)
+
+(defface straight-process-command
+  '((t (:weight bold)))
+  "Face for process commands in `straight-process-buffer'."
+  :group 'straight-faces)
+
 (defcustom straight-arrow
   (if (char-displayable-p ?→) " → " " -> ")
   "The string to use for an arrow in messages."
@@ -817,7 +832,7 @@ be interpreted later as a symlink."
     (condition-case _
         (if straight-use-symlinks
             (if (straight--windows-os-p)
-                (straight--get-call
+                (straight--process-output
                  "cmd" "/c" "mklink"
                  (subst-char-in-string ?/ ?\\ link-name)
                  (subst-char-in-string ?/ ?\\ link-target))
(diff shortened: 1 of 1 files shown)


progfolio:
A few examples:

    ;;(0 "hi " nil)
    (straight--process-run "bash" "-c" "echo hi")
    
    ;;(0 nil "hi")
    (straight--process-run "bash" "-c" "echo hi 1>&2")
    
    ;;(2 nil "hi")
    (straight--process-run "bash" "-c" "echo hi 1>&2; exit 2")

produces a `straight-process-buffer` which looks like this on my machine:

image

`straight--process-with-result` makes for (IMO) readable code when working with the results.
A simplified example:

    (straight--process-with-result
        (straight--process-call "foo")
      (if invoked
          (message (if success stdout stderr))
        (error "process error: %s" result)))

I've run tests successfully using `straight-bug-report` which install a number of packages.
Everything seems to be working in that regard, but seeing as this has touched a lot of the codebase, more testing is in order.

Thoughts?

straight.el
```diff
@@ -885,138 +878,155 @@ against the wrong repositories.
 If you set this globally to something other than nil, you may be
 eaten by a grue.")
 
+(defconst straight--process-stderr
+  (expand-file-name "straight-stderr" temporary-file-directory)
```
pdbrown, line 882: Hard-coding this path might result in concurrent emacs/straight processes trampling each others' results. Could we use `make-temp-file` (+ unwind-protected cleanup in `straight--process-call`) instead?

straight.el
```diff
@@ -885,138 +878,155 @@ against the wrong repositories.
...
+           (with-current-buffer
+               (find-file-noselect straight--process-stderr
+                                   'nowarn 'raw)
+             (prog1
+                 (let ((s (buffer-string)))
+                   (unless (string-empty-p s) s))
+               (kill-buffer)))))
+      (error e))))
+
+(defmacro straight--process-with-result (result &rest body)
```
pdbrown, line 919: This is neat! I haven't spent a ton of time writing elisp and would have done this by making result a plist and destructuring that at call sites, or maybe with a plain alist if it were outside of straight. I'd use :exit :stdout :stderr and an :error or :failure key hold any invocation error. I think the anaphoric bindings are a bit less obvious to read/use at the call site, but would say I'm neutral overall.

pdbrown (commented):
This is great! I left 2 minor notes inline, but it'll definitely improve error/warning checks in #711.

straight.el
progfolio: > Hard-coding this path might result in concurrent emacs/straight processes trampling each others' results

Good observation. We should add some information to make the path unique per Emacs instance (we don't have to worry about concurrent processes within a single instance as we're still using a synchronous model. If/when we move to something asynchronous a good portion of this code will need to be rewritten anyhow.)

> Could we use make-temp-file (+ unwind-protected cleanup in straight--process-call) instead?

I'll append `(emacs-pid)` to the path when setting it.

straight.el
progfolio: >  I think the anaphoric bindings are a bit less obvious to read/use at the call site, but would say I'm neutral overall.

Yes, as always, there's a tradeoff.

The anaphoric macro in this case works well, though. The DSL for the result ensures consistent naming and logic across call sites. This could also be achieved with functions, but most of those would end up being more boilerplate/documentation than they're worth, IMO. (e.g. process--exited-zero-p "Return t if process exited with zero"...and then a simple `(zerop (car process))`. So we get the benefit of encapsulating the logic without all that boilerplate.

straight.el
```diff
@@ -885,138 +878,155 @@ against the wrong repositories.
...
+RESULT must be an expression which evaluates to a list of form:
+  (EXITCODE STDOUT STDERR)
+Anaphroic bindings provided:
+  result: the raw process result list
+  exit: the exit code of the process
+  invoked: t if process executed without an elisp error
+  success: t if process exited with exit code 0
+  failure: t if process did not invoke or exited with a nonzero code
+  stdout: output of stdout
+  stderr: output of stderr"
```
raxod502, line 930: This is great! It's much like `straight-vc-git--destructure`.

straight.el
```diff
@@ -105,6 +105,21 @@ They are still logged to the *Messages* buffer.")))
   :group 'applications
   :prefix "straight-")
 
+(defgroup straight-faces nil
+  "Faces used in straight.el"
```
raxod502, line 109:
```suggestion
  "Faces used in straight.el."
```

raxod502 (commented):
Nice! I like the addition of colors.

Issue #713, opened 2021-03-08
#708 broke Org recipe on Mac with ZSH


progfolio:
FYI, the change in #708 has broken my ability to build org on emacs on a mac with ZSH as my shell because of how ZSH handles globbing.

Perhaps `straight-recipes-org-elpa--build` could be a windows-only operation?

Core issue:
```
(let ((shell-file-name "/bin/zsh")
      (default-directory "~/.emacs.d/straight/repos/org"))
  (prin1 (shell-command-to-string "git describe --match release\* --abbrev=0 HEAD"))
  (prin1 (shell-command-to-string "noglob git describe --match release\* --abbrev=0 HEAD"))
  (setq shell-file-name "/bin/bash")
  (prin1 (shell-command-to-string "git describe --match release\* --abbrev=0 HEAD"))
  nil
  )
"zsh:1: no matches found: release*
""release_9.4.4
""release_9.4.4
```

_Originally posted by @cwhatley in https://github.com/raxod502/straight.el/issues/707#issuecomment-792812751_

progfolio:
@cwhatley: Thanks for bringing this to my attention. I will look into it ASAP. 

> Perhaps straight-recipes-org-elpa--build could be a windows-only operation?

If the issue is with ZSH and globbing, Windows users may still be affected by this same bug.
I may have made an error when translating that command over from Org's makefiles.
I'll do some testing and see if I can work out a solution that is operating system independent and works with bash/zsh.

In the meantime, you can override the `:pre-build` steps in your init with the previous steps as a workaround:

```emacs-lisp
(straight-use-package
 '(org-plus-contrib
   :type git
   :repo "https://code.orgmode.org/bzg/org-mode.git"
   :local-repo "org"
   ;; `org-version' depends on repository tags
   :depth 'full
   :pre-build ("make" "autoloads")
   ;; Org's make autoloads generates org-version.el
   :build '(:not autoloads)
   :files ( :defaults
            "lisp/*.el"
            ("etc/styles/" "etc/styles/*")
            "contrib/lisp/*.el")
   :includes org))
```

cwhatley:
I think you copied the command over just fine. Since building org by hand creates org-version just fine with the same command, I think the core difference is that GNU make uses /bin/sh unless you define a SHELL in your makefile.

https://www.gnu.org/software/make/manual/make.html#Recipes

[...]

PR #719, opened 2021-03-14
[#715] Remove unnecessary backslash

raxod502:
This wasn't doing anything even before https://github.com/raxod502/straight.el/pull/715, which was probably the reason for https://github.com/raxod502/straight.el/issues/713 in the first place. The backslash needs to be doubled `\\` to be passed through to the actual shell. But now that we're not using a shell we don't need any escaping.


progfolio:
Good catch! Thank you.

Issue #505, opened 2020-04-12
Allow users to specify a custom lookup table for git domains


VojtechStep:
Right now, AFAICT, the domain resolution for different `:host`s has a hardcoded dictionary of mappings from a host to a domain. I would like to be able to change this mapping.

I use ssh for interacting with remote git repositories, but since I am a lazy guy, I don't really feel like always typing in the 'github.com' part. So I use my `~/.ssh/config` to resolve `git@gh` as `git@github.com` with the correct identity, and I do this for multiple hosts.
As a result, all my github remotes are of the form `git@gh:user/repo`.

I would like to be able to give this information somehow to straight.el.

ATM, I'm using advices on the functions `straight-vc-git--{decode,encode}-url` and it works well for the most part, except for the very crucial task of installing straight.el itself, since I can't advice functions that aren't defined yet (AFAIK), so during install, straight tries to clone itself via `git@github.com`, which doesn't have an associated ssh identity and fails.

Feel free to close this issue if it's too niche of a use case. I have no problem with setting the default protocol to https for the first time running a fresh install of Emacs, or using any other workaround that I can come up with :smile:.

raxod502:
This was actually a planned feature that I worked on a while ago, but never finished. I pushed the work in progress to a temporary branch in case you or somebody else would like to finish the feature.

VojtechStep:
I might take a look at it.

hristoast:
I'm interested to know what's unfinished for this. I found this issue after realizing I couldn't use a git.sr.ht repo as a source, I'd be willing to try and wrap up whatever work is needed to get this merged.

VojtechStep:
You can use any git repo, you just have to set `:host` to `nil` and then specify the full URL in `:repo`.

As for integrating custom host symbols, I think @raxod502's linked commit should pretty much work, so I too would like to hear from him why he considers it unfinished.

VojtechStep:
Looking into it more, it seems that:
1) The spec->url encoding helpers are there, but they are not used in `straight-vc-git--encode-url` proper
2) The url->spec decoding is not implemented

AFAICT, the decoding is used in two places:
[...]

PR #722, opened 2021-03-20
Clarify usage for the git backend's :host and :repo keywords

hristoast:
As we discussed in #505 I've added a bit more clarification about these keywords. I wasn't sure if the code snippets fit but I decided to just go with them anyways. Let me know if you'd prefer this PR not have them!

### DIFF
--- a/README.md
+++ b/README.md
@@ -2141,8 +2141,23 @@ These are the keywords meaningful for the `git` backend:
 
 * `:repo`: the clone URL for the Git repository.
 * `:host`: either nil or one of the symbols `github`, `gitlab`,
-  `bitbucket`. If non-nil, then `:repo` should just be a string
-  "username/repo", and the URL is constructed automatically.
+  `bitbucket`.
+    * If nil, then `:repo` should be a string which is the full URL of
+    the target repository. For example:
+
+    ```emacs-lisp
+    ( :package "package" :host nil :type git
+      :repo "http://myhost.tld/repo")
+    ```
+
+    * If non-nil, then `:repo` should be a string "username/repo",
+    and the URL is constructed automatically.  For example:
+
+    ```emacs-lisp
+    ( :package "package" :host github :type git
+      :repo "username/repo")
+    ```
+
 * `:branch`: the name of the branch used for primary development, as a
   string. If your version lockfiles do not specify a commit to check
   out when the repository is cloned, then this branch is checked out,


README.md
```diff
@@ -2141,8 +2141,19 @@ These are the keywords meaningful for the `git` backend:
...
 * `:repo`: the clone URL for the Git repository.
 * `:host`: either nil or one of the symbols `github`, `gitlab`,
-  `bitbucket`. If non-nil, then `:repo` should just be a string
-  "username/repo", and the URL is constructed automatically.
+  `bitbucket`.
+    * If nil, then `:repo` should be a string which is the full URL of
+    the target repository. For example:
+
+            ( :package "package" :host nil :type git
+              :repo "http://myhost.tld/repo")
```
progfolio, line 2149: Could you please change these examples to use the fenced codeblock syntax?
e.g.
     ```emacs-lisp
    (example... ;etc
    )
    ```

progfolio:
Don't worry about those two CI failures. They're due to the concurrency limit with our CI service.

progfolio:
Thanks again.

raxod502:
Huh... is CircleCI reporting build failures just because of the concurrency limit? If so, that seems pretty bad: I would expect it to simply take longer to report a result.

If this behavior happens on a regular basis, we might want to try out GitHub Actions, which has worked well for me in the past and has been simple to set up.

progfolio:
I believe what happens is the job is started concurrently, then when the limit is hit the job fails due to a network timeout e.g.:

```
error pulling image configuration: Get https://production.cloudflare.docker.com/registry-v2/docker/registry/v2/blobs/sha256/fa/fa24704bae4cb348ddffd7af0137f9e3755504f2ad4b0965f88a69b04d23f22e/data?verify=1616250992-AkGSZTb1X2qfDTmOeeiFr2YvtI0%3D: net/http: TLS handshake timeout
```

raxod502:
I am pretty sure that is not an error on the side of CircleCI. I have seen the concurrency limit warning you mentioned, and I am pretty sure it just means that some of the jobs were queued before being executed. That error message seems to indicate a configuration error on Docker Hub's side. Even if we were hitting the Docker Hub free tier rate limit which was instituted recently, we should be getting an API error back, not a TLS handshake timeout. If it happens on a regular basis, and this error hasn't been well-discussed online, I'd reach out to Docker Hub, to be honest.

Issue #643, opened 2020-12-03
gccemacs update - solving "emacs-version-changed"


VanLaser:
On my Arch Linux machine, I use an "emacs native compilation branch" package (from the `archlinuxcn` repository), and I recently switched to `straight.el` for package management ("develop" branch). Every time my emacs package gets updated, `straight.el` bails out with 'emacs-version-changed'. This, I found out to be caused by natively compiled files left over from the before-the-update emacs binary, in the `"~/.emacs.d/eln-cache"` folder (see 'comp-eln-load-path' variable doc). So, I fixed it using a slightly changed version of the bootstrap sequence recommended on straight.el homepage, as follows, to delete that folder (the bottom part):

```elisp

(setq load-prefer-newer t)
(setq straight-repository-branch "develop"
      straight-vc-git-default-clone-depth 1)

;;; package manager bootstrap
(defvar bootstrap-version)
(let ((bootstrap-file
       (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
      (bootstrap-version 5))
  (unless (file-exists-p bootstrap-file)
    (with-current-buffer
        (url-retrieve-synchronously
         "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
         'silent 'inhibit-cookies)
      (goto-char (point-max))
      (eval-print-last-sexp)))
  ;; catch emacs updates that have native compiled leftovers
  (unless (catch 'emacs-version-changed
	    (load bootstrap-file nil 'nomessage))
    (when (boundp 'comp-eln-load-path)
      ;; remove leftovers, with confirmation just to be safe
      (when (yes-or-no-p (format "Delete '%s'?" (car comp-eln-load-path)))
	(delete-directory (expand-file-name (car comp-eln-load-path)) t))
      ;; and try again
      (load bootstrap-file nil 'nomessage))))
```

Now (finally!) the question: is that the solution? In other words, who's responsible for dealing with that error: a) should straight.el not check for that mismatch, b) is it ok if I do it there, or c) should the package installer remove that folder for me (but, since it's customizable from my init.el, how could it know what folder to delete, in this case)?

I know 'straight.el' is actually checking for compiled files mismatch, but the native compilation is adding an extra layer of 'compiled' in the mix.

------

[...]

PR #729, opened 2021-03-28
Bootstrap attempts to load straight eln then elc

### LINKED ISSUE #697: throw/catch for bytecode detection does not work on nativecomp branch
## What's wrong

I've been trying out the nativecomp branch and I use straight.el as my package manager. It's very possible this is an Emacs bug, but when reporting it to emacs-devel they suggested I report it here to try and narrow it down.

One of the things that it does is throw when it detects that its own compiled bytecode has changed. It does this from within straight.el:

https://github.com/raxod502/straight.el/blob/2d407bccd9378f1d5218f8ba2ae85c6be73fbaf1/straight.el#L45-L49

It then catches within bootstrap.el, which loads straight.el:

https://github.com/raxod502/straight.el/blob/2d407bccd9378f1d5218f8ba2ae85c6be73fbaf1/bootstrap.el#L47-L54

Unfortunately, this doesn't work with nativecomp. It claims that there's no matching catch when a throw actually happens.

```
Warning (initialization): An error occurred while loading
‘/Users/aaronjensen/.emacs.d/early-init.el’:

No catch for tag: emacs-version-changed, nil
```

## Directions to reproduce

Build emacs with nativecomp. Run Emacs with an init.el containing just the bootstrap code. Watch the async compilation and wait for it to at least compile `straight.el`. Once that's done, rebuild emacs (you can just `touch xdisp.c`) and run emacs again.

### Version information

* Emacs version: 28.0.50 nativecomp
* Operating system: macOS

### LINKED ISSUE #724: Straight `bootstrap` trigger native compiling on every launch of Emacs.
Straight `bootstrap` trigger native compiling `straight.el` on every launch of Emacs.
I've traced it down to this line

https://github.com/raxod502/straight.el/blob/e1390a933b6f5a15079d6dec91eac97a17aad10c/bootstrap.el#L46

```
Compiling /Users/.../.config/emacs/straight/repos/straight.el/straight.el...
Compilation finished.
```

```emacs-lisp
(straight-bug-report)
```
</details>

- Test run at: `2021-03-22 15:34:21`
- system-type: `darwin`
- straight-version: `prerelease (grafted, HEAD -> develop, origin/develop) 770bbc6 2021-03-21`
- emacs-version: `GNU Emacs 28.0.50 (build 1, x86_64-apple-darwin20.3.0, NS appkit-2022.30 Version 11.2.1 (Build 20D74))
 of 2021-03-21`

```emacs-lisp
Bootstrapping straight.el...
Looking for emacsmirror-mirror recipe → Cloning el-get...
Looking for emacsmirror-mirror recipe → Cloning el-get...done
Building straight...
Building straight...done

Test run with version: prerelease (HEAD -> develop, origin/develop) 770bbc6 2021-03-21

```
</details>

aaronjensen:
Works around #724, should effectively fix #697 and #643 as well (though most of that fix was upstream)

This allows Emacs to first attempt to load eln, then elc. It will fall back to el if neither exist, but the `byte-recompile-file` should eliminate that possibility.

Theoretically this is "slower" because it does a few more file existence checks based on `load-suffixes`, but in practice it's unlikely to be noticeable for one file and it eliminates the need to check for native compilation support and load that explicitly.

I still need to do a little looking into the file modification check to make sure it actually fixes #697 and #643

### DIFF
--- a/bootstrap.el
+++ b/bootstrap.el
@@ -43,9 +43,9 @@
         ;; `emacs-version-changed' if the version of Emacs has changed
         ;; since the last time it was byte-compiled. This prevents us
         ;; from accidentally loading invalid byte-code, hopefully.
-        (load (expand-file-name (concat straight.el "c")
-                                default-directory)
-              nil 'nomessage 'nosuffix)
+        (load (file-name-sans-extension
+               (expand-file-name straight.el default-directory))
+              nil 'nomessage)
         (setq emacs-version-changed nil))
       (when emacs-version-changed
         ;; In safe mode, sacrifice performance for safety.
@@ -54,9 +54,9 @@
           ;; Don't use the optional LOAD argument for
           ;; `byte-compile-file' because it emits a message.
           (byte-compile-file straight.el)
-          (load (expand-file-name (concat straight.el "c")
-                                  default-directory)
-                nil 'nomessage 'nosuffix))))))
+          (load (file-name-sans-extension
+                 (expand-file-name straight.el default-directory))
+                nil 'nomessage))))))
 
 ;; This assures the byte-compiler that we know what we are doing when
 ;; we reference functions and variables from straight.el below. It


aaronjensen:
This change should be good to go. It does appear to fix #697 and #643 as far as I can tell.

The real fix for #724 is on its way upstream (though this change here is still likely correct for this as we don't actually want to explicitly load the elc).

The file modification stuff appears to work on the latest master for me.

progfolio:
Apologies for the delay, Aaron- and thanks again for this pull!

Issue #727, opened 2021-03-24
straight-prune-build-cache: print messages when pruning


progfolio:
OK, after tracking how `straight-check-for-modifications` was being used, I figured out why the rebuilding is happening in my case. I am sorry to say it was completely my fault. 

My `use-package` statement was in a file which was lazy loaded and my Radian config was calling `straight-prune-build-cache` when emacs starts. This caused the build commands to be deleted from the build cache and a rebuild to happen the next time the file is loaded and the `use-package` statement is executed. 

I think printing a message when a package is pruned would be pertinent, considering how easy it is to run into such a situation while everything seemingly "works".

_Originally posted by @haji-ali in https://github.com/raxod502/straight.el/issues/712#issuecomment-805267715_

PR #734, opened 2021-04-08
straight-prune-build-cache: print message when pruning

progfolio:
See #727


haji-ali:
Thanks @progfolio!
I have learned more about straight and saw that another function is `straight-prune-build-directory` which is even more destructive than `straight-prune-build-cache` (especially if the wrong directory is being deleted for some reason). Do you think a similar message should be added there as well?

progfolio:
> Do you think a similar message should be added there as well?

Added in #735

PR #737, opened 2021-04-17
Fix magit-log-arguments working directory

pdbrown:
Here's a small fix to the new magit-log feature. We need to set the `default-directory` around `magit-log-arguments` in addition to `magit-log-setup-buffer`.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -2252,13 +2252,18 @@ next time around."
               (when (and (straight--ensure-magit-p 'magit-log)
                          (fboundp 'magit-log-setup-buffer)
                          (fboundp 'magit-log-arguments))
-                (let ((args (car (magit-log-arguments)))
-                      (default-directory straight--default-directory))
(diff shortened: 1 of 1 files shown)


raxod502 (commented):
I think that we only want to bind `default-directory` around the invocations of `magit-log-arguments` and `magit-log-setup-buffer` specifically. If we bind it around `straight--recursive-edit`, then it screws up various things with buffer-local variables unfortunately---hence the existence of `straight--default-directory`, which _is_ safe to bind around `straight--recursive-edit`, and which is then handled lower down by the process invocation functions.

pdbrown:
Ah ok! I pushed the recursive edit out of `default-directory`'s scope and left a note.

progfolio (approved):
This looks good to me. Thank you.

raxod502 (approved):
LGTM :sparkles:

pdbrown:
No problem, thanks for the reviews!

PR #754, opened 2021-05-05
Fix/full repo uri with non nil host

progfolio:


progfolio:
Most forges dont allow  `:` in the repo name.
While this is a little naive, I believe this solution will catch the common case we're bumping into.
Let me know what you think.

raxod502:
Yeah, that looks great!

PR #770, opened 2021-05-30
Monitor correct directory for changes

raxod502:
It's unclear to me how this was working before, but I noticed that my modifications were not being detected, investigated, and discovered that the filesystem watcher was being started in `~/.emacs.d/straight/watcher`, when it needed to be watching `~/.emacs.d/straight/repos`:

https://github.com/raxod502/straight.el/blob/b4effa92f82c60fe4e3d97e0f2ba3db473ae164c/straight.el#L4215

Notably, if we extend the watchexec command line like in this pull request but with multiple `-w` arguments, then we can properly support the filesystem watcher for repositories with custom `:local-repo` paths (outside of `~/.emacs.d/straight/repos`, that is).

### DIFF
--- a/watcher/straight_watch.py
+++ b/watcher/straight_watch.py
@@ -47,7 +47,16 @@ def write_process_data(pid_file):
 def start_watch(repos_dir, modified_dir):
     callback_cmd = [CALLBACK_SCRIPT, repos_dir, modified_dir]
     callback_sh = " ".join(map(shlex.quote, callback_cmd))
-    cmd = ["watchexec", "--no-vcs-ignore", "-p", "-d", "100", callback_sh]
+    cmd = [
+        "watchexec",
+        "--no-vcs-ignore",
+        "-p",
+        "-d",
+        "100",
+        "-w",
+        repos_dir,
+        callback_sh,
+    ]
     cmd_sh = " ".join(map(shlex.quote, cmd))
     print("$ " + cmd_sh, file=sys.stderr)
     subprocess.run(cmd, cwd=repos_dir, check=True)


raxod502:
I suspect I might be misunderstanding something, because taken at face value this PR suggests that filesystem watching was never working, which isn't true.

progfolio:
It looks like the watcher is watching the correct directory on my system:

```shell
λ pgrep -a watchexec
143116 watchexec --no-vcs-ignore -p -d 100 /home/n/.emacs.d/straight/repos/straight.el/watcher/straight_watch_callback.py /home/n/.emacs.d/straight/repos /home/n/.emacs.d/straight/modified
```

I'll take a look at this soon to see what could be going on.

watcher/straight_watch.py
```diff
@@ -47,7 +47,16 @@ def write_process_data(pid_file):
...
     callback_sh = " ".join(map(shlex.quote, callback_cmd))
-    cmd = ["watchexec", "--no-vcs-ignore", "-p", "-d", "100", callback_sh]
+    cmd = [
+        "watchexec",
+        "--no-vcs-ignore",
+        "-p",
+        "-d",
+        "100",
+        "-w",
+        repos_dir,
```
progfolio, line 57: Explicitly watching the repos_dir makes sense to me.

raxod502:
> It looks like the watcher is watching the correct directory on my system

Ah! But you actually can't see what directory is being watched from that command line. It shows you the command-line arguments that were passed, but not the working directory of the process. And if you don't explicitly pass `-w` to watchexec, then it watches the working directory.

At least if I understand correctly.

progfolio:
Haven't looked too closely at the watcher script, but I was under the impression that the subprocess for wathexec was started here:

```python
def start_watch(repos_dir, modified_dir):
    callback_cmd = [CALLBACK_SCRIPT, repos_dir, modified_dir]
    callback_sh = " ".join(map(shlex.quote, callback_cmd))
    cmd = ["watchexec", "--no-vcs-ignore", "-p", "-d", "100", callback_sh]
    cmd_sh = " ".join(map(shlex.quote, cmd))
    print("$ " + cmd_sh, file=sys.stderr)
    subprocess.run(cmd, cwd=repos_dir, check=True)
    ```
    
    Which sets cwd=repos_dir.
    Is that correct?

raxod502:
Oh, thank you for the pointer---I had totally missed that setting when reviewing the code. I will revert the change in my local repo and, if the problem happens again, will track it down properly.

PR #775, opened 2021-06-01
Remove advice to load org-contrib as a means to load Org

chrikoehn:
`org-plus-contrib` was removed from the README (#774) but replacing `org-plus-contrib` with `org-contrib` does not make sense in this case because org-contrib does not contain Org.


progfolio:
Thank you

Issue #503, opened 2020-04-09
Add programmatic interface for dependency information


aGolduck:
How can I find out which packages depend on a specific package? In general, is there a way to inspect dependency information programmatically?

raxod502:
There is no way to do this using the current API, which is considered a bug. Here is how you can do it using the internal functions:

```elisp
(nth 1 (gethash "magit" straight--build-cache))
```

wpchou:
Actually I mean, how to find out packages which are dependent on some installed package. Am I speak it wrong? English is not my mother tongue.

raxod502:
Sorry! What you said in your first comment was correct, but I misread it because people have asked much more frequently about dependencies rather than dependents. There's still no public API, but you can try this:

```elisp
(cl-remove-if-not
 (lambda (package)
   (member "dash" (nth 1 (gethash package straight--build-cache))))
 (hash-table-keys straight--recipe-cache))
```

wpchou:
It works. I was trying to find out packages that introduce some installed package. Thanks!

progfolio:
> It works. I was trying to find out packages that introduce some installed package. Thanks!

You should be able to use `straight-dependents` now to do this. See: #787 for details.

raxod502:
Awesome, that seems like a great improvement!

aGolduck:
Thank you. @progfolio

PR #787, opened 2021-06-09
Add straight-dependencies, straight-dependents

progfolio:
See: #503

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -7005,6 +7005,65 @@ locally bound plist, straight-bug-report-args."
                         (delete-directory ,temp-emacs-dir 'recursive)))))
        (message "Testing straight.el in directory: %s"
                 ,temp-emacs-dir))))
+
+;;;; Dependency Info
+
+;;;###autoload
+(defun straight-dependencies (&optional package)
+  "Return a list of PACKAGE's dependencies."
+  (interactive (list
+                (straight--select-package
+                 "Dependencies of"
+                 ;; Only offer candidates which have dependencies.
+                 (lambda (recipe)
+                   (and (straight--installed-p recipe)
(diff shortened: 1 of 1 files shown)


progfolio:
\`straight-dependencies\` recursively constructs a list of dependencies for a package like so:

    (straight-dependencies "helm")

(async popup (helm-core async))

Each element is a direct dependencies of the package or a list of the direct dependency and transitive dependencies.
I thought about making transitive dependencies optional, but it's trivial to flatten the list for now. e.g.:

    (delete-dups (flatten-tree (straight-dependencies "helm")))

(async popup helm-core)

\`straight-dependents\` does the opposite. Each member of the returned list depends on the package or is a list of
packages which transitively depend on it.

    (straight-dependents "epl")

    ((pkg-info (projectile helm-projectile) (flycheck flycheck-package)))
     ((annalist evil-collection) general elfeed-org esup (epl (pkg-info (projectile helm-projectile) (flycheck flycheck-package))) (package-lint flycheck-package) doom-themes helm-describe-modes (flx helm-flx) helm-projectile hydra js2-mode (avy link-hint ace-window) macrostep (evil evil-anzu evil-collection) (alert mu4e-alert) (esxml nov) org-mime shackle tern yasnippet)

We can easily walk this tree to see that \`helm-projectile\` and \`flycheck-package\` transitively depend on \`epl\` through \`projectile\` and \`flycheck\` which both require pkg-info.

We could use these as a base to draw more user-friendly dependency graphs/trees in the future.

raxod502:
Should we document this in the README?

progfolio:
Sure, I think that could be useful info. I'd also like to write a dependency visualizer that's easier to read eventually.

Issue #780, opened 2021-06-04
Always rebuild when filename contains unicode.


nasyxx:
## What's wrong

`straight` always rebuilds when the filename contains Unicode.

## Directions to reproduce

A directory and a file:

```
;;; ~/.emacs.d/测试/测试.el
(defun 測試 () t)
(provide '测试)
```

```
;;; ~/.emacs.d/init.el

(defvar bootstrap-version)
(let ((bootstrap-file
       (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
      (bootstrap-version 5))
  (unless (file-exists-p bootstrap-file)
    (with-current-buffer
        (url-retrieve-synchronously
         "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
         'silent 'inhibit-cookies)
      (goto-char (point-max))
      (eval-print-last-sexp)))
  (load bootstrap-file nil 'nomessage))

(straight-use-package
  `(测试 :local-repo ,(locate-user-emacs-file "测试/")))
```

While starting emacs, the `测试` will always rebuild.

<img width="486" alt="Screen Shot 2021-06-04 at 06 04 27 AM" src="https://user-images.githubusercontent.com/5007613/120785035-bad01800-c4fa-11eb-8b8e-fb013bd4fcad.png">

In fact, I'm not sure if using a file name with non-ASCII chars will cause problems.  While I tried without `straight` and found no problems, still not sure if it is an issue of `straight.`

### Version information

* Emacs version: `GNU Emacs 28.0.50 (build 1, x86_64-apple-darwin20.4.0, NS appkit-2022.44 Version 11.3.1 (Build 20E241)) of 2021-05-24`
* Operating system: macOS 11.3.1

raxod502:
Can confirm, and it's definitely a bug with `straight.el`. I bet one of the places where I used `insert-file-contents-literally` is to blame.

progfolio:
>  I bet one of the places where I used insert-file-contents-literally is to blame.

This sounds like a good place to investigate.
At first glance, my guess would be where we load the build-cache.

```emacs-lisp
(insert-file-contents-ilterally (straight--build-cache-file))
```

progfolio:
@nasyxx: Could you try evaluating the following test case in your \*scratch* buffer?
It will install straight.el with a patch from my fork in a temporary environment and attempt to install your package.
[...]

PR #794, opened 2021-06-13
[#780] Don't insert file contents literally

raxod502:
I was able to reproduce both issues reported in https://github.com/raxod502/straight.el/issues/780 (modifications always detected, and text decoding errors), and with this change, neither occur anymore.


progfolio:
Excellent. This sidesteps the rabbit hole I was going down trying to figure out how to properly encode/decode while still inserting literally.

PR #801, opened 2021-06-16
Remove straight--normalize-alist

progfolio:
Function is only called during `straight-expand-files-directive`.
We can use `cl-pushnew` instead.

I believe this will be equivalent. Seem OK to you @raxod502?

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -313,20 +313,6 @@ Backward compatible shim for `flatten-tree'."
 
 ;;;;; Association lists
 
-(defun straight--normalize-alist (alist &optional test)
-  "Return copy of ALIST with duplicate keys removed.
-The value for a duplicated key will be the last one in ALIST.
-Duplicates are tested with TEST, which must be accepted by the
-`make-hash-table' function and which defaults to `eq'. The order
-of the entries that are kept will be the same as in ALIST."
-  (let ((hash (make-hash-table :test (or test #'eq)))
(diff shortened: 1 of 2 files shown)


straight.el
```diff
@@ -4763,15 +4749,16 @@ other value, the behavior is not specified."
...
+    (let ((copy nil)
+          (absolutes (mapcar (lambda (mapping)
+                               (cl-destructuring-bind (src . dest) mapping
+                                 ;; Make the paths absolute.
+                                 (cons (expand-file-name src src-dir)
+                                       (expand-file-name dest dest-dir))))
+                             mappings)))
+      ;; If there are duplicate keys, only keep the last declared cell.
+      (dolist (el (nreverse absolutes) copy)
+        (cl-pushnew el copy :test (lambda (a b) (equal (car a) (car b))))))))
```
raxod502, line 4761: I think you can use `:test #'equal :key #'car` for this.

raxod502 (approved):
Yeah, looks good to me. I didn't review the logic closely but the approach seems correct. It'll be a little slower, but I doubt that will have a substantive impact on anything since this isn't on the critical path during startup and the size of the lists is very small.

straight.el
progfolio: Ah, good catch! I always forget about cl-libs key projections.

PR #812, opened 2021-07-06
straight--build-compile: properly set load-path

progfolio:
Previous method involved looking up each package's dependencies.
It did not account for local-repo names and transitive dependencies.
Now we just recursively add straight's build dir to the load-path
when byte-compiling. This should cover all cases.


raxod502:
Hmmm... that seems like it would _almost always_ work, but doesn't it technically make the environment a bit nondeterministic on account of random old things lying around in `~/.emacs.d/straight/build` could be added to the `load-path`?

PR #825, opened 2021-07-24
Auto custom types

raxod502:
@progfolio I think this ought to be a lot easier for you to maintain than https://github.com/raxod502/straight.el/pull/798 going forward :)

Basically, I just used the built-in Customize functions to get a list of all the user options, retrieve their defined `:type`s, and validate the standard values against the types automatically. That way there's no need to update tests when modifying user options. While doing this work, I found a few small bugs:

* `defgroup` declarations were in the wrong order, so most of the `straight.el` user options were listed in the `straight-faces` group (it defaults to whichever `defgroup` appeared most recently, in the absence of `:group` on the `defcustom`).
* `straight-byte-compilation-buffer` was using the nonexistent `or` keyword instead of `choice`.
* `straight-recipes-gnu-elpa-ignored-packages` was using `list` when it should have been using `repeat`.


progfolio:
Yes, this looks like a much better approach to me!

progfolio (approved):

raxod502:
Awesome.

PR #862, opened 2021-10-16
Add tests to CI

progfolio:
Switch test suite to use ert.el.
This prevents us from having to bootstrap Buttercup during CI.

Add `straight-deftest` macro for defining tests via templates.


raxod502:
Amazing!! Now that it has been quite some time since I worked directly on `straight.el`, I am seeing how big of an issue it is that I didn't write any tests in the first place. Thank you for cleaning up my mess!

PR #874, opened 2021-10-29
Completion functions and annotations

roshanshariff:
This patch modifies `straight--select-package` and `straight--get-recipe` to use completion functions instead of tables. The completion function allows for
* Setting the completion category to `'straight-recipe`, which is useful for packages like Embark.
* Showing extra metadata when the completion UI supports annotations (see screenshot below, using Vertico).

image

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -5281,36 +5281,6 @@ RECIPE is a straight.el-style plist."
 ;;;; Interactive helpers
 ;;;;; Package selection
 
-(defconst straight--recipe-completion-metadata
-  `((category . straight-recipe))
-  "Metadata to be passed to `completing-read' when selecting packages.")
-
-(defun straight--recipe-completion (recipes)
-  "Completion function for recipe names listed in RECIPES."
-  (lambda (string pred action)
-    (pcase action
-      ('metadata
-       (cons 'metadata straight--recipe-completion-metadata))
-      (_
-       (complete-with-action action recipes string pred)))))
-
-(defun straight--cached-recipe-completion (&optional filter)
-  "Completion function for cached recipe names.
-Candidates are the keys of `straight--recipe-cache'.
-
-See documentation of `straight--select-package' for a description
-of FILTER."
-  (lambda (string pred action)
-    (pcase action
-      ('metadata
-       (cons 'metadata straight--recipe-completion-metadata))
-      (_
-       (complete-with-action
-        action straight--recipe-cache string
-        (lambda (package recipe)
-          (and (or (null filter) (funcall filter recipe))
-               (or (null pred) (funcall pred package)))))))))
-
 (defun straight--select-package (message &optional filter)
   "Use `completing-read' to select a package.
 MESSAGE is displayed as the prompt; it should not end in punctuation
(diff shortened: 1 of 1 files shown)


straight.el
```diff
@@ -5281,6 +5281,54 @@ RECIPE is a straight.el-style plist."
 ;;;; Interactive helpers
 ;;;;; Package selection
 
+(defconst straight--recipe-completion-metadata
+  `((category . straight-recipe)
+    (annotation-function . straight--recipe-annotation-function))
```
progfolio, line 5286:
```suggestion
  '((category . straight-recipe)
    (annotation-function . straight--recipe-annotation-function))
```
No need to backquote the list if nothing is being evaluated inside it.

straight.el
```diff
@@ -5281,6 +5281,54 @@ RECIPE is a straight.el-style plist."
...
+  "Metadata to be passed to `completing-read' when selecting packages.")
+
+(defun straight--recipe-annotation-function (package)
+  "Information about PACKAGE to be displayed for completion."
```
progfolio, line 5290:
```suggestion
  "Return PACKAGE information to be displayed during completion."
```

straight.el
```diff
@@ -5281,6 +5281,54 @@ RECIPE is a straight.el-style plist."
...
+  (let* ((source (straight-recipe-source package))
+         (recipe (gethash package straight--recipe-cache))
+         (localrepo (and recipe
+                         (straight--repository-is-available-p recipe)
+                         (plist-get recipe :local-repo)))
+         (annot (concat
+                 (propertize " " 'display '(space :align-to 40))
+                 (if source (format " %s" source))
```
progfolio, line 5298:
```suggestion
                 (when source (format " %s" source))
```
Prefer `when` over `if` if there is no `else` branch.

straight.el
```diff
@@ -5281,6 +5281,54 @@ RECIPE is a straight.el-style plist."
...
+                 (propertize " " 'display '(space :align-to 60))
```
progfolio, line 5299: Should the `:align-to` values be hard coded? What happens when the completion window isn't wide enough to accommodate these values? I assume they will wrap.

straight.el
```diff
@@ -5281,6 +5281,54 @@ RECIPE is a straight.el-style plist."
...
+                 (if localrepo (if (string= localrepo package)
+                                   " (installed)"
+                                 (format " (in %s)" localrepo))
+                   localrepo))))
```
progfolio, line 5303: Similar to above. Instead of using if, we can use `when`:

```emacs-lisp
(when localrepo
  (format " (%s %s)"
          (if (string= localrepo package) "installed" "in")
          package))
```

straight.el
```diff
@@ -5281,6 +5281,54 @@ RECIPE is a straight.el-style plist."
...
+    (propertize annot 'face 'completions-annotations)))
+
+(defun straight--recipe-completion (recipes)
+  "Completion function for recipe names listed in RECIPES."
+  (lambda (string pred action)
+    (pcase action
+      ('metadata
+       (cons 'metadata straight--recipe-completion-metadata))
+      (_
+       (complete-with-action action recipes string pred)))))
```
progfolio, line 5313: Does this need to be a `pcase`? At present we're just checking whether or not action is eq to 'metadata.
We could stick with a simple `if` replace it later if we need to support more actions.

straight.el
```diff
@@ -5281,6 +5281,54 @@ RECIPE is a straight.el-style plist."
...
+
+(defun straight--cached-recipe-completion (&optional filter)
+  "Completion function for cached recipe names.
+Candidates are the keys of `straight--recipe-cache'.
+
+See documentation of `straight--select-package' for a description
+of FILTER."
+  (lambda (string pred action)
+    (pcase action
```
progfolio, line 5322: Same as above. Is the `pcase` necessary?

straight.el
```diff
@@ -5474,24 +5516,25 @@ action, just return it)."
                      straight-recipe-repositories
                      nil
                      'require-match)))))
-  (let ((sources (or sources straight-recipe-repositories)))
-    (let* ((package (intern
-                     (completing-read
-                      "Which recipe? "
-                      (straight-recipes-list sources)
-                      (lambda (_) t)
```
progfolio, line 5482: Is there a reason to remove `completing-read`'s PREDICATE argument in the replaced code?

progfolio (changes requested):
Thanks for the pull.
I've yet to try it locally, but I've left some general notes/questions from my cursory glance.
I will test this out later tonight and leave more if I notice anything else.

roshanshariff:
> I will test this out later tonight and leave more if I notice anything else.

Thanks for the helpful review @progfolio! I'm working on a newer version that fixes the style issues you've pointed out. I'm also improving the annotator function:
* If a package recipe is cached, it shows the source repository
* Otherwise it shows the recipe source (melpa, elpa, etc.)

You might want to wait for the new version before testing it, to save some time :)

straight.el
roshanshariff: A nil predicate behaves exactly like it is always true, so there's no need to create a lambda in this case.

straight.el
roshanshariff: The completion UIs I have seen truncate the annotation if it is too long to fit in one line, which is probably the desired behaviour. It is possibly to dynamically truncate the columns to fit the minibuffer width, but that is unnecessary in this case since it's okay if the right-most information gets hidden. No matter what, the package name itself will always be displayed fully.

I chose to reserve 40 characters for the package name to accommodate almost everything that's currently available. If a package has a longer name it will still be fully visible, but will push the second column to the right. I will adjust the second and third columns with the update I'm working on, but we can't really do much without changes in the Emacs completion-read protocol (that are being discussed, afaik).

roshanshariff:
I have pushed the new version with better annotations and improved coding style and documentation. Let me know what you think.

The completion for installed packages now looks like this: Note that the remote URL for "helpful" shown here is from my personal fork, not the original recipe. Remote URLs are truncated by Vertico if they don't fit on one line.
image

The completion for recipe-related functions looks like this: It now shows the upstream remote for "helpful".
image

straight.el
```diff
@@ -5290,16 +5290,121 @@ FILTER is a function accepting one argument: a straight style recipe plist.
...
+or built-in) and its remote repository."
+  (let ((annotfmt
+         (propertize
+          (concat (propertize " " 'display '(space :align-to 40)) " %s"
+                  (propertize " " 'display '(space :align-to 51)) " %s")
+          'face 'completions-annotations)))
+    (lambda (package)
+      (let* ((recipe (gethash package recipes))
+             (remote (straight--recipe-annotation-remote recipe)))
+        (straight--with-plist recipe (type)
```
progfolio, line 5359: If we're just binding the recipe's type, it would probably be more efficient to just use `plist-get` directly with the recipe instead of `straight--with-plist` and add the binding to the `let*`. e.g.

```emacs-lisp
(type (plist-get recipe :type)))
```

straight.el
```diff
@@ -5290,16 +5290,121 @@ FILTER is a function accepting one argument: a straight style recipe plist.
...
+  "Make an annotation function for keys of RECIPES.
+See documentation of `straight--package-completion' for a
+description of RECIPES.
+
+Each package is annotated with its installation status (installed
+or built-in) and its remote repository."
+  (let ((annotfmt
+         (propertize
+          (concat (propertize " " 'display '(space :align-to 40)) " %s"
+                  (propertize " " 'display '(space :align-to 51)) " %s")
```
progfolio, line 5354: :align-to 51 or 50?

straight.el
```diff
@@ -5290,16 +5290,121 @@ FILTER is a function accepting one argument: a straight style recipe plist.
...
+Returns a string, or nil if the remote repository could not be
+determined."
+  (ignore-errors
+    (straight--with-plist recipe (type)
+      (when (eq type 'git)
+        (straight-vc-git--destructure recipe (repo host)
+          (replace-regexp-in-string     ; Remove extraneous URL components...
+           (rx bos (? (*? anything) ":" (? "//")) ; prefix "protocol://"
+               (group (*? anything))              ; keep this part
+               (? ".git") eos)                   ; suffix ".git"
```
progfolio, line 5404: Instead of loading `rx` for this regexp, we can construct it via `concat` and annotate it similar to what you've done:

```emacs-lisp
(concat
 "\\`" ; beginning of string
 "\\(?:[^z-a]*?:\\(?://\\)?\\)?" ; prefix "protocol://"
 "\\([^z-a]*?\\)" ; captured text up to repo extension
 "\\(?:\\.git\\)?" ; ".git" extension
 "\\'" ; end of string
 )
```

progfolio (changes requested):
Had the chance to test this locally. Right now the biggest blocker is figuring out the formatting situation. I uploaded a few images in a review comment to give you an idea of what it looks like on my system (with and without helm). If we can get something uniform across completion frameworks, that would be ideal. If not, we could add this feature behind an opt-in defcustom (something like `straight-annotate-recipe-completions`).

Thanks again for your work on this. Let me know if you have any questions.

straight.el
roshanshariff: I chose 51 to leave two spaces after the string "installed" because that seemed to look the best :) But I will have to revisit this after we figure out how to do formatting properly.

straight.el
```diff
@@ -5290,16 +5290,121 @@ FILTER is a function accepting one argument: a straight style recipe plist.
...
+See documentation of `straight--package-completion' for a
+description of RECIPES.
+
+Each package is annotated with its installation status (installed
+or built-in) and its remote repository."
+  (let ((annotfmt
+         (propertize
+          (concat (propertize " " 'display '(space :align-to 40)) " %s"
+                  (propertize " " 'display '(space :align-to 51)) " %s")
+          'face 'completions-annotations)))
```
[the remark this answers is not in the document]
roshanshariff: Ah, I'd expected the annotation to be ignored in the first case, but I'll investigate this further. Maybe the solution is for the annotation-function to be a small string to be displayed alongside the candidate, whereas an additional affixation function can include the longer, column-separated string. It seems like completion UIs prefer an affixation function when available, falling back to annotation function otherwise. And from a quick glance, it looks like Helm only looks at annotations and not affixations.

I'll also have to look into what Helm is doing with the annotation strings on their way to being displayed. Thanks for pointing this out! If there isn't a good solution that works in all cases, I will think about the best way to make this functionality customizable, as you suggested.

progfolio:
Based off of lack of activity, I'm closing this pull request. If you are still interested in working on this, let me know and we can reopen. Thanks.

PR #883, opened 2021-11-17
straight-bug-report: Open test case details by default

progfolio:


raxod502:
Oh, cool, today I learned! I had never seen `<details open>` befoer.

Issue #844, opened 2021-09-18
Support :type nil?


progfolio:
Confirmed with the following:

[YODEL](https://github.com/progfolio/yodel) REPORT (2021-09-18 17:59:12):

<details> <summary>Test Case:</summary>

```emacs-lisp
(yodel
  :save t
  :user-dir "local-repo-test.straight"
  :formatter yodel-format-as-github-markdown
  :post*
  (yodel-file "./local-repo/local-repo.el"
    :save t
    :overwrite t
    :with*
    ";;;###autoload"
    (defun local-repo-test nil
      (message "PASS"))
    (provide "local-repo"))
  "Straight bootstrap code"
  (defvar bootstrap-version)
  (let
      ((bootstrap-file
        (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
       (bootstrap-version 5))
    (unless
        (file-exists-p bootstrap-file)
      (with-current-buffer
          (url-retrieve-synchronously
           (concat "https://raw.githubusercontent.com/" "raxod502/straight.el/develop/install.el")
           'silent 'inhibit-cookies)
        (goto-char
         (point-max))
        (eval-print-last-sexp)))
    (load bootstrap-file nil 'nomessage))
  (condition-case nil
      (message "Test run with version: %s"
               (straight-version))
    (error nil))
  "End bootstrap"
  (straight-use-package
   '(local-repo-test
     :local-repo "./local-repo"
     :type nil))
  (local-repo-test))
```

</details>
<details>
  <summary>STDOUT:</summary>

```emacs-lisp
Test run with version: prerelease (HEAD -> master, origin/master, origin/HEAD) 1e27b05 2021-06-13
Cloning ./local-repo (for local-repo-test)...
```

</details>

<details>
  <summary>STDERR:</summary>

```emacs-lisp
(error "No such VC backend ‘nil’")
```

</details>

<details>
  <summary>Environment</summary>

- **emacs version**: GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, X toolkit, cairo version 1.17.4, Xaw3d scroll bars)
 of 2021-09-16
- **system type**: gnu/linux

</details>

I've never used a recipe in this way. Indeed it does look like specifying `:type nil` will fail during `straight-vc`.
It looks like this behavior would require implementing a VC backend called "nil" according to:
19c34fc96b3668d2b1266cf18fb8008903c5c98b

@raxod502: Is this a use case we want to support?

[...]

PR #885, opened 2021-11-21
[#844] Support :type nil properly

raxod502:
Closes https://github.com/raxod502/straight.el/issues/844

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -1506,38 +1506,21 @@ This method simply delegates to the relevant
 
 ;;;;; No VC management
 
-
-(dolist (method (list "check-out-commit"
-                      "clone"
-                      "commit-present-p"
-                      "fetch-from-remote"
-                      "fetch-from-upstream"
-                      "get-commit"
-                      "keywords"
-                      "local-repo-name"
-                      "normalize"
-                      "merge-from-remote"
-                      "merge-from-upstream"
-                      "push-to-remote"))
-  (defalias (intern (format "straight-vc-nil-%s" method)) #'ignore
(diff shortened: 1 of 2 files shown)


straight.el
```diff
@@ -1504,6 +1504,24 @@ This method simply delegates to the relevant
...
+                      "get-commit"
+                      "keywords"
+                      "local-repo-name"
+                      "normalize"
+                      "merge-from-remote"
+                      "merge-from-upstream"
+                      "push-to-remote"))
+  (defalias (intern (format "straight-vc-nil-%s" method)) #'ignore
+    "Psuedo VC backend method for packages with :type nil."))
+
```
progfolio, line 1524: Maybe we could just combine the initialization of this with the `built-in` back end since they do the same thing.
Something like?:

```emacs-lisp
;;;;; Psuedo VC Backends

;; Recipes with :type built-in and :type nil should ignore VC operations.
(dolist (method '(check-out-commit
                  clone
                  commit-present-p
                  fetch-from-remote
                  fetch-from-upstream
                  get-commit
                  keywords
                  local-repo-name
                  merge-from-remote
                  merge-from-upstream
                  normalize
                  push-to-remote))
  (dolist (backend '(built-in nil))
    (defalias (intern (format "straight-vc-%s-%s" backend method)) #'ignore
      (format "Psuedo VC backend method for packages with :type %s." backend))))
```

progfolio (approved):

straight.el
raxod502: Yeah, for sure, let me do that now.

Issue #671, opened 2021-01-08
Documentation length


titibandit:
I would suggest to organize the documentation differently.
Having everything in this one super massive readme can be really daunting for people not accustomed to straight.el
Then the package appears as being very complex when in reality, much of what's the readme, is not documentation per say. (comparisons with other package manager for example)

I've thought of this when I saw Denis Ogbe here (https://ogbe.net/emacs/emacs.html), detailing this (very interesting!) configuration for emacs. 
I quote:
"There are alternatives to the built-in package.el, for example straight.el. It looks like a great project. But the manual is incredibly long, it seems like an incredibly "different" way to handle packages, and I am frankly intimidated by it's apparent complexity."

Which is the shame, since the guy seem clearly able to handle straight.el.

Maybe organize the documentation with several markdown documents? with the wiki?
Just a suggestion that could avoid turning people off because of the appearance the length of the documentation gives to the project.
Cheers!

progfolio:
Thanks for bringing this to our attention.
I agree we could simplify the README and move some of the other information to a secondary site or document(s). Will discuss with @raxod502

raxod502:
You aren't the first person to point out this problem. I think you are correct that having multiple sub-pages is better than having one gigantic large page. I think it would be nice to keep the documentation in Markdown format in this repository, although we could certainly make it auto-publish to GitHub Pages for even easier accessibility.

4aHxKzD:
I saw a post on reddit regarding this issue, and just wanted to say that the long format of @raxod502's READMEs is one of the reasons why I like to use his packages. Other Emacs packages make me search information all across the web (Github, reddit, EmacsWiki, Blog posts, etc.), but with packages like straight and selectrum I know that all the relevant information is right there in a single file.

progfolio:
https://github.com/raxod502/straight.el/issues/827#issue-953473692

Explicitly mention, provide example usage of `:type` nil.

martenlienen:
[...]

PR #906, opened 2021-12-19
Improve local-repo docs

axelson:
From my initial reading of the docs (as a relative straight.el newbie) I didn't realize that local-repo was meant for working with files on the local file system, I assumed that it was something that you would use with your own git repository server.

I hope that by being more specific about what a "local repository" is that it will be easier for others to understand how to use the option.

I hope my understanding of what `local-repo` does is good, if not then hopefully we can still improve the wording while remaining accurate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1890,12 +1890,13 @@ meaning in a recipe (unknown keywords are ignored but preserved):
 
 * `:local-repo`
 
-  This is the name of the local repository (either a git directory on
-  the local file system or a plain directory on the local file system)
-  that is used for the package. If a local repository by that name
-  does not exist when you invoke `straight-use-package`, one will be
-  cloned according to the package's [version-control
-  settings][#user/recipes/vc-backends].
+  This is the name of the local repository that is used for the
+  package. If a local repository by that name does not exist when you
+  invoke `straight-use-package`, one will be cloned according to the
+  package's [version-control settings][#user/recipes/vc-backends].
+  Note that the repository does not have to be version-controlled; it
+  can also just be a directory containing Emacs Lisp files (set `:type
+  nil` in this case).
 
   Multiple packages can use the same local repository. If so, then a
   change to the local repository will cause both packages to be
@@ -1906,13 +1907,6 @@ meaning in a recipe (unknown keywords are ignored but preserved):
   derived from the [version-control
   settings][#user/recipes/vc-backends], or as a last resort the
   package name.
-
-  For example:
-
-  ```emacs-lisp
-  (straight-use-package
-   '( example :local-repo "/home/user/repos/example.el" ))
-  ```
 
 * `:files`
 
--- a/README.md
+++ b/README.md
@@ -1890,10 +1890,12 @@ meaning in a recipe (unknown keywords are ignored but preserved):
 
 * `:local-repo`
 
-  This is the name of the local repository that is used for the
-  package. If a local repository by that name does not exist when you
-  invoke `straight-use-package`, one will be cloned according to the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1890,10 +1890,12 @@ meaning in a recipe (unknown keywords are ignored but preserved):
 
 * `:local-repo`
 
-  This is the name of the local repository that is used for the
-  package. If a local repository by that name does not exist when you
-  invoke `straight-use-package`, one will be cloned according to the
-  package's [version-control settings][#user/recipes/vc-backends].
+  This is the name of the local repository (either a git directory on
```
progfolio, line 1893: Not sure about the phrases "git directory" and "plain directory" here.
A "git directory", if I understand your intention, is just a repository.
A "plain directory" is just a directory. Those terms would probably be better. Keep in mind that other version control back ends may be implemented, so it's best to avoid referring to git directly when it's unnecessary.

README.md
```diff
@@ -1905,6 +1907,13 @@ meaning in a recipe (unknown keywords are ignored but preserved):
   settings][#user/recipes/vc-backends], or as a last resort the
   package name.
 
+  For example:
+
+  ```emacs-lisp
+  (straight-use-package
+   '( example :local-repo "/home/user/repos/example.el" ))
```
progfolio, line 1914: The text you added above mentions that `:local-repo` may be a local repository or directory, but the example is referring directly to an elisp file. Unless the directory is named "example.el", in which case the file path should end in a forward slash. It would be better to pick a name for the directory which didn't include the ".el" extension to avoid confusion, too.

Does this example work if you run it?

progfolio:
Apologies for the confusion. The documentation could use an overhaul in general (see #671). 
While the current README is thorough, I feel it misses what many users are actually looking for: "How do I do X with straight?"
Ideally, what you're referring to here would fall under "How do I manage locally hosted elisp with straight.el?"

You have the right idea. You'll want to point the `:local-repo` at the directory containing the elisp.
If that directory is not under version control, you would want to add `:type nil` to prevent straight from attempting to manage it as if it were under version control. To add to your example:

```emacs-lisp
(straight-use-package '(my-local-package :local-repo "/elisp/my-local-package/" :type nil) ;;directory non under version control
```
Until the new documentation is written, we would probably be better to adding this use-case to the FAQ.
Does that work for you?

README.md
axelson: I got tripped up initially because repository can mean many different things in computer science so when I was initially reading through the list of options I wasn't sure what `repository` meant in this context. 

Maybe this would be more straightforward: "This is the path to a local vcs repository (git by default)"

Although since with `:type nil` it doesn't have to actually be a vcs repository so maybe something could be added to that effect.

README.md
axelson: Yeah I intended it to be a directory named `example.el`, but I agree that is confusing. I haven't run straight.el stand-alone so I'm not 100% sure this example would work on it's own.

axelson:
I think that a workflow-based section of the README would be helpful, although I don't think it would supplant the reference portion (although perhaps the reference portion could shrink a bit).

I don't think I would have noticed this if it was in the FAQ unless it showed up when I did ctrl-f but I'm not entirely sure what the best keywords for that would be. I think I missed the FAQ completely when I was initially looking, including the FAQ link at the very top of the README. One issue with the FAQ as it stands is that in order to see the index of questions in the FAQ you need to scroll down 3 full pages. One way to help that would be to insert a table of contents at the beginning of the FAQ section although that might be a maintenance burden to keep up to date.

Another idea could be to tidy up the Getting Started section and link a few workflows (in the FAQ) from there.

PR #926, opened 2022-02-06
Use emacs invocation name

emacs18:
When launching emacs sub-process to build org package, use emacs that is currently running rather than "emacs" found in current PATH.  Why?  Because emacs in current path my be different than currently running emacs.  For example /usr/bin/emacs is 24.3, but I use my own build of emacs 28 which is launched via wrapper shell script.
When emacs 28 tries to build org package, then it ends up launching emacs 24 as sub-process which cannot deal with modern org mode.


emacs18:
Sorry, I must have made mistake in selecting incorrect branch name.  My change should be one line change!

Issue #893, opened 2021-12-02
Rebuild happens when opens a different version emacs


feng-qi:
I recently want to try emacs with tree-sitter support. When I run the newly built emacs, the rebuild starts. This rebuild also happens when some time ago I want to see if a feature is supported by a older version emacs.

And it will rebuild again when I switch back to the emacs version I normally use.

This is very time consuming. I don't know if there's already a method to handle this situation in straight. If not, can we add some facilities to handle this better by, for example, multiple build directories?

Thanks.

progfolio:
Duplicate of https://github.com/raxod502/straight.el/issues/484

Please see discussion there.

feng-qi:
Thanks. This comment https://github.com/raxod502/straight.el/issues/484#issuecomment-927365350 does help me a lot. The following simple setting seems to meet all my needs.

```elisp
(setq straight-build-dir (concat "build-" emacs-version))
```

PR #931, opened 2022-02-25
Support running multiple versions of Emacs against the same install

rdparker:
### Use Case
My home directory is on a network share and different systems have different versions of Emacs installed. Some of these are out of my control.

### Proposed Solution

Add the `straight-use-version-specific-build-dir` option to enable running multiple versions of Emacs against the same
`user-emacs-directory`.


progfolio:
The reason an error is thrown is because elisp byte-compiled using one version of Emacs is not guaranteed to work on another version.

progfolio:
The solution here does not work for your use case?: https://github.com/raxod502/straight.el/issues/893

rdparker:
Assuming this:
```emacs-lisp
(setq straight-build-dir (concat "build-" emacs-version))
```

is the solution you meant, no it does not. That only works on versions of Emacs without native compilation.

I had to retest it to make sure my memory was correct and that some version of Emacs generates an `.elc` in the `straight\repos\straight.el` directory regardless of the value of `straight-build-dir`. When I tested with a 29.0.50 build from the Emacs git repo, it wrote a version-specific `straight.elc` into the  straight repo directory. I believe the 28.0.91 pre-release does the same, but don't have it installed on this computer for testing right now.

Why does this happen?

In order for native compilation to work the `.elc` *must* be in the same directory as the source `.el` so that gccemacs can find the source to generate the `.eln`, while it is possible to coerce Emacs into placing byte-compiled files elsewhere, see [Peeve](https://github.com/rdparker/peeve/blob/master/peeve.el), this is fragile. This is due to "the assumption that the source and compiled files are found in the same directory is hard-coded in various places in Emacs", as the `byte-compile-dest-file-function` doc string mentions.

The part of my PR that touches `bootstrap.el` addresses this by suppressing byte-compilation of `straight.el` during bootstrap just as setting `straight-safe-mode` does. Note, this only supresses the generation of this specific `.elc` not the `straight.elc` that gets generated later in the build directory.

I would also suggest that an option is a far better UX than having to search GitHub issues for how to get this working.

rdparker:
I just noticed the `straight.elc` is still generated, but the `bootstrap.el` change prevents it from being read by Emacs that do not have native compilation enabled.  I probably need to test that two native-comp versions work with this patch or if additional changes are needed, since it appears the file may still be generated.

raxod502:
This thread is being closed automatically by [Tidier](https://github.com/radian-software/tidier) because it is labeled with "waiting on response" and has not seen any activity for 90 days. But don't worry—if you have any information that might advance the discussion, leave a comment and I will be happy to reopen the thread :)

raxod502:
I'm sorry, I think this thread was closed erroneously by my bot. Reopening for consideration

kngwyu:
Is there any change for this PR to be merged? I have multiple containers with Emacs on my HPC cluster account, where multiple versions of Emacs can use the same `~/.emacs.d/` directory. So this patch would really help.

raxod502:
Let me see if I can get this running properly and merge it.

raxod502:
Thanks, and I'm very sorry for the delay.

kngwyu:
Thanks :)

PR #937, opened 2022-03-27
Watcher path concatenation is corrected

aaronjensen:
Without this, I am getting modified files in the modified directory without a slash between parts: `nano-themenano-theme.el`, for example, rather than `nano-theme/nano-theme.el`. I am on `watchexec 1.18.6`.


progfolio:
Thanks, Aaron. This looks reasonable to me.

PR #957, opened 2022-06-09
Rephrase README that native-comp is now in emacs/master

meliache:
The readme mentioned that native-compilation is under development under the native-comp branch of emacs. However, this feature has already been merged into master and will be available in emacs 29 as far as I understand. However, even though it's merged, the `--with-native-compilation`  flag is disabled by default.


raxod502 (approved):
:sparkles: Very good, thanks.

PR #967, opened 2022-06-23
fix install.el script

ayrat555:
Currently, package freezing fails because of the repo in the install.el script is not correct

Screenshot 2022-06-23 at 07 43 31


ayrat555:
it seems it's resolved in the develop branch

PR #971, opened 2022-06-30
Replaced all raxod502 with radian-software in README.md

akawama:
I got an error complaining 'origin' url has radian-software but recipe url has raxod502. So replaced raxod502 with radian-software throughout  README.md.


progfolio:
Pull requests should be based off the develop branch. However, this has already been addressed there.

PR #972, opened 2022-07-02
Use correct repo link for install instructions in README

t4sk3:


progfolio:
Pull requests should be made against the "develop"  branch. However, this has already been changed there.

Issue #985, opened 2022-08-11
autoload.el deprecated in newest Emacs development version


andersjohansson:
This is just a heads up that autoload.el seems to be in the process of being replaced by loaddefs-gen.el, which was introduced in the last months (starting from Emacs commit 1d4e9034178)

For straight.el this would probably involve invoking `loaddefs-generate` (introduced in loaddefs-gen.el) instead of `make-directory-autoloads` or `update-directory-autoloads` in `straight--build-autoloads`. Maybe a check for `(fboundp 'loaddefs-generate)` and calling that instead would be enough, but I am not sure whether it is a drop in replacement.

andersjohansson:
I will try out my changes in #986.

progfolio:
Thanks for bringing this to our attention.
Was there a discussion of this on emacs-devel? I'm curious to learn more about this change.

andersjohansson:
I actually didn’t find anything on emacs-devel, may be in some bug report, otherwise most information I found was in the commits from Lars and notes in NEWS.

progfolio:
Considering this closed as #986 was merged. Thanks again, @andersjohansson.

PR #986, opened 2022-08-11
Update for Emacs 29 where loaddefs-gen.el replaces autoloads.el

andersjohansson:
Fixes issue #985.

I believe this may be what is needed, but I haven’t tested it extensively yet.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -5098,14 +5098,15 @@ modifies the build folder, not the original repository."
   ;; it out, then straight.el will fail with a mysterious error and
   ;; then cause Emacs to segfault if you start it with --debug-init.
   ;; This happens because if you take out `eval-and-compile', then
-  ;; `autoload' will not be loaded at byte-compile time, and therefore
-  ;; `generated-autoload-file' is not defined as a variable. Thus
-  ;; Emacs generates bytecode corresponding to a lexical binding of
-  ;; `generated-autoload-file', and then chokes badly when
-  ;; `generated-autoload-file' turns into a dynamic variable at
-  ;; runtime.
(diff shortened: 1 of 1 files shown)


straight.el
```diff
@@ -5142,11 +5144,15 @@ modifies the build folder, not the original repository."
           ;; Actually generate the autoload file.
           ;; Emacs > 28.1 replaces `update-directory-autoloads' with
           ;; `make-directory-autoloads'
```
raxod502, line 5146: Probably want to extend this comment.

straight.el
```diff
@@ -5098,14 +5098,16 @@ modifies the build folder, not the original repository."
...
+  ;; time, and therefore `generated-autoload-file' is not defined as a
+  ;; variable. Thus Emacs generates bytecode corresponding to a
+  ;; lexical binding of `generated-autoload-file', and then chokes
+  ;; badly when `generated-autoload-file' turns into a dynamic
+  ;; variable at runtime.
   (eval-and-compile
-    (require 'autoload))
+    (condition-case nil
+        (require 'loaddefs-gen)
+      (error (require 'autoload))))
```
raxod502, line 5110: Might be able to use the `NOERROR` parameter of `require` here.

raxod502 (approved):
:sparkles: Thanks!

straight.el
andersjohansson: Done!

straight.el
andersjohansson: Ah, you’re right, I didn’t read the documentation for require well enough. That becomes a cleaner test.

raxod502:
Thanks again.

PR #1027, opened 2022-12-19
Do not test against Emacs master

raxod502:
cc @progfolio - trying to unbreak CI for https://github.com/radian-software/straight.el/pull/931. I've been doing this on other Emacs projects as well because I have found that testing against Emacs master introduces a lot of flakiness and generally imposes maintenance burden that doesn't seem to have a good return on investment given that we have limited resources.

In the medium-term future, I plan to make further changes to my Emacs projects to increase their maintainability and make it possible for me to do more active maintenance of all of them, including `straight.el`. I really appreciate that you already started effort in that direction by adding some tests. I should have done that in the first place, and I will likely add more in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        emacs_version: [25, 26, 27, "master"]
+        emacs_version: [25, 26, 27, 28]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,10 @@
 name: CI
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - develop
+      - master
+  pull_request: {}
 jobs:
   ci:
     runs-on: ubuntu-latest


progfolio:
> Emacs master introduces a lot of flakiness and generally imposes maintenance burden that doesn't seem to have a good return on investment given that we have limited resources.

Agreed. Testing stable releases sounds like a better plan.

> In the medium-term future, I plan to make further changes to my Emacs projects to increase their maintainability and make it possible for me to do more active maintenance of all of them, including straight.el. I really appreciate that you already started effort in that direction by adding some tests. I should have done that in the first place, and I will likely add more in the future.

That's great to hear. Straight should have a fair amount of coverage in unit tests. Testing/mocking up some of the more complex behavior would be nice, but I couldn't think of a clean way to do that.

raxod502:
I think what will need to be done is separating out some of the code into separate modules so that they can be tested individually, and then use a kind of dependency injection pattern to combine them into the full package. Refactoring will be needed, for sure. Right now, I'm focusing on https://github.com/radian-software/riju, because that one is so broken I cannot even compile it from source, much less deploy contributions safely. But `straight.el` might be the next one I look at.

progfolio:
> separating out some of the code into separate modules

This would be a great improvement for maintainability and development in general.

> Right now, I'm focusing on https://github.com/radian-software/riju

I don't think I mentioned this before, but that is a cool project. Nice work.

PR #1031, opened 2022-12-23
move indent.el to scripts subdirectory

progfolio:
Prevents load-path shadowing of Emacs built-in indent.el.

@raxod502 I don't think this will harm anything, but I just want to double check with you before merging.
On a similar note,  "install.el" and "bootstrap.el" should probably be renamed to "straight-install.el" and "straight-bootstrap.el" to prevent similar load-path shadowing of such generic names.


raxod502 (approved):
:sparkles: Sure, there is no problem with this. However, it's surprising to me that this is causing an issue - after all, the actual `straight.el` repository should not be added to `load-path` under normal operation; rather, a subset of its files should be symlinked to the build directory, which is what will be added to `load-path`:

```
Permissions Size User     Date Modified Name
.rw-rw-r--   17k raxod502 21 Dec 17:20  straight-autoloads.el
lrwxrwxrwx    77 raxod502 21 Dec 17:20  straight-ert-print-hack.el -> /home/raxod502/.emacs.d/straight/repos/straight.el/straight-ert-print-hack.el
.rw-rw-r--   500 raxod502 21 Dec 17:20  straight-ert-print-hack.elc
lrwxrwxrwx    64 raxod502 21 Dec 17:20  straight-x.el -> /home/raxod502/.emacs.d/straight/repos/straight.el/straight-x.el
.rw-rw-r--  6.1k raxod502 21 Dec 17:20  straight-x.elc
lrwxrwxrwx    62 raxod502 21 Dec 17:20  straight.el -> /home/raxod502/.emacs.d/straight/repos/straight.el/straight.el
.rw-rw-r--  257k raxod502 21 Dec 17:20  straight.elc
```

There is likewise nothing wrong in concept with renaming `install.el` and `bootstrap.el`, but backwards-compatibility considerations will have to be made for those, since code in users' configurations refers to both of these filenames explicitly.

progfolio:
> However, it's surprising to me that this is causing an issue - after all, the actual straight.el repository should not be added to load-path under normal operation;

You're right. I now use Elpaca to install straight. Specifying a `:files` directive similar to the one in bootstrap.el should be enough to solve this issue on my end. I think we can close this without merging.

raxod502:
Heh, nice. Speaking of which, looking forward to when Elpaca becomes the hot new package manager and I can make `straight.el` into a compatibility layer for it ;)

I'm going to start working on transitioning over to monitoring all inbound issues and pull requests again for `straight.el`, by the way, since volume has been pretty low on my other projects. Hopefully that will put me in a better position for driving foundational improvements a little down the line once I get Riju stabilized.

progfolio:
> Heh, nice. Speaking of which, looking forward to when Elpaca becomes the hot new package manager and I can make straight.el into a compatibility layer for it ;)

It's been a lot of fun trying to make it work. I plan on recording another video demo soon. I'd love to hear your thoughts when I do if you have time.

> I'm going to start working on transitioning over to monitoring all inbound issues and pull requests again for straight.el, by the way, since volume has been pretty low on my other projects. Hopefully that will put me in a better position for driving foundational improvements a little down the line once I get Riju stabilized.

That's good to hear! I'll still be chipping in when I can. If you have any issues you'd like to work on together, let me know.

PR #1034, opened 2022-12-26
straight--build-info: use configured info programs

mjoerg:


progfolio:
Thank you.  Out of curiosity, do you have either of the options customized?

progfolio:
Merged to "develop". Thanks again.

PR #1037, opened 2023-01-16
Fix parsing of nongnu-elpa recipes

raxod502:
Backwards compatibility so both the old and new formats work. See https://github.com/radian-software/straight.el/issues/1012, https://github.com/radian-software/straight.el/pull/1013.

This resolved `M-x straight-use-package` failing for me because I had an old checkout of NonGNU ELPA.


progfolio (approved):
LGTM

Issue #1024, opened 2022-12-12
document how to rebuild caches


aspiers:
I created a new recipe in my local fork of the `melpa` for the `lsp-bridge` package which is not yet in MELPA, and wanted to test installation from that recipe.  However straight did not notice the addition of that recipe, so I couldn't.  For example, `M-x straight-rebuild-recipe` doesn't accept `lsp-bridge` as a valid package name.

What's the correct procedure for achieving this?  I'm guessing this requires some cache rebuilding; if so, it would be great if this was implemented in a convenient user-facing interactive function, and documented accordingly.  Thanks!

raxod502:
Yep, it's definitely a valid request. The existing documentation has most of what you need, but it's not organized in a way that makes it easy to get started. (Sorry!)

> headings like "Where do recipes come from?" and "What does it mean to register a package?" which sound like they are written for people who mainly want to understand the internals of straight.el

Actually, that is intended to be the user facing part. You don't even want to know what the [developer facing part](https://github.com/radian-software/straight.el/issues/51) would look like ;)

Obviously, this can and should be improved, because having a full conceptual understanding of all of the pieces of your package manager should not be a requirement to use it effectively. This is something that I will address when I have time to focus primarily on `straight.el`; in the meantime, any improvements are welcome.

aspiers:
Thanks for the reply!  I would suggest considering _three_ target audiences, rather than two:

1. emacs users who just want to install and use packages
2. developers who want to write and/or improve recipes
3. developers who want to make `straight.el` even better

This issue focuses purely on 2., which I assume would be pretty easy to address since recipes are generally so simple and `straight.el` already has all the mechanics to handle them.

In contrast, #51 seems focused purely on 3., which seems like a much more complicated thing to document and understand :-)

aspiers:
I think I eventually found the answer I was looking for:

`M-x straight-use-package NEW-RECIPE` invokes `(straight--register-recipe recipe)` which puts a new entry for `NEW-RECIPE` into `straight--recipe-cache`.  Once it's in that cache, commands like `straight-rebuild-package` work fine.  So I guess the process for writing a new MELPA recipe (say) is something like:

- Create a new branch in `~/.emacs.d/straight/repos/melpa`
- Put the new recipe in `~/.emacs.d/straight/repos/melpa/recipes/NEW-RECIPE`
- `M-x straight-use-package NEW-RECIPE`
- Test it, and if any issues are found, fix them then jump back to the previous step
[...]

PR #1044, opened 2023-01-29
straight-get-recipe: change UNFILTERED to generic FILTER parameter

progfolio:
straight-use-package: FILTER out installed packages
See: #1024


raxod502 (approved):
:sparkles:

PR #1054, opened 2023-02-19
[#1053] Handle renamed native-comp variable in Emacs 30

### LINKED ISSUE #1053: Obsolete function "native-comp-deferred-compilation-deny-list"
## What's wrong

When starting emacs 30.0.50 with --debug-init I get 
Symbol's value as variable is void: native-comp-deferred-compilation-deny-list

To ensure normal operation, you should investigate and remove the
cause of the error in your initialization file.  Start Emacs with
the ‘--debug-init’ option to view a complete error backtrace.
 ■  Warning (comp): straight.el:5236:33: Warning: ‘native-comp-deferred-compilation-deny-list’ is an obsolete variable (as of 29.1); use ‘native-comp-jit-compilation-deny-list’ instead.
 ■  Warning (comp): straight.el:5234:19: Warning: ‘native-comp-deferred-compilation-deny-list’ is an obsolete variable (as of 29.1); use ‘native-comp-jit-compilation-deny-list’ instead.

If I change the two lines in straight.el as suggested, the error vanishes.

I think this is due to commit
0be5f7ab6368175953c0a5bcbbd485fd9edda2b0

[0be5f7ab6368175953c0a5bcbbd485fd9edda2b0](https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=0be5f7ab6368175953c0a5bcbbd485fd9edda2b0)

* Emacs version: 30.0.50 from git master
* Operating system: Arch Linux

raxod502:
Closes #1053

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -78,6 +78,7 @@ They are still logged to the *Messages* buffer.")))
 ;; `comp'
 (declare-function native-compile-async "comp.el")
 (defvar native-comp-deferred-compilation-deny-list)
+(defvar native-comp-jit-compilation-deny-list)
 
 ;; `finder-inf'
 (defvar package--builtins)
@@ -5250,9 +5251,12 @@ background after `straight-use-package' returns."
         (when (and straight--build-cache
                    (gethash package straight--build-cache))
           (let ((regexp (format "^%s" build-dir)))
-            (setq native-comp-deferred-compilation-deny-list
-                  (cl-remove-if (lambda (denied) (string= denied regexp))
-                                native-comp-deferred-compilation-deny-list))))
+            (dolist (list-var '(native-comp-deferred-compilation-deny-list
+                                native-comp-jit-compilation-deny-list))
+              (when (boundp list-var)
+                (set list-var
+                     (cl-remove-if (lambda (denied) (string= denied regexp))
+                                   (symbol-value list-var)))))))
         (let ((inhibit-message t)
               (message-log-max nil))
           (native-compile-async build-dir 'recursively))))))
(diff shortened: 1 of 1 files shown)


frumigoul:
Hi, when will this fix be merged into the Develop branch?
I tried to work from this commit, but somehow did not find the way to implement it in my init.el (sorry, emacs newbie here).
Thanks for the good work!

Panadestein:
Pardon the impatience, but I would like to second the previous comment. This PR is kind of a bottleneck in my config now. Thanks for the effort!

petergardfjall:
This PR solves the issue for me as well.

@j0ni: your fix with setting `straight-repository-branch` prior to running [the init snippet](https://github.com/radian-software/straight.el#getting-started) did not work for me.

petergardfjall:
> @petergardfjall I guess I should have mentioned that you would need to remove your existing straight installation subdir to force it to re-clone. My apologies.

Interestingly, when I tried your approach on an older version of Emacs (https://github.com/emacs-mirror/emacs/commit/fdac69b45e608f9f3610066ed873dacfd971119a) just deleting `straight/repos/straight.el/` and then restarting emacs worked fine: the `straight-repository-branch` setting was honored and the right `straight.el` branch was checked out.

However, when running on a more recent commit (on branch `emacs-29`) your proposed approach did not work and I had to check out the branch myself (as I hinted at in my [earlier comment](https://github.com/radian-software/straight.el/pull/1054#issuecomment-1437171407)). 

If I haven't completely messed up somehow it appears like the `install.el` script broke after Emacs changes introduced somewhere between [fdac69b..02aba20](https://github.com/emacs-mirror/emacs/compare/fdac69b45e608f9f3610066ed873dacfd971119a..02aba20d528f1ec45ad36424e8d65ace985dd04a).

zetaomegagon:
Thanks @j0ni worked perf for me :+1:

benthamite:
I can confirm that @petergardfjall's method works with Emacs 30:

1. `$ cd ~/.emacs.d/straight/repos/straight.el`
2. `$ git checkout --track origin/rr-fix-renamed-variable`
3.  Launch Emacs.

raxod502:
I unfortunately don't have enough free time to attend to PRs more often than every week or two. But you should be able to use any desired revision of `straight.el` by simply visiting `~/.emacs.d/straight/repos/straight.el` and checking out whatever you like.

benthamite:
Also worth noting that this fix won't work if you are on the "master" branch. Just set `straight-repository-branch` to `"develop"`.

zetaomegagon:
@raxod502 IMO, that's fine.* `straight.el` is great and I super duper appreciate your work. 🙏

*esp when I run off `emacs/master`. I could be running off a stable branch, but *I chose not to*.

PR #1064, opened 2023-03-04
Slight improvement to magic snippet

dabrahams:
Should work on Emacs  >= 23.1
I don't think this needs a changelog, but please LMK if you disagree.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1410,7 +1410,7 @@ care of all these details for you:
 ```emacs-lisp
 (defvar bootstrap-version)
 (let ((bootstrap-file
-       (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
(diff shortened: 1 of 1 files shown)


raxod502 (commented):
The minimum supported version of Emacs for `straight.el` [is Emacs 25.1](https://github.com/radian-software/straight.el/tree/3eca39dfc6797243ec7d1c6a7d45142407f73f88#getting-started). In particular, Emacs versions 24.3 and below were never supported, even in 2017 when the project was first begun. Does it change anything to add backwards compatibility here?

raxod502:
Oh, I see. You're not adding backwards compatibility, you're just using a more readable function instead of `expand-file-name`.

raxod502 (commented):
Ok, this looks good, but can you update also the other instances of the bootstrap snippet with the same change, and also bump `bootstrap-version` from 6 to 7?

PR #1068, opened 2023-03-15
Table versioning

arialdomartini:
The README paragraph about versioning lists some commands that, to me, are worth a table.

If you like the result, just comment this PR: I will complete it converting all the other bullet point lists to a HTML table.

Cheers.


raxod502:
Sure, that sounds good! CI will need to be updated to allow having long lines in the table, though, since it's an exception to the usual line-wrapping rules.

PR #1094, opened 2023-05-24
Fix argument error straight  not found

eamonnsullivan:


progfolio:
Thanks for pointing this out. I've pushed a fix on the latest "develop" branch which should take care of this.

PR #1095, opened 2023-05-27
Add proper custom host support

raxod502:
Support for custom entries in `straight-hosts` was previously a bit limited; there were various places in the code where specific built-in hosts had special cases or were hardcoded. This pull request changes the implementation to be fully generic, so that all features that were previously available for built-in hosts are now also available for custom ones. The way we do this is with a new format for the `straight-hosts` alist, but the old format is also supported for backwards compatibility (it's automatically translated to the new format if used).

This isn't quite done; I'm going to add a few more unit tests to exercise all the added and changed functionality, and also tweak the interface so that it's possible to configure the browser-friendly URL separately from the Git HTTPS URL.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -284,17 +284,67 @@ computes the fork as \"githubUser/fork\"."
                            (const :tag "bitbucket" bitbucket))
                 :value-type (string :tag "username")))
 
-(defcustom straight-hosts '((github "github.com" ".git")
-                            (gitlab "gitlab.com" ".git")
-                            (codeberg "codeberg.org" ".git")
-                            (sourcehut "git.sr.ht")
-                            (bitbucket "bitbucket.com" ".git"))
+(defun straight-host-like-github (name website)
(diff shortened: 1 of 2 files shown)


straight.el
```diff
@@ -1862,69 +1918,76 @@ edit. Otherwise, PROMPT and ACTIONS are as for
...
   (pcase host
     ('nil repo)
     ((pred (lambda (host) (alist-get host straight-hosts)))
      (when (string-match-p ":" repo)
        (error "Malformed protocol detected: (:host %S :repo %S)"
               host repo))
-     (let* ((spec (alist-get host straight-hosts))
-            (domain (car spec))
-            (suffix (cadr spec)))
-       ;;@FIX: It is sloppy to hardcode the sourcehut case here.
```
progfolio, line 1881: Glad to see this hardcoded sourcehut stuff go. Thank you!

PR #1100, opened 2023-06-13
Use insert-file-contents when reading process stderr

league:
The function `straight--process-call` was using `find-file-noselect` to read in the standard error after the process exits. This apparently runs a bunch of hooks and potentially turns on minor modes, which can be disruptive and slow. (Context of the particular issue I had is below.)

It seems like the recommended way to just read a file to a string is `with-temp-buffer` and `insert-file-contents`. See, for example, top answer to “[What's wrong with `find-file-noselect`?](https://emacs.stackexchange.com/a/2898/21511)” on stackexchange.

So I think this should be an invisible, work-alike improvement to an internal function.

Now, for my context: I'm using `undo-tree`, and by default, `undo-tree-auto-save-history` so that undo info is saved across killing and reopening the file or restarting Emacs. I also configured the recommend advice to gzip the undo-tree files, which by default is verbose.

With that configuration in place, whenever I ran a batch straight process such as `straight-freeze-versions` or `straight-normalize-all` or `straight-fetch-all`, the useful echo-area progress output of straight would be completely overwhelmed by repeated "compressing undo-tree" messages referencing the `straight-stderr-NNNN` file... something like this:

```
Processing repository "rainbow-mode"...
compressing .straight-stderr-1462415.~undo-tree~.gz...done
Wrote /tmp/.straight-stderr-1462415.~undo-tree~.gz
compressing .straight-stderr-1462415.~undo-tree~.gz...done
Wrote /tmp/.straight-stderr-1462415.~undo-tree~.gz
[...Above 2 lines repeat about 14 times...]
Processing repository "rainbow-mode"...done
Processing repository "nix-mode"...
compressing .straight-stderr-1462415.~undo-tree~.gz...done
Wrote /tmp/.straight-stderr-1462415.~undo-tree~.gz
[...14 times...]
Processing repository "nix-mode"...done
Processing repository "echo-bar.el"...
compressing .straight-stderr-1462415.~undo-tree~.gz...done
Wrote /tmp/.straight-stderr-1462415.~undo-tree~.gz
[...etc...]
```

Obviously, `undo-tree` is not relevant when straight is trying to read its stderr file after each process call. At first I pursued ways to prevent undo-tree from activating on this file. That seems harder than I expected – something like this was apparently not enough:
```elisp
  (add-to-list 'auto-mode-alist (cons "straight-stderr-" #'special-mode))
  (add-to-list 'undo-tree-incompatible-major-modes #'special-mode)
```

I thought about hacking `undo-tree` so it can prevent turning on by filename regexp and not just major-mode. But when I read the tips about `find-file-noselect` it seems to me that hacking straight as shown here is the Right Thing™ instead.  Thanks!

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -936,10 +936,11 @@ If the process is unable to start, return an elisp error object."
                 nil args)
          (let ((s (buffer-substring-no-properties (point-min) (point-max))))
            (unless (string-empty-p s) s))
-         (progn (insert-file-contents
-                 straight--process-stderr nil nil nil 'replace)
-                (let ((s (buffer-substring-no-properties (point-min) (point-max))))
-                  (unless (string-empty-p s) s)))))
+         (progn
+           (insert-file-contents
+            straight--process-stderr nil nil nil 'replace)
+           (let ((s (buffer-substring-no-properties (point-min) (point-max))))
+             (unless (string-empty-p s) s)))))
     (error e)))
 
 (defmacro straight--process-with-result (result &rest body)
--- a/straight.el
+++ b/straight.el
@@ -926,22 +926,21 @@ eaten by a grue.")
   "Run PROGRAM syncrhonously with ARGS.
 Return a list of form: (EXITCODE STDOUT STDERR).
 If the process is unable to start, return an elisp error object."
-  (let* ((program (if (string-match-p "/" program)
-                      (expand-file-name program)
-                    program)))
-    (condition-case e
-        (with-temp-buffer
-          (list
-           (apply #'call-process program nil
-                  (list (current-buffer) straight--process-stderr)
-                  nil args)
-           (let ((s (buffer-string)))
-             (unless (string-empty-p s) s))
-           (with-temp-buffer
-             (insert-file-contents straight--process-stderr)
-             (let ((s (buffer-string)))
-               (unless (string-empty-p s) s)))))
-      (error e))))
+  (when (string-match-p "/" program)
+    (setq program (expand-file-name program)))
+  (condition-case e
+      (with-temp-buffer
+        (list
+         (apply #'call-process program nil
+                (list t straight--process-stderr)
+                nil args)
+         (let ((s (buffer-substring-no-properties (point-min) (point-max))))
+           (unless (string-empty-p s) s))
+         (progn (insert-file-contents
+                 straight--process-stderr nil nil nil 'replace)
+                (let ((s (buffer-substring-no-properties (point-min) (point-max))))
+                  (unless (string-empty-p s) s)))))
+    (error e)))
 
 (defmacro straight--process-with-result (result &rest body)
   "Provide anaphoric RESULT bindings for duration of BODY.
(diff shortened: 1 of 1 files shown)


straight.el
```diff
@@ -937,12 +937,10 @@ If the process is unable to start, return an elisp error object."
...
-           (with-current-buffer
-               (find-file-noselect straight--process-stderr
-                                   'nowarn 'raw)
-             (prog1 (let ((s (buffer-string)))
-                      (unless (string-empty-p s) s))
-               (kill-buffer)))))
+           (with-temp-buffer
+             (insert-file-contents straight--process-stderr)
+             (let ((s (buffer-string)))
+               (unless (string-empty-p s) s)))))
```
progfolio, line 943:
```suggestion
           (with-temp-buffer
             (insert-file-contents straight--process-stderr)
             (let ((s (buffer-substring-no-properties (point-min) (point-max))))
               (unless (string-empty-p s) s)))))
```

Ignoring buffer string properties is about 3x faster and we don't need them in this case.

progfolio (changes requested):
Thanks for this. That should be a nice optimization. Testing locally with the single-buffer version of the function below, I see about a 2x improvement in performance compared to the current version of `straight--process-call`. This function is used all over the place, so really good catch.

straight.el
progfolio: We may also be able to get away with using a single buffer.
`insert-file-contents` has a `REPLACE` parameter which will replace the buffer's contents (which we have already stored in our return list).
Something like the following should work:

```emacs-lisp
(defun straight--process-call (program &rest args)
  "Run PROGRAM syncrhonously with ARGS.
Return a list of form: (EXITCODE STDOUT STDERR).
If the process is unable to start, return an elisp error object."
  (let* ((program (if (string-match-p "/" program)
                      (expand-file-name program)
                    program)))
    (condition-case e
        (with-temp-buffer
          (list
           (apply #'call-process program nil
                  (list (current-buffer) straight--process-stderr)
                  nil args)
           (let ((s (buffer-string))) (unless (string-empty-p s) s))
           (progn (insert-file-contents
                   straight--process-stderr nil nil nil 'replace)
                  (let ((s (buffer-string)))
                    (unless (string-empty-p s) s)))))
      (error e))))
```

league:
That looks great. I wondered about optimizing the nested `with-temp-buffer`, so the REPLACE is a great idea.  I'm surprised if `buffer-substring-no-properties` is making a difference, because there shouldn't be any properties (no sentinel manipulating the standard out). But indeed, if there were properties we wouldn't want them! I'll build and try with this, then add a commit, feel free to squash of course.

Should this have a changelog for an internal improvement?

progfolio (approved):

progfolio:
> I wondered about optimizing the nested with-temp-buffer, so the REPLACE is a great idea.

Seems to work good on my end with Emacs 27, 28, and 30.

> I'm surprised if buffer-substring-no-properties is making a difference, because there shouldn't be any properties (no sentinel manipulating the standard out).

I think the actual C function is just faster.
That seems to be the general case.

> Should this have a changelog for an internal improvement?

I think we're good here.


> feel free to squash of course.

I've done so and merged to "develop". Thanks again!

PR #1109, opened 2023-08-08
fix byte-compilation

progfolio:


colonelpanic8:
@progfolio shouldn't this get merged into master. This is a pretty significant bug, IMO.

colonelpanic8:
FWIW, it did fix the issue for me.

colonelpanic8:
@progfolio I'm not sure if this is related, but this issue has reemerged for me in a very consistent way for me now today:

https://github.com/magit/magit/discussions/4997

Seems like this change might be related, but I haven't had time to look in to it deeply yet.

Issue #1120, opened 2023-09-19
Problem moving straight directory after update


nivaca:
This issue has just begun after updating my config. (I was on bookstrap-version 5 and moved up to 6; don't know if that's important.)

I need to have my `straight/` directory inside my `~/.emacs.d/` (as I share my main Emacs config with other machines using Dropbox, which is syncing my main emacs directory, `~/Dropbox/emacs/`).

Before this, I had this line in my bootstrap code, which used to work:

```elisp
(setq straight-base-dir (expand-file-name "~/.emacs.d"))
```

However, the straight directory is now deployed in my main emacs directory (`~/Dropbox/emacs/`).

Is there any way to revert this issue?

I'm running GNU Emacs 29.1 (build 1, x86_64-redhat-linux-gnu, GTK+ Version
 3.24.38, cairo version 1.17.8) of 2023-07-27
On Fedora Linux 28.

Thanks!

progfolio:
Is "~/.emacs.d/" a symlink?

nivaca:
No, just the real directory.

progfolio:
Please link to or share your configuration.

nivaca:
It's here:
https://raw.githubusercontent.com/nivaca/emacsconf/main/mylisp/mypackages.el
Notice that I had to comment lines 6-7 to avoid error messages.

progfolio:
> It's here: https://raw.githubusercontent.com/nivaca/emacsconf/main/mylisp/mypackages.el Notice that I had to comment lines 6-7 to avoid error messages.

What are the error messages?

nivaca:
OK. When I uncomment both lines, I get no error, but the `straight` directory is created inside my `~/emacs` directory (which is actually symlinked), which I don't want.
But when I have the two following lines:

```elisp
(setq straight-base-dir (expand-file-name "~/.emacs.d"))
(setq straight-build-dir (expand-file-name "~/.emacs.d"))
```

I get the following error:

`File is missing: Cannot open load file, No such file or directory, /home/nivaca/emacs/straight/repos/straight.el/bootstrap.el`

raxod502:
Get rid of the `straight-build-dir` setting; that almost certainly doesn't do what you want.

For `straight-base-dir`, try https://github.com/radian-software/straight.el/pull/1123.

nivaca:
Ok, I have this in my init:

```lisp
(setq straight-base-dir (expand-file-name "~/.emacs.d"))

[...]

PR #1123, opened 2023-09-23
[#1120] Respect straight-base-dir in bootstrap

raxod502:
Pretty self explanatory. If the user modifies `straight-base-dir` then of course they must modify the bootstrap snippet to use that directory instead of `user-emacs-directory`. However, we might as well do it automatically for them, instead of requiring this extra step.


progfolio (approved):

progfolio:
> we might as well do it automatically for them, instead of requiring this extra step.
 
agreed!

Issue #1121, opened 2023-09-20
straight-mock-repo.el shows as a broken symlink in my local repo


doolio:
https://github.com/radian-software/straight.el/blob/9b11112b2e7aedd994feb2d8f95bd66dbc5749a5/tests/mocks/.emacs.d/straight/build/straight-mock-repo/straight-mock-repo.el#L1

![image](https://github.com/radian-software/straight.el/assets/13859009/980a2989-b461-4aa0-93e6-563928bffa63)

progfolio:
Is this causing a test to fail, or any sort of breakage?

doolio:
No, not that I'm aware of at least. I  only discovered it because the `chkstow` tool that is part of `stow` (which I use to manage my dotfiles) reported it as a bogus link. Feel free to close if you believe this is a non-issue.

progfolio:
Thanks for reporting, but I don't think this is harming anything.

raxod502:
It's a little odd that it has an absolute path hardcoded for your personal computer, though. That hardcoding makes me worry that maybe a test will pass on your machine but not anywhere else. I filed https://github.com/radian-software/straight.el/pull/1124 to change this to a relative link, so at least it will be equally inconvenient for everybody. Long term, it might make sense to have `straight.el` always use relative symlinks, since that would make the `straight` directory relocatable.

progfolio:
> It's a little odd that it has an absolute path hardcoded for your personal computer, though.

Fair point. I agree it could be a relative link.

> Long term, it might make sense to have straight.el always use relative symlinks, since that would make the straight directory relocatable.

Good idea. I've run into problems with links breaking when moving test environments into storage to keep for later.

PR #1124, opened 2023-09-23
[#1121] Change broken symlink

raxod502:
https://github.com/radian-software/straight.el/issues/1121


progfolio (approved):
Looks good to me.

Issue #964, opened 2022-06-21
Straight Doesn't Add Some of the Dependencies When Building Magit


nikuolive:
## What's wrong
Straight doesn't use some of the dependencies when building package such as Magit (in this case it doesn't add the Transient dependency) which leads to one of the macro used by Magit to be undefined when byte-compiled.

<details open><summary>Test Case</summary>

```emacs-lisp
(straight-bug-report
  :preserve t
  :post-bootstrap 
  (straight-use-package 'magit))
```
</details>

- Test run at: `2022-06-21 18:04:08`
- system-type: `gnu/linux`
- straight-version: `prerelease (grafted, HEAD -> develop, replaced) 0e204d4 2022-06-16`
- emacs-version: `GNU Emacs 28.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.33, cairo version 1.17.6)
 of 2022-04-28`

<details><summary>Output</summary>

```emacs-lisp
Bootstrapping straight.el...
Bootstrapping straight.el...done
Looking for gnu-elpa-mirror recipe → Cloning melpa...
Looking for gnu-elpa-mirror recipe → Cloning melpa...done
Looking for nongnu-elpa recipe → Cloning gnu-elpa-mirror...
Looking for nongnu-elpa recipe → Cloning gnu-elpa-mirror...done
Looking for emacsmirror-mirror recipe → Cloning nongnu-elpa...
Looking for emacsmirror-mirror recipe → Cloning nongnu-elpa...done
Looking for emacsmirror-mirror recipe → Cloning el-get...
Looking for emacsmirror-mirror recipe → Cloning el-get...done
Looking for straight recipe → Cloning emacsmirror-mirror...
Looking for straight recipe → Cloning emacsmirror-mirror...done
Building straight...
Building straight...done

Test run with version: prerelease (HEAD -> develop, origin/develop) 0e204d4 2022-06-16
Cloning magit...
Cloning magit...done
Building magit...
Building magit → Cloning compat...
Building magit → Cloning compat...done
Building magit → Building compat...
Building magit → Building compat...done
Building magit → Cloning dash.el...
Building magit → Cloning dash.el...done
Building magit → Building dash...
Building magit → Building dash...done
Building magit → Building git-commit...
Building magit → Building git-commit → Cloning transient...
Building magit → Building git-commit → Cloning transient...done
Building magit → Building git-commit → Building transient...
Building magit → Building git-commit → Building transient...done
[...]

PR #1125, opened 2023-10-21
[#964] Allow installing seq from GNU ELPA

raxod502:
I think this is the right solution to https://github.com/radian-software/straight.el/issues/964 - per Jonas, the version of `seq.el` on GNU ELPA is newer than what is in Emacs now, and it is a legit package that people need to be able to install as a dependency. I'm not sure why this was in the list to begin with, maybe things were different in the past.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -3323,8 +3323,7 @@ does such a good job of discouraging contributions anyway."
 (defcustom straight-recipes-gnu-elpa-ignored-packages
   '(cl-generic
     cl-lib
-    nadvice
-    seq)
+    nadvice)
   "Packages from GNU ELPA that we should pretend don't exist.
 Such packages would break things if they were installed. For
 example, the `cl-lib' package from GNU ELPA is not the


progfolio:
> I'm not sure why this was in the list to begin with, maybe things were different in the past.

Git blame says seq was ignored in light of this: https://github.com/radian-software/straight.el/pull/538

raxod502:
Blargh, guess I should have checked that.

But I still think this is the right change, actually, and the addition of `seq` in that PR was only a workaround to another bug, which I think may have been fixed since.

I had a read through https://github.com/doomemacs/doomemacs/issues/3406, and it looks from the backtrace in https://github.com/doomemacs/doomemacs/issues/3406#issuecomment-645649199 like the issue is `cl-subseq` being defined using `seq-subseq`, and `seq-subseq` also being defined using `cl-subseq`.

Now looking at [the current version of `seq`](https://github.com/emacs-straight/seq/blob/4355cde01767911660f761cf874edb412d06bd7a/seq-25.el#L176-L208), it seems the implementation of `seq-subseq` no longer relies on `cl-subseq`. That changed in [this 2020-09-04 commit in GNU ELPA Mirror](https://github.com/emacs-straight/seq/commit/e318b9ee31c18467090b48f66aa06487991899c2#diff-ea7017bd4314bcdc86e31f9ee39b12db023e7946277cd2bbb94e641833b7e922L140-R181), which corresponded to [this 2020-09-04 commit in GNU ELPA](https://git.savannah.gnu.org/cgit/emacs/elpa.git/commit/?h=externals/seq&id=c60389dc386f8cce940b9d8e85b997ad1ca5022f), which corresponded to [this 2019-10-27 commit in Emacs core](https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=0e4dd67aae8b10032317a29a6bd99d2d4a64c897).

The Doom bug was reported 2020-06-17, which precedes all of this by a couple months, so I think it is safe to say that with three years having gone by since the functions no longer have a mutual dependency, we can probably remove this workaround. Does that seem reasonable to you?

progfolio:
Sounds good to me

progfolio (approved):

Issue #1131, opened 2023-11-15
tree-sitter tsc module will be re-built instead of being downloaded with straight-rebuild-package


scturtle:
## What's wrong

Tree-sitter `tsc` module will be re-built instead of being downloaded with `straight-rebuild-package`.

## Directions to reproduce.

`.emacs.d/init.el`:
```
(defvar bootstrap-version)
(let ((bootstrap-file
       (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
      (bootstrap-version 5))
  (unless (file-exists-p bootstrap-file)
    (with-current-buffer
        (url-retrieve-synchronously
         "https://raw.githubusercontent.com/radian-software/straight.el/develop/install.el"
         'silent 'inhibit-cookies)
      (goto-char (point-max))
      (eval-print-last-sexp)))
  (load bootstrap-file nil 'nomessage))

(straight-use-package 'use-package)
(setq straight-use-package-by-default t)

(use-package tree-sitter)
(use-package tree-sitter-langs)
```

Run `emacs --init-directory /tmp/.emacs.d` the first time, the `tsc` module should be downloaded correctly.

In Emacs, run `C-u M-x straight-rebuild-package` and enter `tree-sitter` to rebuild `tree-sitter` and `tsc`. The `tsc` module will be built from source this time.

## details

`*straight-byte-compilation*`:
```
In toplevel form:
tree-sitter-query.el:15:2: Error: Module is not GPL compatible: "/tmp/.emacs.d/straight/build/tree-sitter-langs/bin/scheme.dylib"
Compiling /tmp/.emacs.d/straight/build/tree-sitter/tree-sitter.el...
Done (Total of 6 files compiled, 1 failed, 1 skipped)
```

it seems that `.emacs.d/straight/build/tree-sitter-langs/bin` is included in `load-path` while doing `straight-rebuild-package`. Then during downloading `tsc` module with `url-copy-file`, not knowing why, Emacs will try to load the tree-sitter language files `*.dylib` / `*.so` as Emacs module and fail. Then tree-sitter will turn to compile `tsc` from source.

### Version information

* Emacs version: 29.1
* Operating system: MacOS and Linux

raxod502:
I think you're right. Furthermore I think the reason the subdirectory is (incorrectly) added to `load-path` is this:

[...]

PR #1132, opened 2023-11-18
straight--build-compile: fix compile process load-path

progfolio:
Compute child byte-compilation process load-path via straight-dependencies.
This should include the necessary transitive dependencies as well as
direct dependencies.

See: #1131

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -5257,19 +5257,22 @@ individual package recipe."
 RECIPE should be a straight.el-style plist. Note that this
 function only modifies the build folder, not the original
 repository."
-  (let* ((dir (straight--build-dir (plist-get recipe :package)))
+  (let* ((pkg (plist-get recipe :package))
+         (dir (straight--build-dir pkg))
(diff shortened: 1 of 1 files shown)


raxod502 (approved):
:sparkles:

Patryk27:
Hi, I can also confirm this works!

I've got a problem where `eat` wasn't getting byte-compiled at all within Doom Emacs, because the current approach causes the byte-compiler to fail and stop on:

```
Checking /home/pwy/.emacs.d/.local/straight/build-29.4/eat/...
Compiling /home/pwy/.emacs.d/.local/straight/build-29.4/eat/eat-autoloads.el...
Compiling /home/pwy/.emacs.d/.local/straight/build-29.4/eat/eat.el...

In toplevel form:
eat.el:91:2: Error: Module is not GPL compatible: "/home/pwy/.emacs.d/.local/straight/build-29.4/tree-sitter-langs/bin/json.so"
```

Using the patch here helps and everything seems to be working correctly now 🙂

PR #1149, opened 2024-02-19
Set print-symbols-bare to t while saving the build cache

PythonNut:
This fixes a bug where recipes may contain symbols with positions, which have a printed representation but no read syntax, causing the subsequent load of the build cache to fail. Symbols with positions may be generated during the execution of the byte compiler.

The fix is to bind `print-symbols-bare` to `t` during the printing of the build cache in `straight--save-build-cache`. This strips the attached positions from the symbols. The attached positions are not used by `straight.el` in any way, so the build cache remains valid when loaded. The distinction between bare symbols and symbols with positions is only relevant here (during build cache save), since `symbols-with-pos-enabled` is bound to `t` during the execution of the byte compiler, causing symbols with positions to transparently behave as their contained bare symbols.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -4260,11 +4260,17 @@ The name of the cache file is stored in
   (straight--log 'modification-detection "Saving build cache")
   (unless straight-safe-mode
     (with-temp-buffer
-      ;; Prevent mangling of the form being printed in the case that
-      ;; this function was called by an `eval-expression' invocation
-      ;; of `straight-use-package'.
-      (let ((print-level nil)
-            (print-length nil))
+      (let (;; Prevent mangling of the form being printed in the case that
+            ;; this function was called by an `eval-expression' invocation
+            ;; of `straight-use-package'.
+            (print-level nil)
+            (print-length nil)
+            ;; Print symbols with attached positions, which can occur
+            ;; during byte compilation, as bare symbols so they can be
+            ;; read when loading the cache.
+            (print-symbols-bare t))
+        ;; Stop the byte-compiler from complaining about unused binding.
+        (ignore print-symbols-bare)
         ;; The version of the build cache.
         (print straight--build-cache-version (current-buffer))
         ;; Record the current Emacs version. If a different version of


straight.el
```diff
@@ -4260,11 +4260,15 @@ The name of the cache file is stored in
...
-            (print-length nil))
+      (let (;; Prevent mangling of the form being printed in the case that
+            ;; this function was called by an `eval-expression' invocation
+            ;; of `straight-use-package'.
+            (print-level nil)
+            (print-length nil)
+            ;; Print symbols with attached positions, which can occur
+            ;; during byte compilation, as bare symbols so they can be
+            ;; read when loading the cache.
+            (print-symbols-bare t))
```
progfolio, line 4271:
```suggestion
            (print-symbols-bare t))
          (ignore print-symbols-bare)
```

It looks like that variable was only introduce in Emacs 29.
That's why the CI is failing with a byte-compiler warning:

https://github.com/radian-software/straight.el/actions/runs/7953631018/job/21709882276?pr=1149#step:3:2400

PythonNut: That (at least partly) explains why I only noticed this problem recently!

progfolio: > You could imagine why something like this might be plausible when the package needs to be installed at macro expansion time for whatever reason.

I'm not sure I follow. When/why would one want to install packages during macro expansion?
@raxod502 any thoughts on this use case?

> Interestingly, I couldn't get this to reproduce using eval-and-compile. This is because eval ignores positions when symbols-with-pos-enabled is non-nil (as it is during byte-compilation).

Why not have the macro expansion make use of `eval-and-compile` in that case?:

```emacs-lisp
(defmacro test-macro (&rest args)
  `(eval-and-compile ...))
```

PythonNut: > I'm not sure I follow. When/why would one want to install packages during macro expansion? @raxod502 any thoughts on this use case?

Well, for example [use-package](https://github.com/jwiegley/use-package/blob/ff8c57268a3b655098bd9cfd05e9d4a689533c3a/use-package-ensure.el#L192) does this. The reason is that the package may define macros which are used subsequently in the file. To ensure the later macros can be expanded, the package must be installed at compile time.

> Why not have the macro expansion make use of `eval-and-compile` in that case?:

I think it should be possible to use `eval-and-compile` or possibly `eval-when-compile` to achieve the desired effect, but since the fix is simple and self contained, I don't see a reason not to accommodate users who are not using those macros (or use packages that do not). Especially so given that the unpatched behavior 1. isn't obviously correct and 2. leads to a silent failure that is difficult to diagnose.

progfolio: To be clear, I'm not against the solution. I'm just trying to understand the problem fully before we apply a fix.
That byte compilation warning still needs to be addressed before we can merge, too.

PythonNut: Got it! Let me take a stab at fixing the warning and see if it's to your liking.

progfolio:
@PythonNut see the suggested change above. It should be enough to let-bind the variable and then immediately pass it to ignore.

PythonNut:
Oh, do you prefer that to the `defvar`?

progfolio:
> Oh, do you prefer that to the `defvar`?
 Sorry I didn't see the earlier edit. I can switch over to `ignore` later today.

Not a problem. Yes the let-binding is a simpler solution and won't create a global variable.

straight.el
raxod502: Installing packages during byte-compilation doesn't seem like a great idea to me, but it also doesn't seem so wrong that I would say we should not allow it. Of course we should fix the bug so I agree with the submitted patch.

PythonNut:
@progfolio Is there anything else I need to do to get this across the finish line?

raxod502:
It looks fine to me!

PR #1151, opened 2024-03-02
[#1150] Fix load-history when eval autoloads

### LINKED ISSUE #1150: `use-package` integration causes incorrect `symbol-file`
## What's wrong

If I add a package with `use-package` and `:straight t`, then evaluating `(symbol-file 'the-package)` will give the location of the `use-package` call, not the location of the package itself. This is *not* the case when one adds the package with straight directly.

This has caused bugs in packages I use such as [julia-snail](https://github.com/gcv/julia-snail)

## Directions to reproduce

Put the following in your init.el:

```elisp
(use-package julia-snail
  :straight t)
```
and then evaluate `(symbol-file 'julia-snail)`. This will return e.g. `"/home/mason/.emacs.d/myinit.el"`, whereas if you do
```
(straight-use-package 'julia-snail)
```
then `(symbol-file 'helm)` correctly returns `"/home/mason/.emacs.d/straight/build/julia-snail/julia-snail.elc"`

### Version information

* Emacs version: GNU Emacs 29.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.40, cairo version 1.18.0)
* Operating system: Manjaro Linux

raxod502:
Resolves #1150

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -5491,11 +5491,22 @@ RECIPE is a straight.el-style plist."
                        straight--autoloads-cache)))
           ;; Some autoloads files expect to be loaded normally, rather
           ;; than read and evaluated separately. Fool them.
+          ;;
+          ;; We also need to abuse `current-load-list' so that
+          ;; autoload entries go properly into the current entry,
+          ;; since normally that is hardcoded to happen during `load'
+          ;; which we are not using here.
+          ;;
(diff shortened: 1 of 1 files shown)


wyuenho:
This works well, any reason why it's not merged yet?

raxod502:
Nope, just lack of attention on my part.

Issue #1167, opened 2024-09-08
bbdb fails to install


rookiejet:
## What's wrong

bbdb package fails to install.

## Directions to reproduce

<details open><summary>Test Case</summary>

```emacs-lisp
(straight-bug-report
  :post-bootstrap 
  (straight-use-package 'bbdb))
```
</details>

- Test run at: `2024-09-08 14:02:16`
- system-type: `windows-nt`
- straight-version: `prerelease (HEAD -> master, origin/master, origin/develop, origin/HEAD) 88e574a 2024-05-23`
- emacs-version: `GNU Emacs 29.4 (build 2, x86_64-w64-mingw32)
 of 2024-07-05`

<details><summary>Output</summary>

```emacs-lisp
Bootstrapping straight.el...
Bootstrapping straight.el...done

Test run with version: prerelease (HEAD -> develop, origin/master, origin/develop, origin/HEAD) 88e574a 2024-05-23

Packages:
"straight"                n/a                  develop 88e574a 2024-05-23
"org-elpa"                n/a                  n/a
"melpa"                   n/a                  master 0c608bf8 2024-09-08
"gnu-elpa-mirror"         n/a                  master 1d6212f 2024-08-28
"nongnu-elpa"             n/a                  main 9496622 2024-09-04
"el-get"                  melpa                master 2f3d2be3 2024-09-04
"emacsmirror-mirror"      n/a                  master 6c72db6 2024-09-07

Error: (error "Invalid sub-list head in :files directive: :rename")
Looking for gnu-elpa-mirror recipe -> Cloning melpa...
Looking for gnu-elpa-mirror recipe -> Cloning melpa...done
Looking for nongnu-elpa recipe -> Cloning gnu-elpa-mirror...
Looking for nongnu-elpa recipe -> Cloning gnu-elpa-mirror...done
Looking for emacsmirror-mirror recipe -> Cloning nongnu-elpa...
Looking for emacsmirror-mirror recipe -> Cloning nongnu-elpa...done
Looking for emacsmirror-mirror recipe -> Cloning el-get...
Looking for emacsmirror-mirror recipe -> Cloning el-get...done
Looking for straight recipe -> Cloning emacsmirror-mirror...
Looking for straight recipe -> Cloning emacsmirror-mirror...done
Building straight...
Building straight...done
Cloning bbdb...
Cloning bbdb...done
Building bbdb...

```
</details>

<details><summary>init debug output</summary>

```emacs-lisp
[...]

PR #1169, opened 2024-09-22
[#1167] Support MELPA :rename

raxod502:
Close https://github.com/radian-software/straight.el/issues/1167. Remove the default behavior of renaming `.el.in` files in MELPA recipes as it has been removed upstream. Support `:rename` (which was added to upstream as a replacement) by translating it in the MELPA recipe fetcher into the cons-cell style renaming that was already supported in `straight.el`.

Add tests for the modified function, and mock data to support the tests. Change the indentation in the testing code to what Emacs wanted it to be by default, and add linting for the indentation in this file. (The changes are not entirely improvements, but if we want better indentation, we should make sure Emacs is configured to do it properly in this project rather than having it be a personal choice for each contributor.)


progfolio (approved):
Looks good to me

PR #1178, opened 2025-01-20
feat: respect absolute path for straight-profiles for bootstrap install

milanglacier:
Currently the bootstrap assumes that the lockfile path must be relative file relative to `straight-base-dir/striahgt/versions`,

however, some users might prefer to manage their lockfile in their version controlled dotfiles repo (`~/.config/emacs/straight/versions`), and manage the straight data under XDG spec, aka `~/.local/share/emacs/`.

This patch updates the bootstrap script so that the user's lockfile file path is an absolute path, it will be used

### DIFF
--- a/install.el
+++ b/install.el
@@ -118,9 +118,10 @@
     ;; the recipe specification, and forgot to update which repository
     ;; their init-file downloaded install.el from).
     (dolist (lockfile-name (mapcar #'cdr straight-profiles))
-      (let ((lockfile-path (concat straight-install-dir
-                                   "straight/versions/"
-                                   lockfile-name)))
+      (let ((lockfile-path
+             (expand-file-name
+              lockfile-name
+              (concat straight-install-dir "straight/versions"))))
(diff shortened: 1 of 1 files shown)


milanglacier:
Hi, the CI error is:

```
make: *** [Makefile:101: docker] Error 2
Error: Process completed with exit code 2.
```

I have no idea what is this error. Looks like this is not an error from elisp but rather from build script?

raxod502:
Could we use `expand-file-name` instead of `concat`, which handles both files the same?

You can see the build error above in the log output:

image

milanglacier:
> Could we use expand-file-name instead of concat, which handles both files the same?


Thanks, updated. Good to know that expand-file-name can handle cases like this :)

raxod502 (approved):
thx- looks like we don't need any version number bumps because this lives purely in `install.el` and is backwards compatible. You can ensure your new patch is available to your Emacs configuration when bootstrapping, by changing `master` to `develop` in the bootstrap snippet.

PR #1179, opened 2025-01-20
Resolve Docker Hub rate limits

raxod502:
Pull images from GHCR instead of directly from Docker Hub. Right now, I just published the tags manually, but in future we can add an Actions job in the GitHub organization that will periodically update them to apply security patches etc.


raxod502:
I don't understand why the latest version of this PR is failing with a 401 error. I filed https://support.github.com/ticket/personal/0/3192280 to ask GitHub support.

Issue #902, opened 2021-12-13
check-indent test fails across Emacs versions.


progfolio:
There have been some recent changes to the default indentation in emacs-lisp-mode:

https://github.com/emacs-mirror/emacs/commit/32df2034234056bf24312ef5883671b59a387520

The previous indentation was clearly a bad heuristic, so I'm glad it's been changed. However, the upshot is that now our checkindent test will fail for the affected cases. (I was bitten by this on Emacs 29 when let-binding `default-directory` in #901. If it was indented correctly for v29, it failed in earlier versions and vice versa.)

I think the best approach is to drop the checkindent test from CI for the time being. We can always add it back if we can think of a test that will work across all versions, or (in the distant future) when we have dropped support for versions which have different indentation.

PR #1194, opened 2025-06-10
Normalize straight.el indentation on all Emacs versions

raxod502:
* Enable linting against Emacs 30 in CI
* Re-enable checkindent linter
* Update indentation so that it is the same in all Emacs versions 25-30
    * Check the comments in `indent.el` for an explanation of how


raxod502:
FYI @progfolio re #902, #903, #908

PR #1203, opened 2025-06-11
Automatically detect when filesystem watcher fails to start

raxod502:
In such cases, notify the user, and provide a way to refresh it.


raxod502:
This fixes https://github.com/radian-software/straight.el/issues/447

PR #1208, opened 2025-06-13
Fix ordering issue with el-get bootstrap

raxod502:
Re-loading the init-file would get you a warning about the recipe for el-get being redefined, which turned out to be because recipe inheritance was enabled, so the recipe for `straight.el` itself was trying to be looked up in existing recipe repos, which were still defined from last init, but not with their recipes, only the symbols, so the recipe repo recipes themselves were being looked up, and it turns out el-get had a recipe for el-get inside it, which was different from the one listed in `straight.el`. Hilarious!

### DIFF
--- a/bootstrap.el
+++ b/bootstrap.el
@@ -106,20 +106,25 @@
 ;; have to explicitly clear the caches.
 (straight--reset-caches)
 
-;; First we register (and build) straight.el itself.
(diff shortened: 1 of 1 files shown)


progfolio:
> it turns out el-get had a recipe for el-get inside it

lol.. oops!

PR #1272, opened 2025-11-07
Use lexical-binding when loading entries from autoloads cache

tarsius:
straight--activate-package-autoloads: Use lexical bindings

Autoload files use lexical bindings and as a result some packages
actually depend on that.  So we need to use lexical bindings when
evaluating entries from the autoloads cache.

For an example see https://github.com/magit/magit/issues/5476.

Ps: The contributor guide told me to update `CHANGELOG.md`.  I would have done that, but I could not find that file.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -6255,7 +6255,7 @@ RECIPE is a straight.el-style plist."
                 (current-load-list nil))
             ;; car is the feature list, cdr is the autoloads.
             (dolist (form (cdr (gethash package straight--autoloads-cache)))
-              (eval form))
+              (eval form t))
             (when current-load-list
               (push (cons load-file-name current-load-list) load-history))))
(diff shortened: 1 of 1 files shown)


raxod502:
No changelog for `straight.el` yet, sorry for the wrong instructions. Updating that now. Change looks good, thank you.

nalcabio-tom:
> Do all Straight users use `main` or is there the option to stick to the latest release?

As a unsophisticated user of straight, I just copy and paste from the documentation, and if I remember correctly, everything I've seen says to use the develop branch.

Edit to add: For example, https://github.com/jamescherti/minimal-emacs.d?tab=readme-ov-file#configuring-straightel

raxod502:
> is there the option to stick to the latest release?

`straight.el` does not provide numbered releases, at present.

PR #1279, opened 2026-02-06
allow elpa/ directory to exist if there are no actual packages in it

igmacs:
Currently straight.el does not provide alternatives to package.el's `describe-package` or `list-packages`, so users are forced to use those if they are interested in that feature. However, configuring package.el to retrieve all the packages from Elpa and Melpa will result in the `.emacs.d/elpa` directory being created, which straight.el doesn't like because it thinks it means the user is mixing package managers.

With this PR straight.el would allow that directory to exist, as long as there are no packages installed inside, which is the actual sign that the package managers are being mixed.

### DIFF
--- a/straight.el
+++ b/straight.el
@@ -7472,13 +7472,17 @@ Must be set before bootstrap."))
         (setq straight-package--warning-displayed t))
     (with-eval-after-load 'package
       (unless straight-package--warning-displayed
-        (when (file-exists-p (bound-and-true-p package-user-dir))
-          (display-warning
-           '(straight package)
-           (concat
-            "package.el was loaded when straight.el was already loaded. "
-            tips)
-           :warning))
+        (let* ((user-dir (bound-and-true-p package-user-dir))
+               (files (when (file-exists-p user-dir)
+                        (directory-files user-dir nil "[:alnum]")))
+               (packages (delete "archives" (delete "gnupg" files))))
+          (when packages
+            (display-warning
+             '(straight package)
+             (concat
+              "package.el was loaded when straight.el was already loaded. "
+              tips)
+             :warning)))
         (setq straight-package--warning-displayed t)))))
 
 ;;;;;; Mode variables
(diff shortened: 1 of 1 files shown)


straight.el
```diff
@@ -7472,7 +7472,10 @@ Must be set before bootstrap."))
         (setq straight-package--warning-displayed t))
     (with-eval-after-load 'package
       (unless straight-package--warning-displayed
-        (when (file-exists-p (bound-and-true-p package-user-dir))
+        (when-let* ((user-dir (bound-and-true-p package-user-dir))
+                    (_ (file-exists-p user-dir))
+                    (files (directory-files user-dir nil "[:alnum]"))
```
igmacs: The regexp is to exlcude the "." and ".." directories, which are otherwise listed too

straight.el
```diff
@@ -7472,7 +7472,10 @@ Must be set before bootstrap."))
         (setq straight-package--warning-displayed t))
     (with-eval-after-load 'package
       (unless straight-package--warning-displayed
-        (when (file-exists-p (bound-and-true-p package-user-dir))
+        (when-let* ((user-dir (bound-and-true-p package-user-dir))
+                    (exist? (file-exists-p user-dir))
```
igmacs: CI doesn't let me use the unused prefix '_' because it says the variable is not unused, which I guess is technically true if you consider macro expansion, but still not quite right IMO

straight.el
```diff
@@ -7472,13 +7472,17 @@ Must be set before bootstrap."))
...
-        (when (file-exists-p (bound-and-true-p package-user-dir))
-          (display-warning
-           '(straight package)
-           (concat
-            "package.el was loaded when straight.el was already loaded. "
-            tips)
-           :warning))
+        (let* ((user-dir (bound-and-true-p package-user-dir))
+               (files (when (file-exists-p user-dir)
+                        (directory-files user-dir nil "[:alnum]")))
```
raxod502:
```suggestion
                        (directory-files user-dir nil "[:alnum:]")))
```

meant to use the special character class? alternatively, just

```suggestion
                        (directory-files user-dir nil "^[^.]")))
```

igmacs: Ups, not sure how it worked then, but it did. I'll use your second suggestion.

raxod502 (approved):

raxod502:
Thank you!

PR #1291, opened 2026-07-28
Fix obsolete macro warnings and add lexical-binding

jamescherti:
Fix obsolete macro warnings and add lexical-binding:
- Replace deprecated when-let and if-let macros with their starred counterparts (when-let* and if-let*) in straight.el and straight-x.el.
- Add lexical-binding: t to indent.el

### DIFF
--- a/indent.el
+++ b/indent.el
@@ -1,3 +1,5 @@
+;;; indent.el --- Indent forms from other packages -*- lexical-binding: t -*-
+
 ;; You can load this file to make sure that forms from other packages
 ;; are indented correctly. It should cause the indentation to be the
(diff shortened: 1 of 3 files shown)


raxod502:
Looks like the long-lines script needs to be updated (or the package comment for `indent.el` reworded).

jamescherti:
Hello @raxod502,

Thank you for pointing this out. I have reworded it.

raxod502 (approved):
Well, if it passes byte-compilation, I think that should prove it works! Thank you :)

jamescherti:
My pleasure, @raxod502!

PR #1292, opened 2026-08-15
Fix: Warning: ‘when-let’ is an obsolete macro

jamescherti:
This pull request fixes the warning: ‘when-let’ is an obsolete macro:

- straight.el:3643:12: Warning: ‘when-let’ is an obsolete macro (as of 31.1); use ‘when-let*’ or ‘and-let*’ instead.
- straight.el:4462:25: Warning: ‘when-let’ is an obsolete macro (as of 31.1); use ‘when-let*’ or ‘and-let*’ instead.


raxod502 (approved):
Thanks!

